PR #314, opened 2026-03-16
refactor(auth): split AuthMiddleware into Authenticator + GuardChain (#313)

### LINKED ISSUE #313: refactor(auth): split AuthMiddleware into Authenticator + GuardChain
## Summary

Split `AuthMiddleware` into two properly named, single-responsibility components:

1. **`Authenticator`** — identity resolver ("who is this?") → returns `Identity(trust_level, is_admin, user_id)`
2. **`GuardChain`** — composable policy enforcement ("can they proceed?") → accept/reject

## Problem

- `AuthMiddleware` is misnamed — it resolves identity but doesn't enforce anything; the caller does that with ad-hoc if-checks
- BLOCKED guard logic is duplicated in `telegram_inbound.py:63` and `discord_inbound.py:48`
- Two competing admin-check patterns: `require_admin(msg, admin_user_ids)` in builtin_commands vs `is_admin(msg.user_id)` module-global in plugins
- Guard logic scattered across 4 layers (adapter, shared, pipeline, command handlers)

## Solution

- Rename `AuthMiddleware` → `Authenticator`, `check()` → `resolve()`, return `Identity` dataclass
- New `GuardChain` with `Guard` protocol — `BlockedGuard` replaces ad-hoc if-checks
- Add `is_admin: bool` field to `InboundMessage` (set during authentication)
- Unify `require_admin()` to read `msg.is_admin` instead of threading `admin_user_ids`
- `auth.py` becomes a backward-compat facade re-exporting `Authenticator as AuthMiddleware`

## Correct Order

```
Authenticate (resolve identity) → Guard (enforce policy) → Pipeline
```

## Files

| File | Change |
|------|--------|
| `core/identity.py` | NEW — `Identity` dataclass |
| `core/authenticator.py` | NEW — renamed from `auth.py` |
| `core/guard.py` | NEW — `Guard` protocol, `BlockedGuard`, `GuardChain` |
| `core/auth.py` | REWRITE — facade re-exports |
| `core/admin.py` | SIMPLIFY — deprecate in favor of `msg.is_admin` |
| `core/message.py` | ADD `is_admin: bool = False` field |
| `core/builtin_commands.py` | SIMPLIFY — `require_admin(msg)` 1-arg |
| `core/workspace_commands.py` | SIMPLIFY — same |
| `core/command_router.py` | STOP threading `admin_user_ids` |
| `adapters/telegram_inbound.py` | USE `resolve()` + `guard_chain.run()` |
| `adapters/discord_inbound.py` | USE `resolve()` + `guard_chain.run()` |
| `__main__.py` | WIRE `Authenticator` + `GuardChain` |

## Plan

See `artifacts/plans/guard-authenticator-middleware-plan.mdx`

## Acceptance Criteria

- [ ] `AuthMiddleware.check()` calls replaced by `Authenticator.resolve()` returning `Identity`
- [ ] `BlockedGuard` replaces all ad-hoc `if trust == BLOCKED: return` checks
- [ ] `msg.is_admin` is the single source of truth for admin status
- [ ] `require_admin()` no longer needs `admin_user_ids` argument
- [ ] `auth.py` facade preserves backward compat for existing imports
- [ ] `uv run pytest -x && uv run pyright` passes

MickaelV0:
## Summary

- **`AuthMiddleware` → `Authenticator`** — renamed, `check()` → `resolve()`, returns `Identity(user_id, trust_level, is_admin)` instead of bare `TrustLevel`
- **New `GuardChain`** with `Guard` protocol + `BlockedGuard` — replaces ad-hoc `if trust == BLOCKED: return` in both adapter inbound handlers
- **Unified admin checks** — `msg.is_admin` field on `InboundMessage`, `require_admin(msg)` 1-arg form, plugins read `msg.is_admin` directly
- **`auth.py` facade** preserves all existing imports (`AuthMiddleware`, `_ALLOW_ALL`, `_DENY_ALL`, `TrustLevel`)

## Files

| Change | Files |
|--------|-------|
| NEW | `core/identity.py`, `core/authenticator.py`, `core/guard.py` |
| FACADE | `core/auth.py` |
| MODIFY | `adapters/{telegram,discord}_inbound.py`, `adapters/{telegram,discord}_normalize.py`, `adapters/{telegram,discord}.py` |
| MODIFY | `core/message.py`, `core/builtin_commands.py`, `core/workspace_commands.py`, `core/command_router.py`, `core/admin.py` |
| MODIFY | `plugins/svc/handlers.py`, `plugins/pairing/handlers.py` |
| TESTS | `test_authenticator.py`, `test_guard.py` (new), + 6 test files updated |

## Out of Scope (explicit)

- `admin_user_ids` threading in `CommandRouter`/agent constructors — deferred
- `audio_pipeline.py` BLOCKED check — deferred (operates on `InboundAudio`)
- New guard types (rate-limit, group-mention) — future work on `Guard` protocol

## Test plan

- [x] 1448 tests pass (`uv run pytest -x`)
- [x] Pyright 0 errors (`uv run pyright`)
- [x] Ruff lint clean
- [x] Pre-commit hooks pass
- [ ] Verify `from lyra.core.auth import AuthMiddleware` still works in downstream code
- [ ] Verify blocked user rejection via `BlockedGuard` (adapter integration)
- [ ] Verify admin commands (`/circuit`, `/config`, `/routing`) respect `msg.is_admin`

Closes #313

### DIFF
--- a/tests/test_bootstrap_credentials.py
+++ b/tests/test_bootstrap_credentials.py
@@ -199,7 +199,7 @@ class TestMissingCredentials:
         monkeypatch.setattr(
             AuthMiddleware,
             "from_bot_config",
-            classmethod(lambda cls, raw, platform, bot_id, store=None: MagicMock()),
+            classmethod(lambda cls, raw, platform, bot_id, **kw: MagicMock()),
         )
 
         stop = asyncio.Event()
@@ -233,7 +233,7 @@ class TestMissingCredentials:
         monkeypatch.setattr(
             AuthMiddleware,
             "from_bot_config",
-            classmethod(lambda cls, raw, platform, bot_id, store=None: MagicMock()),
+            classmethod(lambda cls, raw, platform, bot_id, **kw: MagicMock()),
         )
 
         stop = asyncio.Event()
@@ -308,7 +308,7 @@ class TestCredentialResolution:
         monkeypatch.setattr(
             AuthMiddleware,
             "from_bot_config",
-            classmethod(lambda cls, raw, platform, bot_id, store=None: MagicMock()),
+            classmethod(lambda cls, raw, platform, bot_id, **kw: MagicMock()),
         )
 
         stop = asyncio.Event()
@@ -377,7 +377,7 @@ class TestCredentialResolution:
         monkeypatch.setattr(
             AuthMiddleware,
             "from_bot_config",
-            classmethod(lambda cls, raw, platform, bot_id, store=None: MagicMock()),
+            classmethod(lambda cls, raw, platform, bot_id, **kw: MagicMock()),
         )
 
         stop = asyncio.Event()
@@ -420,7 +420,7 @@ class TestCredentialResolution:
         monkeypatch.setattr(
             AuthMiddleware,
             "from_bot_config",
-            classmethod(lambda cls, raw, platform, bot_id, store=None: MagicMock()),
+            classmethod(lambda cls, raw, platform, bot_id, **kw: MagicMock()),
         )
 
         stop = asyncio.Event()
@@ -478,7 +478,7 @@ class TestCredentialResolution:
         monkeypatch.setattr(
             AuthMiddleware,
             "from_bot_config",
-            classmethod(lambda cls, raw, platform, bot_id, store=None: MagicMock()),
+            classmethod(lambda cls, raw, platform, bot_id, **kw: MagicMock()),
         )
 
         stop = asyncio.Event()


MickaelV0:
## Code Review — PR #314 `refactor(auth): split AuthMiddleware into Authenticator + GuardChain`

**Agents:** security-auditor · backend-dev · architect · tester
**Tests:** 22 passed ✅ | **Pyright:** 0 errors ✅

---

### Spec Compliance (SC-1 → SC-14)

| SC | Criterion | Status |
|----|-----------|--------|
| SC-1 | `resolve()` returns `Identity` with correct `trust_level` + `is_admin` | ✅ |
| SC-2 | `from lyra.core.auth import AuthMiddleware` works via facade | ✅ |
| SC-3 | `GuardChain([BlockedGuard()]).run(blocked)` → `Rejection("blocked")` | ✅ |
| SC-4 | `GuardChain([]).run()` → `None` | ✅ |
| SC-5 | `telegram_inbound` uses `resolve()` + `guard_chain.run()` | ✅ code / ⚠️ tests mock `check()` |
| SC-6 | `discord_inbound` same | ✅ code / ⚠️ tests mock `check()` |
| SC-7 | `InboundMessage.is_admin` set from `Identity.is_admin` during normalize | ✅ code / ⚠️ not asserted in adapter tests |
| SC-8 | `require_admin(msg)` 1-arg reads `msg.is_admin` | ✅ |
| SC-9 | `plugins/svc` reads `msg.is_admin` | ✅ |
| SC-10 | `plugins/pairing` reads `msg.is_admin` | ✅ |
| SC-11 | `resolve(None)` → `Identity(user_id="", trust_level=BLOCKED, is_admin=False)` | ✅ |
| SC-12 | `normalize()` in both adapters accepts `is_admin` kwarg | ✅ |
| SC-13 | `pytest -x` passes | ✅ 22 passed |
| SC-14 | `pyright` passes | ✅ 0 errors |

---

### Blockers

**issue: `admin_user_ids` never forwarded to `Authenticator` at startup — `is_admin` via explicit IDs is silently dead in production**
  `src/lyra/bootstrap/multibot.py` (wherever `_build_bot_auths` / `from_bot_config` is called)
  -- security-auditor · backend-dev · architect
  Root cause: `Authenticator.__init__` accepts `admin_user_ids: frozenset[str] = frozenset()`. `resolve()` correctly evaluates `is_admin = user_id in self._admin_user_ids or trust == TrustLevel.OWNER`. But every call to `from_bot_config()` / `from_config()` omits the `admin_user_ids` parameter, so the frozenset is always empty at runtime. Only users with `trust == OWNER` will ever get `is_admin=True`. Any TRUSTED user listed in the admin config will receive `is_admin=False` on every message, silently locking them out of all admin commands. The unit tests pass because they construct `Authenticator` directly with the correct frozenset — the startup wiring path is not exercised.
  Solutions:
    1. Thread `admin_user_ids` through `_build_bot_auths()` and into `from_bot_config()` / `from_config()`. Add a startup log `"admin_user_ids configured: %d IDs"` so misconfiguration is observable.
    2. If `admin_user_ids` wiring is intentionally deferred, remove `admin_user_ids` from `Authenticator` for this PR and add it back in the follow-on — avoid shipping a parameter that looks wired but silently does nothing.
  Confidence: 95%

---

**issue: `CommandRouter._admin_user_ids` stored but never passed to any call site — admin commands silently have no gate**
  `src/lyra/core/command_router.py:107`
  -- backend-dev
  Root cause: `CommandRouter.__init__` still accepts `admin_user_ids` and stores it, but all `builtin_commands` functions now call `require_admin(msg)` (1-arg form reading `msg.is_admin`). The stored set is never passed anywhere. Operators who pass `admin_user_ids` to `CommandRouter` believe they control the gate, while the actual gate is now `msg.is_admin` set at the adapter layer (which is broken per the above). Combined, this means misconfiguration produces zero error and zero effect.
  Solutions:
    1. Remove `admin_user_ids` from `CommandRouter.__init__` entirely. Document that admin is now driven by `msg.is_admin` set at the adapter layer via `Authenticator`.
    2. For a transitional period, forward to `require_admin(msg, self._admin_user_ids or None)` — but this defeats the PR's goal and introduces the divergence problem below.
  Confidence: 95%

---

**issue: Adapter tests mock `auth.check()` — guard wiring exercises dead code path after S3**
  `tests/adapters/test_telegram.py`, `tests/adapters/test_discord.py`
  -- tester
  Root cause: Blocked-user tests do `auth.check.return_value = TrustLevel.BLOCKED`. After this PR, adapters call `auth.resolve()` + `guard_chain.run()`, not `check()`. The mock never fires; the tests pass because `resolve()` on the mock returns `MagicMock` (truthy), which makes `guard_chain.run()` see a non-`BLOCKED` Identity — the blocked-user flow is not actually tested. SC-5 / SC-6 compliance is unverified.
  Solutions:
    1. Replace `auth.check.return_value = TrustLevel.BLOCKED` with `auth.resolve.return_value = Identity(user_id="u1", trust_level=TrustLevel.BLOCKED, is_admin=False)` and inject a real `GuardChain([BlockedGuard()])`. Assert `normalize` is not called.
    2. Add one integration-level test per adapter using a real `Authenticator` (blocked user in store) + real `GuardChain` — strongest proof the wiring is correct end-to-end.
  Confidence: 95%

---

**issue: `msg.is_admin` (SC-7) never asserted in any adapter test**
  `tests/adapters/test_telegram.py`, `tests/adapters/test_discord.py`
  -- tester
  Root cause: Passing-user adapter tests verify `msg.trust_level` but never assert `msg.is_admin`. If `is_admin` is omitted from `normalize()` or if `identity.is_admin` is not forwarded, all existing tests still pass. SC-7 compliance is not demonstrated.
  Solutions:
    1. Add `assert hub_msg.is_admin is False` in the standard passing-user test, and a counterpart where `identity.is_admin=True` is injected and `hub_msg.is_admin is True` is asserted.
    2. Add a focused `test_normalize_propagates_is_admin` test in `tests/adapters/`.
  Confidence: 90%

---

**issue: `require_admin` legacy path: empty `admin_user_ids` set silently denies all users, including real admins**
  `src/lyra/core/builtin_commands.py:40-42`
  -- security-auditor · backend-dev
  Root cause: `if admin_user_ids is not None: if not admin_user_ids or ...` — an empty set is falsy, so any caller passing `set()` gets a blanket denial for every user, including those whose `msg.is_admin=True`. The two paths are not equivalent. Since the PR migrates all callers to the 1-arg form, this is a latent trap for any future callsite that reconstructs the legacy call with an empty or uninitialized set.
  Solutions:
    1. Remove the `admin_user_ids` parameter from `require_admin` entirely — all callers in this PR use the 1-arg form. Add a `# noqa` removal note.
    2. If the parameter must survive for one more release: add an explicit guard `if admin_user_ids is not None and len(admin_user_ids) == 0: log.warning("require_admin called with empty admin_user_ids — falling through to msg.is_admin"); admin_user_ids = None` to prevent silent lockout.
  Confidence: 85%

---

**issue: `is_admin=False` for OWNER users who trigger a public command bypass**
  `src/lyra/core/authenticator.py:resolve()` (~line 115)
  -- security-auditor
  Root cause: `_resolve_trust()` returns `TrustLevel.PUBLIC` when `command in public_commands`, short-circuiting the store lookup for OWNER users. `resolve()` then evaluates `is_admin = user_id in admin_user_ids or trust == OWNER` — but `trust` is now `PUBLIC`, so an OWNER user issuing `/join` receives `is_admin=False`. This is a narrow edge case (OWNER users rarely issue `/join`) but represents a logical inconsistency: admin status should depend on stored identity, not on the effective trust level of a specific command invocation.
  Solutions:
    1. Compute `is_admin` using the raw stored trust level (before command bypass), not the final `trust`: check `self._store_level(user_id) == TrustLevel.OWNER or user_id in self._admin_user_ids` independently of `_resolve_trust`.
    2. Document the behavior explicitly: "OWNER users issuing public commands receive `is_admin=False` for that request — by design." Add a test to pin the intent.
  Confidence: 82%

---

### Suggestions (blocking)

**suggestion(blocking): No integration test for SC-5/SC-6 adapter guard wiring end-to-end**
  `tests/adapters/test_telegram.py`, `tests/adapters/test_discord.py`
  -- tester
  Root cause: All current adapter auth tests use mocks for both `auth` and trust level. A real `Authenticator` + real `GuardChain` integration test would prove that `resolve()` → `guard_chain.run()` → `normalize()` is correctly wired and would have caught the `check()` mock issue above before review.
  Solutions:
    1. Add one integration adapter test per platform: real `Authenticator(store=blocked_store)` + real `GuardChain([BlockedGuard()])` — assert the message is dropped before `normalize()`.
    2. Keep pure unit tests for speed, but add a smoke test in `test_main.py` or a new `test_adapter_auth_integration.py`.
  Confidence: 82%

---

**suggestion(non-blocking): `normalize()` `is_admin=False` default is a footgun for future adapter authors**
  `src/lyra/adapters/telegram_normalize.py:normalize()`, `src/lyra/adapters/discord_normalize.py:normalize()`
  -- security-auditor
  Root cause: Both `normalize()` functions accept `is_admin: bool = False`. A future adapter that forgets to pass `is_admin` will silently drop admin status for all users. The current Telegram/Discord call sites are correct, but the default creates a pit-of-failure pattern. Alternatively, accepting an `Identity` object directly would eliminate this class of error.
  Solutions:
    1. Accept `identity: Identity` as a parameter in `normalize()` and extract `trust_level` + `is_admin` internally — one object, no split.
    2. Add a comment on both signatures: `# REQUIRED: always pass is_admin=identity.is_admin — do not rely on default`.
  Confidence: 70%

**suggestion(non-blocking): `_auth` field typed as `AuthMiddleware` in both adapters — obscures the canonical type**
  `src/lyra/adapters/telegram.py`, `src/lyra/adapters/discord.py`
  -- architect
  Root cause: `self._auth: AuthMiddleware = auth`. `AuthMiddleware` is an alias for `Authenticator`, so pyright accepts this, but the adapter now internally calls `_auth.resolve()` — an `Authenticator` API. When `auth.py` is eventually removed, these annotations will need updating. Declaring the dependency on `Authenticator` now makes the adapter's actual dependency explicit.
  Solutions:
    1. Change annotations to `Authenticator` and import from `lyra.core.authenticator`.
    2. Add a comment: `# AuthMiddleware == Authenticator — update annotation when auth.py facade is retired`.
  Confidence: 88%

**suggestion(non-blocking): voice path drops `is_admin` before `normalize_audio` — asymmetric with text path**
  `src/lyra/adapters/telegram_inbound.py:handle_voice_message`
  -- backend-dev · architect
  Root cause: `handle_voice_message` correctly calls `resolve()` + `guard_chain.run()`, but then calls `normalize_audio(..., trust_level=identity.trust_level)` without `is_admin`. `InboundAudio` has no `is_admin` field (spec defers this), so there is no data loss today. But the gap is invisible at the call site — when `InboundAudio.is_admin` is added in the follow-on issue, this call site will silently default to `False`.
  Solutions:
    1. Add a comment: `# is_admin not propagated — InboundAudio.is_admin deferred to audio pipeline guard issue`.
    2. No action now — address when InboundAudio is extended.
  Confidence: 78%

**suggestion(non-blocking): `require_admin` dual-path divergence should be tracked for removal**
  `src/lyra/core/builtin_commands.py:35-48`
  -- backend-dev
  Root cause: The `admin_user_ids` parameter path and the `msg.is_admin` path are not equivalent (different semantics, see blocker above). Keeping both creates ongoing maintenance burden and misuse risk. All callers within this PR use the 1-arg form — the dual-path is only alive for external callers.
  Solutions:
    1. Add a `# TODO: remove admin_user_ids param in #NNN` comment on the function signature now.
    2. Emit `DeprecationWarning` when `admin_user_ids is not None` to encourage migration.
  Confidence: 85%

---

### Nitpicks

**nitpick: `check()` docstring says "backward-compat alias for resolve()" — inverted relationship**
  `src/lyra/core/authenticator.py:120-121`
  -- backend-dev
  Root cause: Both `check()` and `resolve()` delegate to `_resolve_trust()` independently — `check` is not an alias for `resolve`. The docstring should read: `"Backward-compat method returning TrustLevel only. Prefer resolve() for new code."`.
  Confidence: 90%

**nitpick: `admin.py` deprecation docstring missing tracking issue number**
  `src/lyra/core/admin.py:1-12`
  -- architect
  Root cause: Without a reference issue, the module lingers indefinitely.
  Solutions:
    1. Add `"Removal tracked in #NNN."` once the follow-on issue is created.
  Confidence: 90%

**nitpick: `_ALLOW_ALL` sentinel has no comment noting `is_admin=False` for all users**
  `src/lyra/core/authenticator.py:268-271`
  -- architect
  Root cause: Test authors may construct `_ALLOW_ALL`-based fixtures expecting admin behavior. Spec acknowledges this; a code comment would prevent confusion.
  Solutions:
    1. Add: `# Note: resolve() always returns is_admin=False. For admin identity in tests, use admin_user_ids or trust=OWNER directly.`
  Confidence: 85%

---

### Praise

**praise: Clean separation of concerns across the three new files**
  `src/lyra/core/identity.py`, `src/lyra/core/guard.py`, `src/lyra/core/authenticator.py`
  -- architect · backend-dev
  Root cause: `identity.py` is a pure value object with no upward dependencies. `guard.py` depends only on `identity.py` and `trust.py`. `authenticator.py` is the single locus of trust resolution logic. Dependency direction is clean and consistent.
  Confidence: 97%

**praise: `Guard` as `@runtime_checkable` Protocol is the right choice for this extension point**
  `src/lyra/core/guard.py:22-28`
  -- architect
  Root cause: Enables `isinstance(obj, Guard)` in future adapter registries without requiring base class inheritance. `BlockedGuard` correctly duck-types without explicit inheritance.
  Confidence: 95%

**praise: `Rejection` as frozen dataclass with "presence = signal" semantics correctly encodes the protocol contract**
  `src/lyra/core/guard.py:18-20`
  -- architect
  Root cause: The frozen dataclass with `reason: str` gives future guards a structured place for audit log context without changing the interface.
  Confidence: 97%

**praise: `auth.py` facade is minimal and correct — 11 lines, `__all__` explicit, `noqa: F401` annotations correct**
  `src/lyra/core/auth.py`
  -- architect
  Root cause: `TestFacade.test_auth_middleware_is_authenticator` asserts object identity (`is`), not just name equality — the right assertion.
  Confidence: 97%

**praise: `test_short_circuits_on_first_rejection` uses a `NeverCalled` spy guard and asserts `never.called is False` — the most important GuardChain behavioral guarantee is verified**
  `tests/core/test_guard.py`
  -- backend-dev
  Confidence: 95%

**praise: `test_blocked_user_denied_even_public_command` pins the critical BLOCKED-before-public-command ordering invariant**
  `tests/core/test_authenticator.py`
  -- tester
  Confidence: 100%

---

### Summary

| Category | Count |
|----------|-------|
| 🔴 Blockers (`issue:`) | 6 |
| 🟠 Suggestions (blocking) | 1 |
| 🟡 Suggestions (non-blocking) | 4 |
| 🔵 Nitpicks | 3 |
| ✅ Praise | 6 |

**Verdict: 🔴 Request changes**

The structural separation is sound and all 14 spec criteria are met in the implementation. However three findings must be addressed before merge:

1. **`admin_user_ids` not threaded from startup config** — `is_admin` via explicit IDs silently broken in production for all TRUSTED admins
2. **`CommandRouter._admin_user_ids` dead code** — operators believe they control the gate but the field does nothing
3. **Adapter tests mock `check()` not `resolve()`** — guard wiring is untested; blocked-user tests pass for the wrong reasons

Findings 4–6 (require_admin empty-set trap, is_admin on public command, SC-7 not asserted) are also blocking and should be resolved in the same pass.

PR #464, opened 2026-03-30
fix(commands): guard /help processor listing on driver availability

### LINKED ISSUE #359: feat: guard session command registration on driver availability
## Problem

`_register_session_commands()` in `simple_agent.py` always registers `/vault-add`, `/explain`, and `/summarize` at agent init time, regardless of whether the agent is using the `anthropic-sdk` backend.

As a result:
- These commands always appear in `/help` output for every agent
- When a user invokes one on a non-anthropic-sdk agent, they get: `"Session commands require anthropic-sdk backend."` — with no indication of why or what to do
- This is confusing UX: the command is advertised but silently unusable

## Expected behavior

Session commands should either:
- Only be registered when `_session_driver` is not None (i.e. anthropic-sdk backend is active), so they don't appear in `/help` when unavailable
- Or display a cleaner error message that names the agent's current backend

## Possible fix

In `_register_session_commands()`, check `self._provider` capabilities before registering:

```python
def _register_session_commands(self) -> None:
    if self._provider is None:
        return
    # ... register commands
```

Or conditionally hide them from `/help` when `_session_driver is None`.

## Context

Discovered during architecture review of the `/vault-add` Discord flow.

MickaelV0:
## Summary
- Processor commands (e.g. `/vault-add`, `/explain`, `/summarize`) were always shown in `/help` output regardless of whether the agent's `_session_tools` built successfully
- Pass the router's passthroughs set to `help_command()` so processor commands only appear when the agent actually registered them as passthroughs

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #359: guard session command registration on driver availability | Open |
| Implementation | 1 commit on `feat/359-guard-session-command-registration` | Complete |
| Verification | Tests ✅ (2164 passed) | Passed |

## Test Plan
- [ ] Agent with `anthropic-sdk` backend: `/help` shows processor commands (`/vault-add`, `/explain`, etc.)
- [ ] Agent with `claude-cli` backend where `SessionTools` build fails: `/help` does NOT show processor commands
- [ ] New test `test_help_hides_processor_commands_without_passthroughs` covers both cases

Closes #359

---

### DIFF
--- a/artifacts/analyses/282-60-merged-epic-status.mdx
+++ b/artifacts/analyses/282-60-merged-epic-status.mdx
@@ -1,0 +1,113 @@
+---
+title: "Merged Epic Status: Hub standalone + NATS (#282 + #60)"
+issues: [282, 60]
+status: active
+date: 2026-03-30
+---
+
+## Decision
+
+**Merge epic #282 (Hub as standalone process) into epic #60 (NATS introduction).**
+
+Rationale: #282 planned Unix socket IPC as a stepping stone, but NATS is imminent. The Unix socket protocol (#283) would be throwaway work — NATS replaces it entirely. The real value of #282 is the **process split** (Hub standalone), which is needed regardless of transport. By merging, we skip the intermediate Unix socket layer and go straight to NATS.
+
+## What Exists Today
+
+### Architecture (ADR-021, Phase 1b — shipped)
+- Adapters run as separate supervisor processes (`lyra_telegram`, `lyra_discord`)
+- Each process embeds its own Hub instance (hub-per-adapter model)
+- `--adapter telegram|discord|all` flag controls which adapters start
+- SQLite shared via WAL mode (auth.db, vault, turns)
+- Independent restartability of adapters — but Hub state (pools, L0) lost on restart
+
+### Bus Abstraction (#48 — CLOSED)
+- `Bus[T]` Protocol extracted in worktree #432 (event-bus-pipeline-telemetry)
+- `LocalBus` implements `Bus[T]` for in-process queues
+- `PipelineEventBus` for fire-and-forget telemetry fan-out
+- **This is the key abstraction** — `NatsBus` will implement `Bus[T]` next
+
+### Supervisor Config
+- `supervisor/conf.d/lyra_telegram.conf` — `python -m lyra --adapter telegram`
+- `supervisor/conf.d/lyra_discord.conf` — `python -m lyra --adapter discord`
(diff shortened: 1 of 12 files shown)


MickaelV0:
## Code Review

**Verdict: Approve with comments** — no blockers, 2 substantive findings worth addressing.

### Warnings

**1. `command_metadata()` leaks processor commands regardless of passthrough state** (C: 75%)
`command_router.py:168-176` — Platform command sync (Telegram/Discord) calls `command_metadata()` which returns all processor commands unconditionally, bypassing the new `/help` filter. An agent where `_session_tools` failed still exposes commands in the platform command picker.
- **Fix:** Apply the same `_passthroughs` filter to `command_metadata()`.

**2. `passthroughs=None` collapses to empty frozenset — should mean "show all"** (C: 62%)
`builtin_commands.py:63` — `_active_passthroughs = passthroughs or frozenset()` treats `None` (default) identically to an empty set. A future caller omitting the arg silently hides all processor commands. `None` should mean "unfiltered" for backward compat.
- **Fix:** `if passthroughs is not None:` guard instead of `or frozenset()`.

### Test Coverage Gaps

**3.** `passthroughs=None` default path untested (existing `test_help_includes_session_commands_section` omits the arg but doesn't assert processor commands absent)
**4.** Integration path (router dispatch `/help` → `help_command` with passthroughs) untested
**5.** Only one processor command sampled — doesn't verify others are hidden when not in passthroughs

### Suggestions

- Use a hermetic fake processor registration instead of `reload_processors()` to decouple from production submodules
- Scope registry mutation with teardown/fixture

---
*Reviewed by security-auditor + tester agents via `/code-review`*

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 5 finding(s)
**Deferred:** 0
**Skipped:** 0

### Applied (1b1)

1. **`command_metadata()` leak** — Applied same `_passthroughs` filter to `command_metadata()` so platform command sync (Telegram/Discord) respects the guard
2. **`passthroughs=None` semantics** — Changed from `or frozenset()` to `if passthroughs is not None` guard, so `None` means "show all" (backward compat)
3. **`None` default path test** — New test verifies `passthroughs=None` shows all processor commands
4. **Integration test** — New async test dispatches `/help` through router, verifies passthroughs flow end-to-end
5. **Non-registered commands hidden** — Test registers two fake processors, verifies only the one in passthroughs appears (other is hidden)

Also addressed suggestions: hermetic fake processor registration with `try/finally` teardown instead of `reload_processors()`.

PR #551, opened 2026-04-05
feat(nats): add schema_version to wire format (#530)

### LINKED ISSUE #530: feat(nats): add message versioning to wire format
No version field on NATS messages. Additive schema changes degrade gracefully; breaking changes require simultaneous deploy of all 3 processes. Low urgency until next breaking schema change.

MickaelV0:
## Summary

Closes #530

Adds a `schema_version: int = 1` field to all 5 hub↔adapter NATS envelopes so mismatched producer/consumer pairs fail **loudly** (drop + `log.error` + in-process counter) instead of silently misinterpreting fields.

The issue is not about enabling rolling deploys across breaking schema changes — coordinated deploy of `lyra_hub` + `lyra_telegram` + `lyra_discord` is still required. Versioning exists to make failures **explicit** when someone forgets to coordinate.

## What changed

- **New envelope field.** `schema_version: int = 1` on `InboundMessage`, `InboundAudio`, `OutboundMessage`, `TextRenderEvent`, `ToolSummaryRenderEvent`. Colocated `SCHEMA_VERSION_*` module-level constants are the single source of truth for bumping.
- **New helper.** `src/lyra/nats/_version_check.py` exports `check_schema_version(payload, *, envelope_name, expected, subject, counter)` — pure function with a **caller-owned** counter (no static class state, no test-isolation bugs).
- **Inbound path.** `NatsBus._make_handler.handler()` now `json.loads` → `check_schema_version` → `deserialize_dict` (no double-parse; uses the existing `_serialize.py:54` helper). Instance counter `self._version_mismatch_drops` + public `version_mismatch_count(envelope_name)` accessor.
- **Outbound path.** `NatsRenderEventCodec.decode()` gains a `counter` kwarg. The check runs **before** the manual `payload.get()` extraction in the `tool_summary` branch — this was a flagged landmine in the spec and is tested explicitly.
- **Counter propagation.** `NatsOutboundListener` owns `self._version_mismatch_drops` and passes it through `decode_stream_events(..., counter=...)` to `NatsRenderEventCodec.decode(..., counter=...)`.
- **Docs.** `docs/ARCHITECTURE.md` gains a "Schema versioning" subsection under the NATS wire format section, with the forward-compat rule and a numbered bump procedure.

## Forward-compat rule

A receiver accepts any payload where `schema_version <= expected`. Strictly-greater versions are dropped. Legacy payloads (no field) default to version 1 — so existing wire traffic is never dropped.

| Case | Behavior |
|---|---|
| `schema_version` absent (legacy) | Decoded as v1 |
| `schema_version == expected` | Decoded normally |
| `schema_version < expected` (receiver newer) | Decoded normally (additive fields already degrade) |
| `schema_version > expected` (receiver older) | Dropped + `log.error` + counter++ |
| Non-int / null / ≤0 | Dropped + `log.error` + counter++ |

## Test plan

- [x] Helper unit tests cover 8 cases: absent, match, receiver-newer, receiver-older, string, null, 0/-1, counter isolation
- [x] Legacy round-trip test parametrized over all 5 envelope types (serialize → strip `schema_version` → deserialize → default == 1)
- [x] `NatsBus` integration tests (real nats-server): match, legacy, mismatch, mixed batch `[v1, v2_bad, v1]` survives subscription, non-int drops
- [x] `NatsRenderEventCodec.decode()` unit tests for text + tool_summary branches, including the landmine test that passes a malformed `files` value (would raise if the check ran after extraction)
- [x] `NatsOutboundListener` integration test: counter flows through `decode_stream_events` → codec → listener accessor
- [x] Full `tests/nats/` + `tests/adapters/` sweep: 490 passed, 35 skipped (nats-server tests skip cleanly when binary absent), 0 failures, 0 regressions
- [x] `ruff check` and `pyright` clean on all touched files
- [x] `docs/ARCHITECTURE.md` gains the subsection + numbered bump procedure

## Out of scope

Chunk-envelope wrapper versioning (hand-rolled `{stream_id, seq, event_type, payload, done}` in `render_event_codec.py`), NATS subject-pattern versioning, TTS/STT wire format (voiceCLI boundary), schema registry, version negotiation, prometheus export.

## Artifacts

- Frame: `artifacts/frames/530-nats-message-versioning-frame.mdx`
- Spec:  `artifacts/specs/530-nats-message-versioning-spec.mdx`
- Plan:  `artifacts/plans/530-nats-message-versioning-plan.mdx`

### DIFF
--- a/artifacts/frames/530-nats-message-versioning-frame.mdx
+++ b/artifacts/frames/530-nats-message-versioning-frame.mdx
@@ -1,0 +1,133 @@
+---
+title: Add message versioning to NATS wire format
+issue: 530
+status: approved
+tier: F-lite
+date: 2026-04-05
+---
+
+## Problem
+
+Lyra runs as three separate processes (`lyra_hub`, `lyra_telegram`,
+`lyra_discord`) that communicate exclusively over NATS using JSON-encoded
+dataclasses (`InboundMessage`, `InboundAudio`, `OutboundMessage`, and the
+render-event stream). The wire format has **no version field**.
+
+Today, the serializer in `src/lyra/nats/_serialize.py` is lenient on missing
+fields (`_decode_dataclass` falls through to dataclass defaults at line 272),
+so **additive** schema changes degrade gracefully — a v2 receiver reading a v1
+payload just sees the new field at its default.
+
+**Breaking** schema changes (renaming a field, changing a type, removing a
+field, semantic shifts) are a different story: they require the hub and both
+adapter processes to be redeployed **simultaneously**, because a mismatched
+producer/consumer pair will silently misinterpret the payload with zero signal
+in logs or metrics. The existing `handler` error path in `nats_bus.py:242`
+only catches `deserialize` exceptions — it doesn't notice when JSON decodes
+cleanly but the fields mean something different than the receiver expects.
+
+The immediate goal is not to *enable* rolling deploys across breaking changes
+(that still requires coordination), but to make the failure mode **explicit
+and loud** when someone forgets to coordinate: the receiver should detect the
+mismatch, drop the message, log an ERROR, and emit a metric — not silently
+proceed with corrupt data.
+
+## Who
+
+- **Primary:** Lyra maintainers deploying schema changes across the
+  hub + adapter boundary. Today this means anyone touching `core/message.py`,
+  `core/render_events.py`, or the NATS serializer.
+- **Secondary:** Future contributors evolving the wire format who need a
+  ratchet to signal "this change is breaking — bump the version."
+- **Secondary:** On-call — a clear `schema_version mismatch` log line points
+  directly at a botched deploy instead of a vague "messages disappeared."
+
+## Constraints
+
+- **Backwards compatible on the wire.** Legacy (unversioned) payloads must
+  still decode — `schema_version` defaults to `1` when absent, and existing
+  NATS messages in flight during the rollout must not be dropped.
+- **One field, per envelope.** Each top-level NATS envelope dataclass carries
+  its own `schema_version: int = 1`. No shared wrapper envelope (preserves the
+  current `_serialize.py` contract; minimizes diff surface).
+- **Field name: `schema_version`.** Not `version` — avoids ambiguity with
+  potential domain fields and is self-documenting at the JSON-dump level.
+- **Localized version constants.** Each envelope module exposes a
+  module-level `SCHEMA_VERSION_<ENVELOPE>` constant (e.g.
+  `SCHEMA_VERSION_INBOUND_MESSAGE = 1` in `core/message.py`). Bumping a
+  version is a single-line edit.
+- **Forward-compat rule: `payload.schema_version <= receiver.known`.**
+  A receiver accepts any version ≤ its own known version (old → new is fine,
+  since additive fields already degrade). It **drops strictly greater**
+  versions with an ERROR log and a metric increment.
+- **Reuse existing error path.** Mismatch handling lives inside the
+  `deserialize` → `handler` path in `nats_bus.py` and `render_event_codec.py`
+  — same drop-and-log pattern already used for decode failures. No new
+  top-level error plumbing.
+- **Scope: Lyra hub↔adapter envelopes only.** `InboundMessage`,
+  `InboundAudio`, `OutboundMessage`, and the render-event payloads
+  (`TextRenderEvent`, `ToolSummaryRenderEvent`). TTS/STT client messages
+  (`nats_tts_client.py`, `nats_stt_client.py`) live across the voiceCLI
+  boundary and are versioned separately by that repo.
+- **Observability.** Each drop emits a log line with
+  `(platform, bot_id, envelope_type, payload_version, receiver_version)` and
+  increments a counter metric (via whatever metrics plumbing already exists
+  in the adapter/hub process).
+
+## Out of Scope
+
+- **Rolling deploys across breaking changes.** Coordinated deploy is still
+  required; versioning only surfaces failures faster. A future issue may
+  introduce dual-reader support, but not this frame.
+- **Chunk-envelope versioning.** The hand-rolled wrapper in
+  `render_event_codec.py` (`{stream_id, seq, event_type, payload, done}`) is
+  itself an implicit schema. This frame versions the inner payload only.
+  Chunk-wrapper evolution is a separate concern.
+- **NATS subject-pattern versioning.** The subject layout
+  (`lyra.inbound.<platform>.<bot_id>`) is also schema; this frame does not
+  touch it.
+- **TTS/STT wire format.** Owned by voiceCLI; out of scope.
+- **Schema registry / migration tooling.** No central registry, no migration
+  scripts. Bumping is a manual integer increment tied to a code change.
+- **Version negotiation / downgrade.** No handshake, no "fall back to v1"
+  logic — fail fast is the policy.
+- **Shared wrapper envelope migration.** Not a refactor of the serializer to
+  introduce `{v, type, payload}` wrapping. If we ever want that, this
+  frame's `schema_version` field is the ratchet that makes the migration
+  safe.
+
+## Complexity
+
+**Tier: F-lite** — Clear scope, single domain (NATS wire format), no new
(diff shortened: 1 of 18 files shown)


MickaelV0:
## Code Review

**Reviewers:** security-auditor · architect · product-lead · tester (4 fresh agents, parallel, no implementation context).

**Verdict: Approve with comments.** All 14 success criteria from `artifacts/specs/530-nats-message-versioning-spec.mdx` are met with direct evidence. No security blockers, no correctness bugs, CI green. Three defensive/test-hygiene findings worth addressing before or after merge.

---

### Blockers (3)

**1. `_ENVELOPE_VERSIONS` silent fallback accepts unknown types** — `src/lyra/nats/nats_bus.py:255`

```python
envelope_name, expected = _ENVELOPE_VERSIONS.get(
    self._item_type, (self._item_type.__name__, 1)
)
```

Root cause: A future `NatsBus(item_type=NewEnvelope)` call with a forgotten mapping update will silently run the check with `expected=1` instead of failing loudly. The plan's own risk table (MT-7) called for a plain `dict[type]` lookup that raises `KeyError` on missing keys so omissions surface at test time.

Solutions:
1. `envelope_name, expected = _ENVELOPE_VERSIONS[self._item_type]` — let `KeyError` propagate (recommended). `_make_handler` is called from `start()` so the failure surfaces at startup.
2. Keep the fallback but log a WARNING on every unknown-type hit.

— architect, Confidence: 82%

---

**2. Flaky `asyncio.sleep(...)` in NatsBus drop tests** — `tests/nats/test_nats_bus.py` (`test_version_mismatch_drops`, `test_non_int_schema_version_drops`, `test_mixed_batch_survives_subscription`)

Root cause: Fixed-duration sleeps (150–200 ms) are used to let the NATS handler fire before asserting the staging queue. The project's existing pattern uses event-driven `asyncio.wait_for(bus.get(), timeout=...)`. On a slow CI runner, the sleeps could race with the handler.

Solutions:
1. Replace the drop-assertion pattern with `with pytest.raises(TimeoutError): await asyncio.wait_for(bus.get(), timeout=0.5)` — event-driven, fast on passing cases (recommended).
2. Publish `[v2_bad, v1]` and wait for the `v1` to arrive via `bus.get()` — the v1 arrival proves the v2 was already processed and dropped.

— tester, Confidence: 75%

---

**3. `core/message.py` constants lack an export guard test** — `tests/core/test_render_events.py::test_all_exports_complete` was updated this PR; `core/message.py` has no equivalent since it has no `__all__`.

Root cause: `message.py` gained 3 new `SCHEMA_VERSION_*` constants with no test asserting they exist. A silent deletion would go unnoticed.

Solutions:
1. Add a simple test in `tests/core/test_message.py` (or the nearest existing file) that imports the 3 constants by name and asserts each equals `1` (recommended).
2. Add an `__all__` to `message.py` and mirror the `test_all_exports_complete` pattern.

— tester, Confidence: 80%

---

### Warnings (6)

**4. Log-flood DoS on version mismatch** — `src/lyra/nats/_version_check.py` (`_drop` function)

A flood of v2 payloads produces one unthrottled `log.error` per message. Subscription stays alive (correct), but log volume itself can DoS aggregation.

Solutions:
1. Per-envelope token bucket: emit full detail on first drop, sample subsequently (recommended).
2. Downgrade repeats to `log.warning` after the first per envelope.

— security-auditor, Confidence: 72%

---

**5. Outer chunk envelope unversioned — asymmetric guard** — `src/lyra/nats/render_event_codec.py` (outer wrapper `{stream_id, seq, event_type, payload, done}`)

Documented as out-of-scope in the spec (correct). But the asymmetry — inner payload guarded, outer wrapper not — is not documented in `ARCHITECTURE.md`. A future `event_type` rename or new outer field would not be caught.

Solutions:
1. Add a one-line note to the new `### Schema versioning` section in `ARCHITECTURE.md` stating the outer chunk envelope is an unversioned contract (recommended).

— architect, Confidence: 70%

---

**6. Counter propagation is sound but opaque** — `src/lyra/adapters/nats_outbound_listener.py` `_drain_stream` call site

Counter passes `listener._version_mismatch_drops` → `decode_stream_events(counter=...)` → `NatsRenderEventCodec.decode(counter=...)`. Correct (dicts are passed by reference), but a future contributor may not see the chain.

Solutions:
1. Add a one-line comment at the `_drain_stream` call site pointing back to `version_mismatch_count` (recommended).

— architect, Confidence: 73%

---

**7. SC-11 cumulative counter not tested across separate publish calls** — `tests/nats/test_nats_bus.py::TestNatsBusVersionMismatch`

`version_mismatch_count` is asserted in multiple tests but always after exactly 1 drop. A regression that resets the counter on each drop would pass all current tests.

Solutions:
1. Extend `test_mixed_batch_survives_subscription` to publish a second v2_bad and assert `count == 2`, or add a dedicated test (recommended).

— tester, Confidence: 85%

---

**8. Float `schema_version` not tested** — `tests/nats/test_version_check.py`

Spec's edge-case table mentions "float → drop". `isinstance(raw, int)` correctly rejects floats (JSON `1.0` parses as `float`). No test covers this path.

Solutions:
1. Add `1.0` to the parametrized `test_zero_or_negative_drops` (rename to `test_malformed_values_drop`) or add a dedicated parametrize entry (recommended).

— tester, Confidence: 80%

---

**9. `decode_stream_events` return type annotation** — `src/lyra/adapters/nats_stream_decoder.py` signature

`async def -> AsyncIterator["RenderEvent"]` with a `yield` body is idiomatic and accepted by mypy/pyright in non-strict mode, but strict mode prefers `AsyncGenerator[RenderEvent, None]`. Non-blocking.

— architect, Confidence: 75%

---

### Nitpick (1)

**10. `r.message` vs `r.getMessage()` inconsistency** — `tests/nats/test_nats_bus.py::test_version_mismatch_drops`

One test uses `r.message` (pre-substitution template), others use `r.getMessage()` (formatted). Both work for the current substring match but normalizing would be cleaner.

— tester, Confidence: 70%

---

### Praise

- **Version check runs before deserialization on both paths** (`nats_bus.py:258`, `render_event_codec.py:77`) — security-auditor
- **Counter key is receiver-controlled** — no dict-growth attack possible via crafted payloads — security-auditor
- **Helper function design + caller-owned counter** — no static state, no test isolation bugs, matches spec exactly — architect
- **Layer placement of `SCHEMA_VERSION_*` constants** — colocated with dataclasses for ergonomic bump, no transport leak — architect
- **ToolSummaryRenderEvent landmine correctly handled** — version check runs before manual `payload.get()` extraction — architect
- **Landmine test design** — `"files": "not-a-dict"` would raise `AttributeError` if the gate were bypassed, so a passing test proves the short-circuit — tester
- **Log assertion robustness** — all caplog assertions match on substring `"NATS schema version mismatch"` — tester
- **All 14 success criteria met with direct, traceable evidence** — product-lead

---

**Summary:** 3 blockers (defensive/test-hygiene), 6 warnings, 1 nitpick, 8 praises. Scope discipline is excellent — nothing out of spec, nothing missing. The landmine test for `ToolSummaryRenderEvent` is particularly well-designed.

MickaelV0:
## Review Fixes Applied

Commit: 5f34f0e

**Applied (8 of 10):**
- ✅ `_ENVELOPE_VERSIONS` fallback removed — unknown types now raise `KeyError` at start
- ✅ Flaky `asyncio.sleep` replaced with `asyncio.wait_for` in 3 drop tests
- ✅ Added `tests/core/test_message.py` parity guard for 3 new constants
- ✅ Added outer chunk envelope note to `docs/ARCHITECTURE.md`
- ✅ Added counter propagation comment at `_drain_stream` call site
- ✅ Extended mixed-batch test to prove cumulative counter (now asserts `count == 2`)
- ✅ Added float `schema_version` coverage via parametrize `[0, -1, 1.0, 2.0]`
- ✅ Normalized `r.message` → `r.getMessage()` in one test

**Deferred (2 of 10):**
- ⏭ Log-flood rate limiting — adds state to a pure helper; separate design concern, tracked as follow-up
- ⏭ `decode_stream_events` AsyncGenerator annotation — cosmetic, mypy strict-mode only

**Verification:**
- 532 passed, 35 skipped, 0 failed (up from 522)
- ruff clean
- pyright 0 errors
- `nats_bus.py` 295 lines · `nats_outbound_listener.py` 300 lines (cap: 300) ✓

PR #539, opened 2026-04-04
feat(nats): add queue groups to prevent duplicate delivery

MickaelV0:
## Summary
- Add NATS queue groups to `NatsBus` and `NatsOutboundListener` subscriptions so only one instance per role receives each message during rolling deploys
- Prevents double-sends to users when supervisor briefly runs two adapter/hub instances during restarts
- Follows pattern already established in `stt_adapter_standalone.py` / `tts_adapter_standalone.py`

## Changes
- **`NatsBus`**: new `queue_group: str = ""` kwarg, passed to `nc.subscribe(queue=...)`
- **`NatsOutboundListener`**: new `queue_group: str = ""` kwarg, passed to `nc.subscribe(queue=...)`
- **Hub bootstrap** (`hub_standalone.py`, `unified.py`): `hub-inbound` / `hub-inbound-audio` groups
- **Adapter bootstrap** (`adapter_standalone.py`): `adapter-outbound-{platform}-{bot_id}` groups
- **Tests**: verify `queue=` param passed to `nc.subscribe()` for both classes

Adapter-side `NatsBus` instances are intentionally left unchanged — they publish only (their own subscriptions are unused) and adding them to the hub's queue group would cause message loss.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #527: feat(nats): add queue groups to prevent duplicate delivery | OPEN |
| Frame | [527-queue-groups-duplicate-delivery-frame.mdx](artifacts/frames/527-queue-groups-duplicate-delivery-frame.mdx) | Present |
| Spec | [527-queue-groups-duplicate-delivery-spec.mdx](artifacts/specs/527-queue-groups-duplicate-delivery-spec.mdx) | Present |
| Plan | [527-queue-groups-duplicate-delivery-plan.mdx](artifacts/plans/527-queue-groups-duplicate-delivery-plan.mdx) | Present |
| Implementation | 1 commit on `feat/527-queue-groups` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (4 new, 2442 total passed) | Passed |

## Test Plan
- [ ] Unit tests verify `queue=` is passed to `nc.subscribe()` for both `NatsBus` and `NatsOutboundListener`
- [ ] Single-instance mode behaves identically (queue group of 1 == no group)
- [ ] During rolling restart: start second hub/adapter instance, verify each message is delivered only once across the group
- [ ] Unified mode (`lyra start`) still works end-to-end

Closes #527

---
[...]

Issue #527, opened 2026-04-04
feat(nats): add queue groups to prevent duplicate delivery


MickaelV0:
No queue group configured on outbound subscriptions. During rolling deploy, supervisor may briefly run two adapter instances → both receive every outbound message → double sends.

**Needs:** test NATS queue subscription behavior with supervisor restarts.

Issue #74, opened 2026-03-06
Epic: Voice pipeline — TTS + STT integration


MickaelV0:
## Goal

Give Lyra a voice — both output (TTS) and input (STT) — starting with Telegram.

## Context

- Hub is stable (Phase 1 done) — P2 prerequisite met
- VoiceCLI (Qwen TTS) is already installed on Machine 1
- Exploration doc exists: `artifacts/explorations/03-voice-pipeline.md`
- VRAM budget allows it: TTS ~5GB fits within 10GB budget

## Scope

| # | Issue | Priority |
|---|-------|----------|
| 1 | Voice TTS in Telegram — voicecli integration | P2 |
| 2 | Voice STT — audio transcription (Whisper/faster-whisper) | P2, after TTS |
| 3 | Discord voice channel POC | P3, after STT |

## Architecture

```
User audio → STT (Whisper) → text → Agent → text → TTS (VoiceCLI) → audio → User
```

## Dependencies

- Depends on: Phase 1b (direct SDK agent) for clean integration
- Exploration: `artifacts/explorations/03-voice-pipeline.md`

MickaelV0:
All children completed: #79, #232, and 2 previously closed. Epic done.

PR #553, opened 2026-04-05
fix(nats): publish-only mode for adapter-side NatsBus (#541)

### LINKED ISSUE #541: fix(nats): adapter-side NatsBus creates unused inbound subscriptions
## Problem

In `src/lyra/bootstrap/adapter_standalone.py` (Telegram lines 93-108, Discord lines 227-243), each adapter process creates its own `NatsBus` instances and calls `.register()` + `.start()`. `.start()` creates real NATS subscriptions to `lyra.inbound.{plat}.{bid}` — but the adapter only uses these buses to `put()` outbound publishes. The incoming subscription messages accumulate in the staging queue that is never drained.

## Impact

- **Not user-facing:** The hub still delivers each inbound message exactly once via its own `hub-inbound` queue group (see #527 / PR #539). Users do not see duplicates.
- **Resource waste:** Every inbound message is received twice — once by the hub (consumed) and once by each adapter process (dropped into an unread staging queue).
- **Spurious warnings:** The adapter-side staging queue has a `maxsize=500` default. Under sustained load, the queue fills up and logs `NatsBus staging queue full — dropping message` warnings that are misleading (the hub is not losing anything).
- **Rolling deploy amplification:** During a rolling restart, N adapter instances each independently subscribe and waste N copies.

## Context

Surfaced during code review of #527 (PR #539) by two independent reviewers (backend-dev C=95, architect C=75). Explicitly deferred from that PR because it sits outside the original "duplicate delivery to users" scope.

## Options

1. **Remove `.start()` on adapter-side `NatsBus` instances.** The adapter only needs `put()` to publish. Refactor `NatsBus` to allow publish-only mode, or split into a publisher-only helper class.
2. **Add a distinct `adapter-inbound-{plat}-{bid}` queue group** so only one adapter instance wastes a copy (not N). Easier but still wasteful.
3. **Document the behavior as intentional** and silence the staging-queue-full warnings for publish-only buses.

Option 1 is the correct long-term fix — the subscription is purely accidental.

## Acceptance

- [ ] Adapter-side `NatsBus` instances no longer create unused inbound subscriptions
- [ ] Hub-side `NatsBus` behavior unchanged (still consumes via `hub-inbound` group)
- [ ] No regression in rolling-deploy duplicate-delivery prevention from #527
- [ ] Tests confirm publish-only buses do not subscribe

Related: #527, #539

MickaelV0:
## Summary
- Adds `publish_only: bool = False` kwarg to `NatsBus`: `start()` / `stop()` become no-ops, `get()` raises `RuntimeError`, `put()` and `register()` unchanged
- Wires `publish_only=True` at 4 adapter construction sites in `adapter_standalone.py` (Telegram + Discord × inbound + inbound_audio)
- Eliminates accidental `lyra.inbound.*` subscriptions on adapter processes that filled an unread staging queue and logged misleading "staging queue full" warnings under load

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #541: fix(nats): adapter-side NatsBus creates unused inbound subscriptions | OPEN |
| Frame | [541-adapter-natsbus-unused-subs-frame.mdx](artifacts/frames/541-adapter-natsbus-unused-subs-frame.mdx) | Present |
| Spec | [541-adapter-natsbus-unused-subs-spec.mdx](artifacts/specs/541-adapter-natsbus-unused-subs-spec.mdx) | Present |
| Plan | [541-adapter-natsbus-unused-subs-plan.mdx](artifacts/plans/541-adapter-natsbus-unused-subs-plan.mdx) | Present |
| Implementation | 1 commit on `feat/541-adapter-natsbus-unused-subs` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (5 new) | Passed |

## Context

Surfaced during code review of #527 / PR #539 by two independent reviewers and explicitly deferred. Not user-facing — hub already dedupes via `hub-inbound` queue group. This is a resource-waste + misleading-warnings fix, hardened into a self-describing constructor flag so future adapter sites can't repeat the mistake.

## Test Plan
- [ ] CI runs full `tests/nats/` suite against a real `nats-server` — locally `nats-server` requires sudo, so 5 new tests skip via `requires_nats_server` marker; verified by mock smoke test that `subscription_count == 0`, `nc.subscribe` uncalled, and `get()` raises with `"publish-only"` in message
- [ ] Verify adapter process logs on production do not emit "staging queue full" warnings after deploy
- [ ] Verify `NatsBus.subscription_count` on adapter-side instances stays at 0 (vs. 2 previously per adapter bot)
- [ ] Regression: hub-side \`NatsBus\` (publish_only=False default) still delivers via \`hub-inbound\` queue group (#527 dedup intact) — 2463 passed / 65 skipped locally

Closes #541

---

### DIFF
--- a/src/lyra/bootstrap/adapter_standalone.py
+++ b/src/lyra/bootstrap/adapter_standalone.py
@@ -93,14 +93,14 @@ async def _bootstrap_adapter_standalone(  # noqa: PLR0915, C901
                 token, webhook_secret = tg_creds[bot_id]
 
                 inbound_bus: Bus[InboundMessage] = NatsBus(  # type: ignore[type-arg]
-                    nc=nc, bot_id=bot_id, item_type=InboundMessage,
+                    nc=nc, bot_id=bot_id, item_type=InboundMessage, publish_only=True,
                 )
                 inbound_bus.register(platform_enum)
                 await inbound_bus.start()
 
                 inbound_audio_bus: Bus[InboundAudio] = NatsBus(  # type: ignore[type-arg]
                     nc=nc, bot_id=bot_id, item_type=InboundAudio,
-                    subject_prefix="lyra.inbound.audio",
+                    subject_prefix="lyra.inbound.audio", publish_only=True,
                 )
                 inbound_audio_bus.register(platform_enum)
                 await inbound_audio_bus.start()
@@ -227,14 +227,14 @@ async def _bootstrap_adapter_standalone(  # noqa: PLR0915, C901
                 token = dc_creds[bot_id]
 
                 inbound_bus_dc: Bus[InboundMessage] = NatsBus(  # type: ignore[type-arg]
-                    nc=nc, bot_id=bot_id, item_type=InboundMessage,
+                    nc=nc, bot_id=bot_id, item_type=InboundMessage, publish_only=True,
                 )
                 inbound_bus_dc.register(platform_enum)
                 await inbound_bus_dc.start()
 
                 inbound_audio_bus_dc: Bus[InboundAudio] = NatsBus(  # type: ignore[type-arg]
                     nc=nc, bot_id=bot_id, item_type=InboundAudio,
-                    subject_prefix="lyra.inbound.audio",
+                    subject_prefix="lyra.inbound.audio", publish_only=True,
                 )
                 inbound_audio_bus_dc.register(platform_enum)
                 await inbound_audio_bus_dc.start()
--- a/src/lyra/core/bus.py
+++ b/src/lyra/core/bus.py
@@ -52,7 +52,11 @@ class Bus(Protocol[T]):
         ...
 
     async def get(self) -> T:
-        """Wait for and return the next item from the staging queue."""
+        """Wait for and return the next item from the staging queue.
+
+        Some implementations raise ``RuntimeError`` when the bus is configured
+        as publish-only (see ``NatsBus(..., publish_only=True)``).
+        """
         ...
 
     def task_done(self) -> None:
@@ -60,11 +64,19 @@ class Bus(Protocol[T]):
         ...
 
     async def start(self) -> None:
-        """Start the bus (spawn feeder tasks, open connections, etc.)."""
+        """Start the bus (spawn feeder tasks, open connections, etc.).
+
+        Network-backed implementations may be no-ops when configured as
+        publish-only (e.g. ``NatsBus(..., publish_only=True)``).
+        """
         ...
 
     async def stop(self) -> None:
-        """Stop the bus and clean up resources."""
+        """Stop the bus and clean up resources.
+
+        Network-backed implementations may be no-ops when configured as
+        publish-only (e.g. ``NatsBus(..., publish_only=True)``).
+        """
         ...
 
     def qsize(self, platform: Platform) -> int:
--- a/src/lyra/nats/nats_bus.py
+++ b/src/lyra/nats/nats_bus.py
@@ -80,6 +80,13 @@ class NatsBus(Generic[T]):
         await bus.stop()    # unsubscribes; registrations remain intact
         await bus.start()   # safe to restart without re-registering
 
+    When ``publish_only=True`` the bus operates in publish-only mode: ``start()``
+    and ``stop()`` become no-ops (no subscriptions are ever created), and ``get()``
+    raises ``RuntimeError``.  ``register()`` and ``put()`` continue to work as
+    normal.  This mode is intended for adapter-side buses that only publish
+    outbound messages into NATS and have no reason to open inbound subscriptions
+    back to themselves.
+
     Args:
         nc: Already-connected ``nats.NATS`` client.
         bot_id: Default bot identifier used when ``register()`` is called
@@ -88,6 +95,8 @@ class NatsBus(Generic[T]):
         subject_prefix: NATS subject prefix. Defaults to ``"lyra.inbound"``.
             Use a different prefix (e.g. ``"lyra.inbound.audio"``) to avoid
             subject collisions between different message types.
+        publish_only: When ``True``, ``start()`` and ``stop()`` are no-ops and
+            ``get()`` raises ``RuntimeError``.  Defaults to ``False``.
     """
 
     def __init__(  # noqa: PLR0913
@@ -99,6 +108,7 @@ class NatsBus(Generic[T]):
         *,
         staging_maxsize: int = 500,
         queue_group: str = "",
+        publish_only: bool = False,
     ) -> None:
         validate_nats_token(subject_prefix, kind="subject_prefix")
         validate_nats_token(queue_group, kind="queue_group", allow_empty=True)
@@ -107,6 +117,7 @@ class NatsBus(Generic[T]):
         self._item_type = item_type
         self._subject_prefix = subject_prefix
         self._queue_group = queue_group
+        self._publish_only = publish_only
         self._registrations: set[tuple[Platform, str]] = set()
         self._subscriptions: dict[tuple[Platform, str], Subscription] = {}
         self._staging: asyncio.Queue[T] = asyncio.Queue(maxsize=staging_maxsize)
@@ -150,11 +161,13 @@ class NatsBus(Generic[T]):
 
         Subject pattern: ``{subject_prefix}.{platform.value}.{bot_id}``
 
-        No-op if zero registrations exist.
+        No-op if zero registrations exist or if ``publish_only=True``.
 
         Raises:
             RuntimeError: If subscriptions are already active (double-start).
         """
+        if self._publish_only:
+            return
         if self._subscriptions:
             raise RuntimeError(
                 "NatsBus.start() called while subscriptions are already active — "
@@ -168,7 +181,11 @@ class NatsBus(Generic[T]):
 
         Registered (platform, bot_id) pairs are preserved so that a subsequent
         ``start()`` succeeds without re-registering.
+
+        No-op if ``publish_only=True``.
         """
+        if self._publish_only:
+            return
         for sub in self._subscriptions.values():
             try:
                 await sub.unsubscribe()
@@ -199,7 +216,16 @@ class NatsBus(Generic[T]):
         await self._nc.publish(subject, payload)
 
     async def get(self) -> T:
-        """Wait for and return the next item from the staging queue."""
+        """Wait for and return the next item from the staging queue.
+
+        Raises:
+            RuntimeError: If called on a publish-only bus.
+        """
+        if self._publish_only:
+            raise RuntimeError(
+                "NatsBus.get() called on a publish-only bus — "
+                "publish-only buses never consume inbound messages."
+            )
         return await self._staging.get()
 
     def task_done(self) -> None:
--- a/src/lyra/nats/nats_bus.py
+++ b/src/lyra/nats/nats_bus.py
@@ -80,23 +80,14 @@ class NatsBus(Generic[T]):
         await bus.stop()    # unsubscribes; registrations remain intact
         await bus.start()   # safe to restart without re-registering
 
-    When ``publish_only=True`` the bus operates in publish-only mode: ``start()``
-    and ``stop()`` become no-ops (no subscriptions are ever created), and ``get()``
-    raises ``RuntimeError``.  ``register()`` and ``put()`` continue to work as
-    normal.  This mode is intended for adapter-side buses that only publish
-    outbound messages into NATS and have no reason to open inbound subscriptions
-    back to themselves.
-
     Args:
         nc: Already-connected ``nats.NATS`` client.
         bot_id: Default bot identifier used when ``register()`` is called
             without an explicit ``bot_id``.
         item_type: Concrete type used for deserialization (e.g. ``InboundMessage``).
         subject_prefix: NATS subject prefix. Defaults to ``"lyra.inbound"``.
-            Use a different prefix (e.g. ``"lyra.inbound.audio"``) to avoid
-            subject collisions between different message types.
-        publish_only: When ``True``, ``start()`` and ``stop()`` are no-ops and
-            ``get()`` raises ``RuntimeError``.  Defaults to ``False``.
+        publish_only: If ``True``, ``start()`` is a no-op and ``get()`` raises.
+            For adapter-side buses that only publish (see #541).
     """
 
     def __init__(  # noqa: PLR0913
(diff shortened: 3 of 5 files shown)


MickaelV0:
## Code Review

Multi-agent review on PR #74 — 5 fresh agents (security-auditor, architect, product-lead, frontend-dev, tester), no shared implementation context.

**Spec compliance:** 28/43 criteria unambiguously MET, 13 DEFERRED (browser smoke + post-merge), 2 need spec alignment. CI is green (23s, all jobs).

---

## 🔴 Blockers (7)

### 1. `initDownloads` leaks a `document` click listener on every call
**File:** `plugins/forge/references/gallery-templates/gallery-base.js:409`
**Agents:** frontend-dev (C=92%), architect (C=80%), tester (C=88%) — **3 reviewers agree**

**Root cause:** Every call appends a new `document.addEventListener('click', ...)` with no cleanup, no guard. Called once today, but the API invites repeated calls (hot-reload, tab swap, mode-switch patterns).

**Fix:** Guard with a sentinel:
```js
if (!wrap.dataset.dlInitialised) {
  wrap.dataset.dlInitialised = 'true'
  document.addEventListener('click', (e) => {
    if (!wrap.contains(e.target)) menu.classList.remove('open')
  })
}
```

### 2. `delete btn.dataset.loading` may not remove the DOM attribute
**File:** `gallery-base.js:429` · frontend-dev (C=88%)

**Root cause:** `btn.dataset` is a `DOMStringMap` proxy. The spec does not guarantee attribute removal via property deletion. Firefox historically kept the attribute, leaving the button stuck in loading state.

**Fix:** `btn.removeAttribute('data-loading')`.

### 3. `render` bound to `window.resize` without debounce
**File:** `multi-mode-gallery.html:498` · frontend-dev (C=95%)

**Root cause:** Every pixel of resize triggers a full `innerHTML` rebuild. 200-item gallery during drag = visible jank.

**Fix:** 150ms trailing debounce:
```js
let resizeTimer
window.addEventListener('resize', () => { clearTimeout(resizeTimer); resizeTimer = setTimeout(render, 150) })
```

### 4. `buildDimFilters` line 92: `dim.label` unescaped in `innerHTML`
**File:** `gallery-base.js:92` · security-auditor (C=75%)

**Root cause:** Pre-existing code, but exposed by the new dual-API formalization which invites authors to pass `dim.label` from external data (catalogue JSON). The sibling `buildPivotSegsFromDims` correctly escapes. Inconsistency is a footgun.

**Fix:** `ctrl.innerHTML = \`<span class="ctrl-label">${escHtml(dim.label)}</span>\``

### 5. `console.error` logs raw `err` object — info disclosure
**File:** `gallery-base.js:427` · security-auditor (C=62%)

**Root cause:** If `err` is a `Response` or carries response bodies, secrets can leak to console. Toast uses `err?.message` correctly; console.error bypasses the guard.

**Fix:** `console.error('[initDownloads]', entry.id, err?.message ?? String(err))`

### 6. `{{STORE_KEY_PREFIX}}` written literally to localStorage if not substituted
**File:** `multi-mode-gallery.html:210` · tester (C=91%)

**Root cause:** `const STORE_KEY = '{{STORE_KEY_PREFIX}}'` — author opening raw HTML without substitution pollutes localStorage with literal brace syntax. No error.

**Fix:** Add CUSTOMISE comment + sensible fallback string.

### 7. Undocumented placeholders `{{TITLE_PLAIN}}`, `{{TITLE_ACCENT}}`, `{{ACCENT_DIM}}`
**File:** `multi-mode-gallery.html:86,26`, `README.md:309` · product-lead (C=92%)

**Root cause:** Template uses 8 placeholders; README says "same markers as other templates" (lists 5). First-time author hits 3 undocumented substitutions.

**Fix:** Add placeholder table to "How to customise multi-mode-gallery.html" Step 1.

---

## 🟡 Warnings (7 non-blocking)

| # | Issue | File | Agent | C |
|---|-------|------|-------|---|
| 8 | `.pixelated` vendor prefix ordering wrong (crisp-edges overrides pixelated in FF) | `gallery-base.css:102` | frontend-dev | 85% |
| 9 | `.mode-btn` missing `aria-pressed` — a11y | `multi-mode-gallery.html:418` | frontend-dev | 90% |
| 10 | `.toast` not announced by screen readers — missing `aria-live` / `role="alert"` | `gallery-base.js:360`, `css:120` | frontend-dev | 88% |
| 11 | `wireCardClicks` re-queries every render — use event delegation | `multi-mode-gallery.html:379` | frontend-dev | 90% |
| 12 | `buildDimFilters` reverses dim order via `insertBefore` | `gallery-base.js:95` | architect | 75% |
| 13 | `showToast` doesn't validate `variant` | `gallery-base.js:369` | architect | 85% |
| 14 | `mkCard` doesn't sanitize trailing slash in `item.dir` | `multi-mode-gallery.html:362` | frontend-dev | 72% |

---

## 🔵 Suggestions (5 low-priority)

- `entry.id` → `btn.id` without safeClass validation (security-auditor C=62%)
- JSZip floating `@3` version pin — supply chain risk, recommend SRI hash (security-auditor C=72%)
- `getActiveMode()` called multiple times in switchMode — semantic redundancy (architect C=85%)
- Empty blob in download handler silently downloads 0-byte file (tester C=75%)
- Rapid mode switch has no generation counter — latent if `buildItems` becomes async (tester C=65%)

---

## ✅ Praise

- **`switchMode` atomic 8-step sequence** with numbered inline comments — exactly the right documentation, prevents stale-key bug class entirely. (C=100%, praised by 3 reviewers)
- **Dual-API JSDoc** on `buildDimFilters`/`applyDimFilters` with worked examples + cross-references — exemplary documentation strategy. (C=97%)
- **`showToast` uses `textContent`** not innerHTML — safe by design. (C=100%)
- **`buildPivotSegsFromDims`** escapes both `key` and `dim.label` before innerHTML. (C=100%)
- **`showToast` lazy-creates `toast-stack`** with clean two-phase setTimeout + `.toast-leaving` — leak-resistant. (C=95%)

---

## 📋 Deferred to post-merge (13 ACs)

- **Slice 3.5** — manual browser smoke in Chromium 124+ for stale-key bug (5 ACs). Structural smoke passed in implementation.
- **Slice 6** — `./sync-plugins.sh --local`, cache propagation check, follow-up issue, CI checks (5 ACs).
- **Slice 2** — live `v20-gallery.html` functional check, console error audit (3 ACs).

---

## Verdict: **Request changes**

7 blockers need addressing before merge. All are small (1–5 line fixes), concentrated in `gallery-base.js` + `multi-mode-gallery.html`. Architectural design is sound — all praise findings are on high-value patterns — but implementation details around listener cleanup, attribute removal, debouncing, XSS escape consistency, and placeholder docs need attention.

**Recommended next step:** run `/fix` to auto-apply the 7 high-confidence blockers + the 7 warnings in a single cleanup pass.

---
Reviewed by 5 fresh agents via `/code-review`.

MickaelV0:
## Review Fixes Applied

Iteration 1 response to the code review above.

**Auto-applied (C≥80, 2+ agents or verifier-confirmed):** 4 findings
**Applied via 1b1 walkthrough:** 3 findings
**Skipped (with rationale):** 2 findings
**Failed:** 0
**Deferred:** 0

### Auto-Applied

- **[applied]** `issue:` **f1 — Double-start silent on publish-only** (tester C=92, backend-dev C=80) — Added `_started: bool` flag. `start()` now raises `RuntimeError("already-started")` on a second call regardless of `publish_only`. Symmetric with normal buses.
- **[applied]** `issue:` **f2 — register() guard neutered** (backend-dev C=85, architect verifier C=97) — `register()` guard now checks `_started` instead of `_subscriptions`, closing the silent post-start bypass. Shared root fix with f1.
- **[applied]** `issue:` **f3 — `staging_qsize() == 0` tautology** (tester C=88, backend-dev verifier C=95) — Replaced with `subscription_count == 0` assertion (the meaningful invariant) and expanded field-equality checks in both `test_publish_only_put_still_publishes` and `test_publish_only_adapter_bus_roundtrip`.
- **[applied]** `suggestion:` **f5 — stop() implicit no-op comment** (architect C=80, backend-dev verifier C=82) — Added docstring note to `stop()` documenting the publish-only invariant (_subscriptions is always empty because start() returned early) so future editors don't accidentally add teardown that bypasses it.

### Applied via 1b1

- **[applied]** `issue:` **f4 — test_stop_noop unsubscribe-spy gap** (architect C=78) — Added `TestPublishOnlyInvariants.test_stop_does_not_touch_nats` using a mock NATS client. Asserts `len(nc.method_calls)` is unchanged across `stop()` on a publish-only bus. Catches any future refactor that adds teardown bypassing the empty-`_subscriptions` invariant.
- **[applied]** `suggestion:` **f6 — `subject_prefix` docstring missing** (product-lead C=85) — Restored the clarification: "Use a different prefix (e.g. `lyra.inbound.audio`) to avoid subject collisions between different message types."
- **[applied]** `suggestion:` **f8 — partial field equality in roundtrip tests** (tester C=75) — Expanded assertions to 8 fields (id, platform, bot_id, scope_id, user_id, user_name, text, trust_level) matching the style of `test_put_get_roundtrip`.

### Skipped

- **[skipped]** `suggestion:` **f7 — adapter wiring inline comments** (architect C=75) — The `publish_only=True` kwarg is visible 2 lines above each `start()` call. Adding 4 identical comments was deemed redundant with the constructor arg. `nats_bus.py` is at the 300-line hard CI limit, and the adapter file changes in this PR are already surgical.
- **[skipped]** `suggestion:` **f9 — `PublishOnlyError` subclass** (backend-dev C=65) — Second-opinion review by devops (C=88) confirmed: (1) no exception-grouping layer exists in the stack (plain stdlib logging, supervisord stdout capture, no Sentry/Prometheus/structlog), so a named class gives zero distinct alerting bucket over plain `RuntimeError`; (2) the error is unreachable in any valid deploy — it would surface only as a coding mistake during tests, not in production logs; (3) file at hard 300-line limit, cost non-zero for zero operational gain.

### New tests

- `test_publish_only_double_start_raises` — regression for f1
- `test_publish_only_register_after_start_raises` — regression for f2
- `TestPublishOnlyInvariants.test_stop_does_not_touch_nats` — mock-based invariant check for f4

### Verification

- `ruff check`: clean
- `pytest tests/nats/`: **107 passed, 42 skipped** (42 skips are `requires_nats_server` tests, unchanged count + 2 new tests from f1/f2 regressions + 1 new invariant test from f4)
- `tools/check_file_length.sh`: **PASS** — `nats_bus.py` at 300 lines (hard limit)
- Smoke test via mock NATS: double-start raises, register-after-start raises, get() raises on publish-only, restart after stop works

### Commits

1. `daa1716` — auto-apply: f1, f2, f3, f5
2. `7b44d7f` — 1b1: f4, f6, f8

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/fix`

MickaelV0:
## Code Review — Iteration 2

**3 agents on iteration 2 delta (backend-dev, tester, architect).** Scope: commits `daa1716` + `7b44d7f` (109 insertions, 27 deletions across 3 files).

### Verdict: **Approve with comments**

The `_started` flag refactor is confirmed architecturally sound across all 3 agents. No correctness bugs, no regressions, no security concerns. All 3 `issue:` findings below are explicitly qualified "minor" by reviewers and do not block merge. The suggestions are documentation / test-hygiene improvements.

### Minor issues (non-blocking)

**i1. Missing regression test for publish-only restart after stop()** — tester (C=80)

`test_nats_bus.py` `TestPublishOnlyMode`
The `_started = False` reset in `stop()` is untested on the publish-only path. A future refactor that skips the reset on publish-only would go undetected. Add `test_publish_only_start_stop_start_ok` asserting `start() → stop() → start()` succeeds (does not raise "already-started") on a publish-only bus.

**i2. `_started = True` set before `publish_only` early-return — partial-setup window** — architect (C=75)

`src/lyra/nats/nats_bus.py` `start()`
If `_make_handler` raises on the first iteration, `_started` is `True` but `_subscriptions` is partially populated. A caller catching that exception and calling `stop()` gets a clean state (stop resets `_started=False` and clears subs). But calling `start()` again without `stop()` raises "already-started" even though setup never completed. This is the correct conservative choice but is undocumented. Add one inline comment: `# _started is set before the loop; a failed start() must be recovered via stop()`.

**i3. `test_nats_bus_multibot.py` asserts on private `_subscriptions`** — architect (C=70)

`tests/nats/test_nats_bus_multibot.py:84-86, 142, 148, 162-163, 198, 245-247`
These assertions access `bus._subscriptions` (private) as a proxy for "is started". With `_started` now being the lifecycle source of truth, a publish-only multibot bus would have `_started=True, _subscriptions={}`, creating a semantic mismatch. Recommend replacing `len(bus._subscriptions)` assertions with the public `bus.subscription_count` property. **Note:** this touches pre-existing code not changed in this PR — scope-creep candidate, could be a follow-up.

### Non-blocking suggestions

**s1.** `register()` error message lost "call stop() first" hint — `backend-dev C=75`. Consider: `f"Cannot register platform {platform!r} after start() — call stop() first."`

**s2.** Tighten f2 regression test match from `"after start"` to `"after start\\(\\)"` to anchor the exact message — tester C=70.

**s3.** Use `MagicMock(spec=NATS)` in `TestPublishOnlyInvariants.test_stop_does_not_touch_nats` to catch phantom-method calls — tester C=70.

**s4.** `stop()` docstring says "natural no-op" but now writes `_started = False` — rephrase to "loop no-op" for precision — architect C=65.

### Praise

- **architect:** "The `_started` flag is the correct abstraction — decouples lifecycle state from transport state. `register()` now guards correctly on all bus types with a single cheap boolean read, and the state machine has no degenerate states. Solid refactor."
- **backend-dev:** "`_started` placement is exactly right. The flag is set before the `publish_only` early-return — the only correct placement. Both original bugs are resolved with a single, clean mechanism. Restart-after-stop works for normal and publish-only buses. No concurrency concern."
- **tester:** "The regression coverage for f1 and f2 is mechanically sound — both tests would detect exactly the failure mode described in their docstrings. `subscription_count` replacement for `staging_qsize` is the right invariant. `TestPublishOnlyInvariants` is correctly structured as a no-server unit class. The 8-field assertion expansion is a genuine quality improvement."

### Summary

| Category | Count |
|---|---|
| `issue:` (minor) | 3 |
| `suggestion(non-blocking):` | 4 |
| Praise | many |

No blocking findings. Core refactor solid. Merge as-is is viable; alternatively, a quick iteration-2 fix for i1 (restart test) and i2 (inline comment) is small-scope and closes the real gaps.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/code-review` (iteration 2)

PR #579, opened 2026-04-06
fix(stores): add periodic WAL checkpoint and checkpoint on close

### LINKED ISSUE #502: fix(stores): add periodic WAL checkpoint and checkpoint on close
## Problem

`SqliteStore` enables WAL mode (`PRAGMA journal_mode=WAL`) but never checkpoints — not periodically, not on shutdown. With 3 long-lived processes (hub, telegram, discord) sharing DB files, SQLite's auto-checkpoint (`wal_autocheckpoint=1000`) is effectively blocked because there's almost always a reader holding the WAL open.

### Current behavior

1. `_open_db()` enables WAL + `busy_timeout=30000`
2. Long-lived `aiosqlite.Connection` per store, one per process
3. `close()` calls `aiosqlite.Connection.close()` — no explicit checkpoint
4. Auto-checkpoint blocked by concurrent readers across processes

### Observed on production (2026-04-03)

| DB | DB size | WAL size | Ratio |
|----|---------|----------|-------|
| `turns.db` | 784K | 1.2MB | **152%** |
| `message_index.db` | 220K | 410K | **187%** |
| `discord.db` | 68K | 64K | **95%** |

WALs growing larger than the DBs themselves. Manual `PRAGMA wal_checkpoint(TRUNCATE)` flushed all pages successfully.

### Potential issues

1. **WAL grows unbounded** — over weeks/months, accumulates dead weight
2. **Backup inconsistency** — `.db` file is stale without the WAL; backup tools that copy only `.db` get incomplete data
3. **Startup recovery cost** — on restart, SQLite replays the entire WAL
4. **Disk space waste** — checkpointed WAL pages are never reclaimed

## Target behavior

1. **Periodic checkpoint** — background `asyncio.Task` in `SqliteStore` runs `PRAGMA wal_checkpoint(TRUNCATE)` every 30 minutes
2. **Checkpoint on close** — `close()` runs `PRAGMA wal_checkpoint(TRUNCATE)` before `aiosqlite.close()`
3. **WAL stays bounded** — never exceeds a few MB regardless of uptime
4. **Backups are safe** — external backup cron can copy `.db` files knowing they're up to date

## Implementation notes

- Add to `SqliteStore` base class so all stores benefit
- Start the background task in `_open_db()`, cancel in `close()`
- `TRUNCATE` mode resets WAL to zero bytes (vs `PASSIVE` which leaves the file)
- Checkpoint interval could be configurable via class attribute (default 1800s)
- Log checkpoint results at DEBUG level

## Files

- `src/lyra/core/stores/sqlite_base.py` — main change

MickaelV0:
## Summary
- Add periodic WAL checkpoint background task (every 30 min) to `SqliteStore` base class — prevents WAL from growing unbounded across long-lived hub/adapter processes
- Add checkpoint-on-close so every clean shutdown flushes the WAL to the main DB file

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #502: fix(stores): add periodic WAL checkpoint and checkpoint on close | Open |
| Implementation | 1 commit on `feat/502-fix-stores-wal-checkpoint` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (9 new) | Passed |

## Test Plan
- [ ] All 9 new WAL tests pass (`tests/core/test_sqlite_base_wal.py`)
- [ ] Existing store tests pass without regression (64 total)
- [ ] On production: verify WAL files truncate to 0 bytes after restart

Closes #502

---

### DIFF
--- a/src/lyra/core/stores/sqlite_base.py
+++ b/src/lyra/core/stores/sqlite_base.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
+import sqlite3
 from pathlib import Path
 
 import aiosqlite
@@ -22,11 +24,29 @@ class SqliteStore:
             # any extra setup (cache warming, etc.)
 
     ``_open_db`` is idempotent — calling it when already connected is a no-op.
+
+    WAL management
+    --------------
+    SQLite WAL mode is enabled on every connection.  Without periodic
+    checkpointing the WAL file grows unbounded when long-lived readers prevent
+    SQLite's automatic checkpoint from running.
+
+    This class handles checkpointing in two ways:
+
+    * **On close** — ``close()`` runs ``PRAGMA wal_checkpoint(TRUNCATE)`` before
+      closing the connection so the WAL is flushed on every clean shutdown.
+    * **Periodic** — a background ``asyncio.Task`` runs the same pragma every
+      ``_wal_checkpoint_interval`` seconds (default 30 min) while the store is
+      open.  This keeps the WAL bounded during long-running processes.
     """
+
+    #: Seconds between periodic WAL checkpoints.  Override in subclass to tune.
+    _wal_checkpoint_interval: int = 1800
 
     def __init__(self, db_path: str | Path) -> None:
         self._db_path = str(db_path)
         self._db: aiosqlite.Connection | None = None
+        self._checkpoint_task: asyncio.Task[None] | None = None
 
     def _require_db(self) -> aiosqlite.Connection:
         """Return the open connection or raise RuntimeError."""
@@ -47,9 +67,56 @@ class SqliteStore:
         for stmt in ddl or []:
             await self._db.execute(stmt)
         await self._db.commit()
+        self._checkpoint_task = asyncio.get_event_loop().create_task(
+            self._run_periodic_checkpoint(),
+            name=f"wal-checkpoint:{self._db_path}",
+        )
+
+    async def _checkpoint(self) -> None:
+        """Run ``PRAGMA wal_checkpoint(TRUNCATE)`` and log the result.
+
+        Best-effort — logs and returns silently on ``OperationalError`` (e.g.
+        active transaction on the connection, or concurrent readers blocking
+        the truncation).  The WAL will be checkpointed on the next opportunity.
+        """
+        db = self._db
+        if db is None:
+            return
+        try:
+            async with db.execute("PRAGMA wal_checkpoint(TRUNCATE)") as cur:
+                row = await cur.fetchone()
+        except sqlite3.OperationalError as exc:
+            log.debug("WAL checkpoint skipped (db=%s): %s", self._db_path, exc)
+            return
+        if row is not None:
+            busy, log_pages, checkpointed = row
+            log.debug(
+                "WAL checkpoint: db=%s busy=%s log=%s checkpointed=%s",
+                self._db_path,
+                busy,
+                log_pages,
+                checkpointed,
+            )
+
+    async def _run_periodic_checkpoint(self) -> None:
+        """Background task: checkpoint WAL every ``_wal_checkpoint_interval`` s."""
+        try:
+            while True:
+                await asyncio.sleep(self._wal_checkpoint_interval)
+                await self._checkpoint()
+        except asyncio.CancelledError:
+            pass
 
     async def close(self) -> None:
-        """Close the database connection."""
+        """Checkpoint WAL and close the database connection."""
+        if self._checkpoint_task is not None:
+            self._checkpoint_task.cancel()
+            try:
+                await self._checkpoint_task
+            except asyncio.CancelledError:
+                pass
+            self._checkpoint_task = None
         if self._db is not None:
+            await self._checkpoint()
             await self._db.close()
             self._db = None
--- a/src/lyra/core/stores/sqlite_base.py
+++ b/src/lyra/core/stores/sqlite_base.py
@@ -67,7 +67,7 @@ class SqliteStore:
         for stmt in ddl or []:
             await self._db.execute(stmt)
         await self._db.commit()
-        self._checkpoint_task = asyncio.get_event_loop().create_task(
+        self._checkpoint_task = asyncio.get_running_loop().create_task(
             self._run_periodic_checkpoint(),
             name=f"wal-checkpoint:{self._db_path}",
         )
--- a/tests/core/test_sqlite_base_wal.py
+++ b/tests/core/test_sqlite_base_wal.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 import asyncio
+import sqlite3
 from pathlib import Path
+from unittest.mock import AsyncMock, patch
 
 from lyra.core.stores.sqlite_base import (
     SqliteStore,
@@ -31,6 +33,12 @@ class _SimpleStore(SqliteStore):
 class TestWalCheckpointOnClose:
     """close() must checkpoint WAL before closing the connection."""
 
+    async def test_close_without_connect_is_safe(self, tmp_path: Path) -> None:
+        store = _SimpleStore(tmp_path / "test.db")
+        await store.close()  # must not raise — _db and _checkpoint_task are None
+        assert store._db is None
+        assert store._checkpoint_task is None
+
     async def test_checkpoint_on_close_runs_without_error(self, tmp_path: Path) -> None:
         store = _SimpleStore(tmp_path / "test.db")
         await store.connect()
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Code Review

### Blockers

```
issue(blocking): asyncio.get_event_loop() is deprecated for task creation in running coroutines
  src/lyra/core/stores/sqlite_base.py:70
  -- security-auditor + tester
  Root cause: get_event_loop() from inside an async context works by accident (running loop
  happens to be current thread's loop). Python 3.10+ emits DeprecationWarning; a future asyncio
  policy change or strict enforcement would raise RuntimeError mid-connect(), leaving _db set
  but _checkpoint_task None. get_running_loop() is the guaranteed-safe API inside any coroutine.
  Solutions:
    1. Replace asyncio.get_event_loop().create_task(...) with asyncio.get_running_loop().create_task(...) (recommended)
    2. Use asyncio.ensure_future(...) — equivalent inside a running loop but less explicit.
  Confidence: 92%
```

```
suggestion(blocking): _checkpoint() catches only sqlite3.OperationalError — other DatabaseError subclasses kill the background task silently
  src/lyra/core/stores/sqlite_base.py:88
  -- security-auditor
  Root cause: sqlite3 has six other direct DatabaseError subclasses (IntegrityError, InternalError,
  NotSupportedError, DataError, ProgrammingError, InterfaceError). An unexpected DatabaseError
  propagating from _checkpoint() in the background task sets task.done()=True while _db is still
  open — defeating all future periodic checkpoints with no external signal.
  Solutions:
    1. Broaden except to sqlite3.DatabaseError and log at WARNING (not DEBUG) so operators notice
       checkpoint failures before WAL bloat becomes a problem. (recommended)
    2. Add bare except Exception in _run_periodic_checkpoint() to keep the task alive, while
       narrowing _checkpoint() to sqlite3.DatabaseError.
  Confidence: 82%
```

```
issue(blocking): No test for _checkpoint() when OperationalError is raised — the silent-suppression branch is uncovered
  src/lyra/core/stores/sqlite_base.py:88
  -- tester
  Root cause: The except sqlite3.OperationalError branch in _checkpoint() has zero test coverage.
  The docstring explicitly documents this case ("active transaction on the connection, or concurrent
  readers"). If the except-block were accidentally removed, every caller would see an unhandled
  exception in production.
  Solutions:
    1. Patch store._db.execute to raise sqlite3.OperationalError("locked"), call await store._checkpoint(),
       assert no exception propagates. (recommended)
    2. Use AsyncMock(side_effect=sqlite3.OperationalError) on the context manager returned by db.execute.
  Confidence: 95%
```

```
issue(blocking): No test for close() on a never-connected store
  tests/core/test_sqlite_base_wal.py:41
  -- tester
  Root cause: test_close_is_idempotent covers double-close after connect, not close-without-connect.
  The guard branch `if self._db is not None` in close() is untested. A regression removing the
  guard would produce AttributeError in production and go undetected by the test suite.
  Solutions:
    1. Add test_close_without_connect_is_safe: construct _SimpleStore, call await store.close() directly,
       assert no exception raised. (recommended)
    2. Fold into test_close_is_idempotent as a pre-connect close call.
  Confidence: 96%
```

```
issue(blocking): test_short_interval_checkpoint_fires does not verify the checkpoint actually ran
  tests/core/test_sqlite_base_wal.py:127
  -- tester
  Root cause: assert not store._checkpoint_task.done() confirms the infinite loop is alive, not that
  _checkpoint() was called. A bug that causes the loop to sleep-and-continue without executing
  _checkpoint() would pass this test. The test's docstring claims "checkpoint runs immediately"
  but does not measure that.
  Solutions:
    1. Patch store._checkpoint = AsyncMock(wraps=store._checkpoint) after connect, sleep, then
       assert store._checkpoint.call_count >= 1 (recommended — directly confirms execution).
    2. Verify WAL file size before and after the sleep, similar to test_wal_file_truncated_after_close.
  Confidence: 90%
```

---

### Non-blocking

```
suggestion(non-blocking): Concurrent close() calls race to double-checkpoint and double-close the connection
  src/lyra/core/stores/sqlite_base.py:110
  -- security-auditor
  Root cause: close() is not re-entrant. If two coroutines call close() concurrently, both can pass
  the `if self._db is not None` check before either sets _db = None. aiosqlite's double-close()
  on an already-stopped worker thread is undefined behaviour.
  Solutions:
    1. Use a guard flag (self._closing: bool = False) checked and set before the first await. (recommended)
    2. Wrap close() with an asyncio.Lock — only the first caller proceeds.
  Confidence: 75%
```

```
suggestion(non-blocking): Task name embeds full filesystem path, leaks layout to event-loop introspection
  src/lyra/core/stores/sqlite_base.py:72
  -- security-auditor
  Root cause: f"wal-checkpoint:{self._db_path}" includes the full absolute path (e.g. /home/mickael/.lyra/auth.db).
  asyncio.all_tasks() is accessible to any code in the same loop; paths appear in asyncio debug dumps
  and log aggregators verbatim.
  Solutions:
    1. Use only the basename: name=f"wal-checkpoint:{Path(self._db_path).name}" (recommended)
    2. Use the store class name: name=f"wal-checkpoint:{type(self).__name__}"
  Confidence: 72%
```

```
suggestion(non-blocking): test_checkpoint_task_cancelled_on_close should assert task.cancelled(), not just task.done()
  tests/core/test_sqlite_base_wal.py:57
  -- tester
  Root cause: task.done() is True for both completed and cancelled tasks. A task that finished
  normally (zero-interval exhaustion) would satisfy done() even if the cancel path regressed.
  Solutions:
    1. Add assert task.cancelled() after assert task.done(). (recommended — one line)
    2. Leave as-is; sufficient for most regressions.
  Confidence: 88%
```

```
suggestion(non-blocking): test_wal_file_truncated_after_close intermediate precondition is fragile
  tests/core/test_sqlite_base_wal.py:73
  -- tester
  Root cause: assert wal_path.exists() depends on SQLite creating the WAL file after 50 inserts —
  an implementation detail, not a guaranteed API contract. On some builds the file may not appear
  until page cache is flushed.
  Solutions:
    1. Remove the intermediate assert; rely only on the post-close wal_size == 0 assertion. (recommended)
    2. Insert a larger payload to make WAL creation near-certain and document the dependency.
  Confidence: 78%
```

---

### Praise

```
praise: CancelledError correctly awaited after cancel() — no zombie tasks
  src/lyra/core/stores/sqlite_base.py:114-117
  -- security-auditor
  Confidence: 95%
```

```
praise: _checkpoint() guards against None _db — eliminates race on post-close calls
  src/lyra/core/stores/sqlite_base.py:82-84
  -- security-auditor
  Confidence: 95%
```

```
praise: Concrete subclass pattern gives genuine integration coverage — real aiosqlite I/O flows through every assertion
  tests/core/test_sqlite_base_wal.py:17-23
  -- tester
  Confidence: 98%
```

---

### Summary

| # | Label | File:Line | C |
|---|-------|-----------|---|
| 1 | `issue(blocking)` | `sqlite_base.py:70` | 92% |
| 2 | `suggestion(blocking)` | `sqlite_base.py:88` | 82% |
| 3 | `issue(blocking)` | `sqlite_base.py:88` | 95% |
| 4 | `issue(blocking)` | `test_sqlite_base_wal.py:41` | 96% |
| 5 | `issue(blocking)` | `test_sqlite_base_wal.py:127` | 90% |
| 6 | `suggestion(non-blocking)` | `sqlite_base.py:110` | 75% |
| 7 | `suggestion(non-blocking)` | `sqlite_base.py:72` | 72% |
| 8 | `suggestion(non-blocking)` | `test_sqlite_base_wal.py:57` | 88% |
| 9 | `suggestion(non-blocking)` | `test_sqlite_base_wal.py:73` | 78% |

**Verdict: Request Changes** — 5 blocking findings (2 source, 3 test gaps). All are quick fixes.

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 0 finding(s) — no 2-agent consensus reached on remaining 5
**Skipped:** 5 finding(s) (no 2-agent consensus)
**Failed:** 0

### Auto-Applied
- [applied] `issue(blocking)`: `asyncio.get_event_loop()` → `get_running_loop()` — `sqlite_base.py:70` (92%)
- [applied] `issue(blocking)`: add `test_checkpoint_suppresses_operational_error` — uncovered `except` branch (95%)
- [applied] `issue(blocking)`: add `test_close_without_connect_is_safe` — uncovered `None` guard (96%)
- [applied] `issue(blocking)`: fix `test_short_interval_checkpoint_fires` — spy verifies `_checkpoint()` executes (90%)

### Skipped (no 2-agent consensus)
- F2: broaden `except` to `DatabaseError` — agent 1: 82%, agent 2: 55%
- F6: concurrent `close()` race — agent 1: 75%, agent 2: 18% (unrealistic trigger path)
- F7: task name leaks full path — agent 1: 72%, agent 2: 8% (single-user daemon, no threat actor)
- F8: `assert task.cancelled()` missing — agent 1: 88%, agent 2: 20% (task can only exit via cancel)
- F9: `test_wal_file_truncated` precondition — agent 1: 78%, agent 2: 65% (valid guard, not flaky)

PR #595, opened 2026-04-07
feat(persistence): session resume, /clear rotation, reply-to queue

### LINKED ISSUE #592: bug(persistence): conversation not resumed after restart, /clear broken, reply-to ignores session
## Problem

Conversation persistence is broken across three related scenarios.

### 1 — Restart does not resume session
After a Lyra restart (daemon or process), conversation history is not reloaded.
- Affects: **Discord DMs** and **all Telegram chats** (Discord threads ✅ work correctly)
- Expected: history rehydrated from store → conversation continues seamlessly
- Actual: fresh session, no prior context

### 2 — `/clear` does not create a new session
Running `/clear` should wipe current session and start a clean one.
- Expected: session ID rotated, history cleared
- Actual: no observable effect — history is cleared but same session UUID reused; old turns and new turns merged in TurnStore under same session ID

### 3 — Reply-to previous message does not load that session
When a user replies to a message from an older session, Lyra should detect the session anchor and resume that session.
- Expected: session resolved from replied-to message → history loaded
- Actual: current active session used instead — old context ignored

---

## Scope

| Context | Bug #1 (restart) | Bug #2 (/clear) | Bug #3 (reply-to) |
|---------|-----------------|-----------------|-------------------|
| Discord DM | ❌ broken | ❌ broken | ❌ broken (race) |
| Discord Thread | ✅ works | ❌ broken | ❌ broken (race) |
| Telegram (all) | ❌ broken | ❌ broken | ❌ broken (race) |

---

### Persistence architecture

Three independent stores exist and ARE correctly wired at bootstrap:

| Store | DB file | Purpose |
|-------|---------|---------|
| `TurnStore` | `turns.db` | Session metadata + turn audit trail |
| `MessageIndex` | `message_index.db` | Maps `(pool_id, platform_msg_id)` → `session_id` |
| `ThreadStore` | `discord.db` | Discord-only: thread → session mapping |

Three session-resume paths exist in `message_pipeline.py:280-412`:

| Path | Mechanism | Trigger |
|------|-----------|---------|
| 1 | `MessageIndex.resolve()` | User explicitly replies to a previous message |
| 2 | `thread_session_id` via `ThreadStore` | Discord thread context only |
| 3 | `TurnStore.get_last_session()` fallback | Any scope, but only after ≥1 prior turn stored |

---

### Bug #1 — Daemon restart → no resume

**Root cause:** No session persistence wiring for non-thread scopes (Discord DMs, all Telegram).

**Causality chain — Discord DM:**
```
Restart → PoolManager creates fresh Pool (new random session_id UUID)
→ DM arrives
→ Path 1: skipped (not a reply)
→ Path 2: skipped (platform_meta has no thread_session_id — only threads inject this)
→ Path 3: TurnStore.get_last_session(pool_id) called
    → If no prior row in pool_sessions: returns None → uses fresh session_id ❌
    → If prior row exists: returns old UUID → resume_session() called
        → BUT: only works after first message stored post-restart; on first message = None ❌
```

**Causality chain — Telegram (all chats):**
```
Restart → fresh Pool
→ Message arrives
→ Path 1: skipped
→ Path 2: skipped (telegram_inbound.py never wires thread_session_id)
→ Path 3: same TurnStore fallback, same timing issue ❌
```

Discord threads work because `discord_inbound.py:164-181` injects `thread_session_id` into `platform_meta` and wires a `_session_update_fn` callback that calls `persist_thread_session()` after each turn. This mechanism is absent for DMs and entirely absent in the Telegram adapter.

**Key files:**
- `adapters/discord_inbound.py:164-180` — DM path missing `thread_session_id` injection
- `adapters/telegram_inbound.py:59-89` — zero persistence wiring
- `core/hub/message_pipeline.py:372-412` — Path 3 fallback logic
- `core/stores/turn_store.py:224-242` — `get_last_session()` query

---

### Bug #2 — `/clear` does not rotate session

**Root cause:** `pool.reset_session()` clears history but never rotates `session_id` UUID.

**Causality chain:**
```
User types /clear
→ cmd_clear() (workspace_commands.py:71-78)
    → pool.sdk_history.clear()  ✅
    → pool.history.clear()       ✅
    → pool.reset_session()
        → _observer.reset_session_persisted()   ← only resets a persisted flag
        → _session_reset_fn()                   ← CLI backend only, ignores session_id
        → pool.session_id = UNCHANGED ❌
→ Next message uses same UUID
→ TurnStore logs pre/post-clear turns merged under same session ID
→ MessageIndex routes reply-to messages to wrong session
```

**Missing line in `pool.py:252`:**
```python
async def reset_session(self) -> None:
    self._observer.reset_session_persisted()
    self.session_id = str(uuid.uuid4())  # ← THIS LINE IS MISSING
    if self._session_reset_fn is not None:
        await self._session_reset_fn()
```

**Key files:**
- `core/workspace_commands.py:71-78` — `cmd_clear()` body
- `core/pool/pool.py:252-256` — `reset_session()` missing UUID rotation

---

### Bug #3 — Reply-to loads wrong session (race condition)

**Root cause:** When pool is busy processing a message, Path 1 (reply-to resume) is silently skipped and falls through to the current session.

**Causality chain:**
```
User sends message A → pool._current_task live (pool.is_idle = False)
User replies to a message from session S_OLD before A finishes
→ Path 1 (message_pipeline.py:305-320):
    MessageIndex.resolve() → finds S_OLD ✅
    pool.is_idle = False → resume_session(S_OLD) SKIPPED ❌
    log.info("reply-to-resume: pool busy — skipping resume") ← silent drop
→ Path 3: TurnStore.get_last_session() → returns current session S1
→ Reply routed to S1, not S_OLD ❌
```

**Key file:** `core/hub/message_pipeline.py:305-320`

**Fix direction:** Queue the resume request for when the pool becomes idle, rather than silently dropping it.

---

## Fix complexity

| Bug | File | Fix | Effort |
|-----|------|-----|--------|
| #2 `/clear` | `core/pool/pool.py:252` | Add `self.session_id = str(uuid.uuid4())` | XS |
| #3 Reply-to race | `core/hub/message_pipeline.py:305-320` | Queue resume until pool idle | S |
| #1 Discord DM | `adapters/discord_inbound.py:164-180` | Wire DM scope same as thread | S |
| #1 Telegram | `adapters/telegram_inbound.py:59-89` | Wire session persistence callback | M |

## Acceptance criteria

- [ ] After daemon restart, DM and thread sessions resume from stored history (Discord + Telegram)
- [ ] `/clear` rotates session ID and starts with empty history
- [ ] Reply-to a message from session N loads session N, not the current active session
- [ ] Reply-to while pool is busy: resume is queued, not silently dropped
- [ ] All behaviours covered by integration tests

---

## Bug #4 — Guild channel and Telegram group pool scoping (added 2026-04-07)

**Root cause:** `discord_normalize.py` applies `user_scoped()` to guild channels (non-thread), and `telegram_normalize.py` applies `user_scoped()` to all group chats. This gives each user a private pool in shared spaces — consistent with #356 intent, but inconsistent with how threads work (per-thread shared pool).

**Expected:** Guild channels and Telegram groups should use a shared pool per channel/group, same as threads. Each channel/group = one pool regardless of who sends the message.

**Actual:** Each user in a guild channel or Telegram group gets their own isolated pool.

**Causality chain — Discord guild channel:**
```
Message arrives in #general
→ discord_normalize.py: is_guild_channel = True, no thread_id
→ scope_id = user_scoped("channel:{channel_id}", user_id)
→ pool_id = hash(platform, bot_id, "channel:{channel_id}:user:{user_id}")
→ Each user → different pool ❌
```

**Causality chain — Telegram group:**
```
Message arrives in group
→ telegram_normalize.py: is_group = True
→ scope_id = user_scoped("chat:{chat_id}", user_id)
→ pool_id = hash(platform, bot_id, "chat:{chat_id}:user:{user_id}")
→ Each user → different pool ❌
```

**Fix:**
| Bug | File | Fix | Effort |
|-----|------|-----|--------|
| #4 Discord guild | `adapters/discord_normalize.py:65-67` | Remove `user_scoped()` for guild channels | XS |
| #4 Telegram group | `adapters/telegram_normalize.py:95-99` | Remove `user_scoped()` from `_make_scope_id()` | XS |

**Acceptance criteria (added):**
- [ ] Two users sending messages in the same Discord guild channel land on the same pool
- [ ] Two users sending messages in the same Telegram group land on the same pool
- [ ] `/clear` in a guild channel or Telegram group resets the shared session for that channel/group

MickaelV0:
## Summary
- Fixes conversations not resuming after Lyra restart: adapters now inject `thread_session_id` from TurnStore into inbound messages, so the pool resumes the last Claude session on first message
- Fixes `/clear` not rotating the session: `Pool.reset_session()` now calls `end_session_async` + rotates UUID + calls `start_session` on TurnStore, producing two distinct session IDs
- Fixes reply-to ignoring session: when a pool is busy, `_pending_session_id` is queued and fired after `collect()` drains instead of being silently dropped
- Removes per-user pool scoping from Discord guild channels and Telegram group/topic chats — all users in a shared space now use one pool (prerequisite for multi-user session continuity)

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #592: bug(persistence): conversation not resumed after restart, /clear broken, reply-to ignores session | Open |
| Analysis | — | Absent (F-lite) |
| Spec | [592-persistence-session-resume-spec.mdx](artifacts/specs/592-persistence-session-resume-spec.mdx) | Present |
| Implementation | 1 commit on `feat/592-persistence-session-resume` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (20 new) | Passed |

## Test Plan
- [ ] `/clear` in Telegram DM → new conversation, no history bleed
- [ ] Restart Lyra → next message resumes prior Claude session (check logs for `resume_session`)
- [ ] Reply to a bot message while it's responding → session resume fires after response completes
- [ ] Two users in the same Telegram group → both routed to the same pool
- [ ] Two users in the same Discord guild channel → both routed to the same pool
- [ ] Telegram private chat and Discord DM still get per-user pools (no regression)

Closes #592

---

### DIFF
--- a/src/lyra/adapters/discord.py
+++ b/src/lyra/adapters/discord.py
@@ -13,6 +13,7 @@ if TYPE_CHECKING:
     from lyra.adapters._shared_streaming import PlatformCallbacks
     from lyra.adapters.outbound_listener import OutboundListener
     from lyra.core.bus import Bus
+    from lyra.core.stores.turn_store import TurnStore
 
 from lyra.adapters import discord_audio  # noqa: I001
 from lyra.adapters import discord_audio_outbound
@@ -83,6 +84,7 @@ class DiscordAdapter(discord.Client, OutboundAdapterBase):
         auth: Authenticator = _DENY_ALL,
         thread_store: ThreadStore | None = None,
         watch_channels: frozenset[int] = frozenset(),
+        turn_store: "TurnStore | None" = None,
     ) -> None:
         if intents is None:
             intents = discord.Intents.default()
@@ -115,6 +117,7 @@ class DiscordAdapter(discord.Client, OutboundAdapterBase):
         self._mention_re: re.Pattern[str] | None = None  # compiled on on_ready
         self._owned_threads: set[int] = set()  # populated from ThreadStore on on_ready
         self._thread_store: ThreadStore | None = thread_store
+        self._turn_store: "TurnStore | None" = turn_store
         self._watch_channels: frozenset[int] = watch_channels
         self._thread_sessions: dict[str, tuple[str, str]] = {}
         self._vsm: VoiceSessionManager = VoiceSessionManager()
--- a/src/lyra/adapters/discord_inbound.py
+++ b/src/lyra/adapters/discord_inbound.py
@@ -195,6 +195,33 @@ async def handle_message(adapter: "DiscordAdapter", message: Any) -> None:  # no
     _meta_updates: dict[str, Any] = {}
     if _stored_session_id is not None:
         _meta_updates["thread_session_id"] = _stored_session_id
+
+    # DM session wiring: inject prior session_id + persist callback for DMs.
+    _dm_session_id: str | None = None
+    if _is_dm and adapter._turn_store is not None:
+        from lyra.core.hub.hub_protocol import RoutingKey
+        from lyra.core.message import Platform
+        _pool_id = RoutingKey(
+            Platform.DISCORD, adapter._bot_id, f"channel:{message.channel.id}"
+        ).to_pool_id()
+        try:
+            _dm_session_id = await adapter._turn_store.get_last_session(_pool_id)
+        except Exception:
+            log.exception(  # noqa: TRY401
+                "TurnStore.get_last_session failed for DM pool_id=%s", _pool_id
+            )
+    if _dm_session_id is not None:
+        _meta_updates["thread_session_id"] = _dm_session_id
+    if _is_dm and adapter._turn_store is not None:
+        _dm_ts = adapter._turn_store
+
+        async def _dm_session_update_fn(
+            _msg: InboundMessage, session_id: str, pool_id: str
+        ) -> None:
+            await _dm_ts.start_session(session_id, pool_id)
+
+        _meta_updates["_session_update_fn"] = _dm_session_update_fn
+
     _has_thread_id = hub_msg.platform_meta.get("thread_id") is not None
     if _has_thread_id and adapter._thread_store is not None:
         _ts = adapter._thread_store
--- a/src/lyra/adapters/discord_inbound.py
+++ b/src/lyra/adapters/discord_inbound.py
@@ -212,7 +212,9 @@ async def handle_message(adapter: "DiscordAdapter", message: Any) -> None:  # no
             )
     if _dm_session_id is not None:
         _meta_updates["thread_session_id"] = _dm_session_id
+    _has_thread_id = hub_msg.platform_meta.get("thread_id") is not None
     if _is_dm and adapter._turn_store is not None:
+        # DM path takes priority over thread-session persistence
         _dm_ts = adapter._turn_store
 
         async def _dm_session_update_fn(
@@ -221,9 +223,7 @@ async def handle_message(adapter: "DiscordAdapter", message: Any) -> None:  # no
             await _dm_ts.start_session(session_id, pool_id)
 
         _meta_updates["_session_update_fn"] = _dm_session_update_fn
-
-    _has_thread_id = hub_msg.platform_meta.get("thread_id") is not None
-    if _has_thread_id and adapter._thread_store is not None:
+    elif _has_thread_id and adapter._thread_store is not None:
         _ts = adapter._thread_store
         _bid, _cache = adapter._bot_id, adapter._thread_sessions
 
--- a/src/lyra/adapters/discord_normalize.py
+++ b/src/lyra/adapters/discord_normalize.py
@@ -14,7 +14,6 @@ from lyra.core.message import (
     Platform,
     RoutingContext,
 )
-from lyra.core.scope import user_scoped
 from lyra.core.trust import TrustLevel
 
 if TYPE_CHECKING:
@@ -59,12 +58,7 @@ def normalize(  # noqa: PLR0913 — all kwargs are platform-specific routing con
         else f"channel:{resolved_channel_id}"
     )
 
-    # User-scope guild channels so each user gets their own pool (#356).
-    # Threads and DMs are left as-is (single-user or thread-session-resume).
     user_id = f"dc:user:{raw.author.id}"
-    is_guild_channel = raw.guild is not None and not resolved_thread_id
-    if is_guild_channel:
-        scope_id = user_scoped(scope_id, user_id)
 
     # Detect channel type
     channel_type: str = "text"
--- a/src/lyra/adapters/telegram.py
+++ b/src/lyra/adapters/telegram.py
@@ -18,6 +18,7 @@ if TYPE_CHECKING:
     from lyra.adapters._shared_streaming import PlatformCallbacks
     from lyra.adapters.outbound_listener import OutboundListener
     from lyra.core.bus import Bus
+    from lyra.core.stores.turn_store import TurnStore
 
 from lyra.adapters import telegram_audio  # noqa: I001
 from lyra.adapters._base_outbound import OutboundAdapterBase
@@ -98,6 +99,7 @@ class TelegramAdapter(OutboundAdapterBase):
         circuit_registry: CircuitRegistry | None = None,
         msg_manager: MessageManager | None = None,
         auth: Authenticator = _DENY_ALL,
+        turn_store: "TurnStore | None" = None,
     ) -> None:
         super().__init__()  # no-op today, future-proofs cooperative chain
         if auth is not _DENY_ALL:
@@ -122,6 +124,7 @@ class TelegramAdapter(OutboundAdapterBase):
         self._msg_manager = msg_manager
         self._auth: Authenticator = auth
         self._guard_chain: GuardChain = GuardChain([BlockedGuard()])
+        self._turn_store: "TurnStore | None" = turn_store
         _raw_tmp = os.environ.get("LYRA_AUDIO_TMP") or None
         if _raw_tmp is not None:
             _tmp_path = Path(_raw_tmp)
--- a/src/lyra/adapters/telegram_inbound.py
+++ b/src/lyra/adapters/telegram_inbound.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 import logging
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
@@ -68,6 +69,35 @@ async def handle_message(adapter: TelegramAdapter, msg: Any) -> None:
     # In private chats, always respond.
     if hub_msg.platform_meta.get("is_group") and not hub_msg.is_mention:
         return
+
+    # Session wiring: inject prior session_id + persist callback.
+    _meta_updates: dict[str, Any] = {}
+    if adapter._turn_store is not None:
+        from lyra.core.hub.hub_protocol import RoutingKey
+        from lyra.core.message import Platform
+        _pool_id = RoutingKey(
+            Platform.TELEGRAM, adapter._bot_id, hub_msg.scope_id
+        ).to_pool_id()
+        try:
+            _last_sid = await adapter._turn_store.get_last_session(_pool_id)
+        except Exception:
+            log.exception("TurnStore.get_last_session failed for pool_id=%s", _pool_id)
+            _last_sid = None
+        if _last_sid is not None:
+            _meta_updates["thread_session_id"] = _last_sid
+        _ts = adapter._turn_store
+
+        async def _tg_session_update_fn(
+            msg: InboundMessage, session_id: str, pool_id: str
+        ) -> None:
+            await _ts.start_session(session_id, pool_id)
+
+        _meta_updates["_session_update_fn"] = _tg_session_update_fn
+    if _meta_updates:
+        hub_msg = dataclasses.replace(
+            hub_msg,
+            platform_meta={**hub_msg.platform_meta, **_meta_updates},
+        )
 
     log.info(
         "message_received",
--- a/src/lyra/adapters/telegram_normalize.py
+++ b/src/lyra/adapters/telegram_normalize.py
@@ -13,7 +13,6 @@ from lyra.core.message import (
     Platform,
     RoutingContext,
 )
-from lyra.core.scope import user_scoped
 from lyra.core.trust import TrustLevel
 
 if TYPE_CHECKING:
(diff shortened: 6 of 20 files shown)


MickaelV0:
## Code Review

**Verdict: Request Changes** — 6 blocking findings (4 bugs, 2 test bugs), 5 non-blocking suggestions.

---

### Blockers

---

```
issue: start_session() exception unhandled in reset_session() — pool and store go out of sync
  src/lyra/core/pool/pool.py:255-258
  -- backend-dev
  Root cause: end_session_async() wraps its TurnStore call via PoolObserver's try/except, but the
  start_session() call in reset_session() is bare. If TurnStore raises (DB locked, schema error),
  the UUID has already rotated but no session row exists — the new session will never be resumed
  because it was never registered. Pool and store are now permanently out of sync.
  Solutions:
    1. Wrap `await self._observer._turn_store.start_session(...)` in try/except with error log,
       mirroring PoolObserver's end_session_async pattern. (recommended)
    2. Move start_session into a PoolObserver.start_session_async() helper that includes the guard.
  Confidence: 91%
```

---

```
issue: _pending_session_id overwrites silently — rapid reply-to bursts drop earlier session resumes
  src/lyra/core/hub/middleware_submit.py:147
  src/lyra/core/pool/pool_processor.py:50-53
  -- backend-dev
  Root cause: Two reply-to messages arriving while the pool is busy will both set
  _pending_session_id. The second assignment unconditionally overwrites the first before the
  processor drains — the first queued resume is silently lost with no log or indication.
  Solutions:
    1. Replace scalar `_pending_session_id: str | None` with a `deque[str]` and append/popleft.
       The processor drains all pending resumes before dispatch. (recommended)
    2. Guard the assignment: only set if None, else log a warning "dropping earlier queued resume".
  Confidence: 85%
```

---

```
issue: process_loop fires resume_session() before agent lookup — exception drops the buffered message
  src/lyra/core/pool/pool_processor.py:47-55
  -- backend-dev
  Root cause: _pending_session_id is fired before get_agent() at line 55. If resume_session()
  raises (not currently guarded), the buffered message collected after collect() is silently
  dropped — no agent dispatch, no user-visible response. The outer `except asyncio.CancelledError`
  does not protect this path.
  Solutions:
    1. Wrap the resume_session() call in try/except so failures log and fall through to agent
       dispatch rather than bubbling up. (recommended)
    2. Restructure: resolve agent first, then resume — they are semantically independent.
  Confidence: 82%
```

---

```
suggestion(blocking): _session_update_fn dual-assignment in discord_inbound — thread callback silently shadows DM callback
  src/lyra/adapters/discord_inbound.py:218-235
  -- backend-dev
  Root cause: Both DM and thread-session paths append `_session_update_fn` to `_meta_updates`.
  For a DM message that also has a thread_id (edge case), the thread callback at line 231-234 always
  overwrites the DM callback in the dict. Separately, in middleware_submit.py:83-85, only the first
  message registers the callback — subsequent messages with a different callback are ignored.
  Combined: session persistence for one of the two paths may be silently dropped.
  Solutions:
    1. Make the two callbacks mutually exclusive with an explicit `elif` in discord_inbound.py,
       and document the priority. (recommended)
    2. Use distinct keys (_session_update_fn_dm, _session_update_fn_thread) and let middleware
       prioritize.
  Confidence: 80%
```

---

```
issue: test negative-path extracts wrong argument — assert is vacuous (Discord DM + Telegram)
  tests/integration/test_session_dm_discord.py:147
  tests/integration/test_session_telegram.py:181
  -- backend-dev
  Root cause: Both negative-path tests read `mock_bus.put_nowait.call_args[0][0]`, but
  push_to_hub_guarded calls `bus.put(platform, msg)` making msg at index [0][1]. Index [0][0] is
  the Platform enum — `"thread_session_id" not in Platform.DISCORD.platform_meta` would raise
  AttributeError rather than pass, or the conditional guard swallows the check entirely when
  put_nowait was never called.
  Solutions:
    1. Change `[0][0]` to `[0][1]` in both files, and add `assert mock_bus.put_nowait.called`
       before extraction to make the assertion non-conditional. (recommended)
  Confidence: 92%
```

---

```
issue: asyncio.get_event_loop().create_task() deprecated — may attach to wrong loop under pytest-asyncio
  tests/integration/test_session_reply_to.py:111
  -- backend-dev
  Root cause: asyncio.get_event_loop() is deprecated in Python 3.10+ when called outside a running
  event loop context. Inside an async test function, asyncio.create_task() should be used — it
  automatically uses the currently running loop managed by pytest-asyncio.
  Solutions:
    1. Replace `asyncio.get_event_loop().create_task(asyncio.sleep(100))` with
       `asyncio.create_task(asyncio.sleep(100))`. (recommended)
  Confidence: 89%
```

---

### Non-Blocking Suggestions

---

```
suggestion(non-blocking): No test exercises process_loop actually consuming _pending_session_id
  (test suite gap)
  -- backend-dev
  Root cause: test_session_reply_to.py only verifies that middleware SETS _pending_session_id when
  the pool is busy. No test verifies that process_loop CONSUMES it and calls resume_session() with
  the correct session_id. The most critical integration path is untested.
  Solutions:
    1. Add a test: pool with mocked _session_resume_fn, make it busy, set _pending_session_id,
       submit a message, run the event loop, assert _session_resume_fn was called. (recommended)
  Confidence: 95%
```

---

```
suggestion(non-blocking): increment_resume_count() not called for pending-path resumes (busy-pool reply-to)
  src/lyra/core/pool/pool_processor.py:50-53
  -- backend-dev
  Root cause: middleware_submit.py:161 calls increment_resume_count() on the idle-path resume, but
  pool_processor.py does not after firing _pending_session_id. TurnStore resume counters will be
  under-counted for all busy-pool reply-to resumes.
  Solutions:
    1. Add a _on_resume_fn callback to Pool (mirroring _session_reset_fn pattern) that pool_processor
       calls after a successful resume, letting hub wire in increment_resume_count. (recommended)
    2. Accept as a known limitation and document it.
  Confidence: 87%
```

---

```
suggestion(non-blocking): Stale "RED phase" docstrings in test_session_clear.py mislead readers
  tests/integration/test_session_clear.py:1-9
  -- backend-dev
  Root cause: Module and test docstrings say "RED phase" and "will fail until T2 is implemented."
  T2 (Pool.reset_session() rotation) is fully implemented. These comments are TDD artifacts.
  Solutions:
    1. Remove RED-phase comments, update docstrings to describe what each test validates. (recommended)
  Confidence: 97%
```

---

```
thought: middleware_submit path 1 (busy pool) should return early after setting _pending_session_id
  src/lyra/core/hub/middleware_submit.py:147-162
  -- backend-dev
  Root cause: After setting _pending_session_id and logging, execution falls through to paths 2 and
  3. Path 3 is safely skipped (pool.is_idle == False), but path 2 (thread_session_id) could attempt
  an additional resume for the same message if both reply-to and thread_session_id are present.
  An explicit return ResumeStatus.SKIPPED after the log at line 152 would clarify intent and prevent
  path 2 from running.
  Confidence: 83%
```

---

```
thought: thread_session_id survives the NATS boundary — scope-validation at middleware_submit.py:168 is the guard
  src/lyra/nats/_sanitize.py:22 (pre-existing, not introduced by this PR)
  -- security-auditor
  Root cause: thread_session_id is in PLATFORM_META_ALLOWLIST since PR #535, so it crosses the
  NATS wire in the three-process topology. The defense against injection is middleware_submit.py:168
  checking session_pool == pool_id (pool_id derived from authenticated scope). This guard is sound
  given Discord's gateway auth and Telegram's HMAC webhook verification. Worth documenting explicitly.
  Confidence: 78%
```

---

### Praise

---

```
praise: reset_session() ordering is correct — end_session → rotate → start_session
  src/lyra/core/pool/pool.py:251-261
  -- backend-dev
  This is the only safe ordering. Rotating before end would orphan the old session; starting before
  end would leave a window where neither is active.
  Confidence: 97%
```

---

```
praise: _dm_ts / _dm_session_update_fn variable naming caught and fixed the closure capture bug
  src/lyra/adapters/discord_inbound.py:218-222
  -- backend-dev
  Renaming to _dm_ts avoids the bug where the closure would have captured the later _ts reassignment
  (thread store). This kind of closure-capture fix is subtle and was handled correctly.
  Confidence: 95%
```

---

### Summary

| Severity | Count |
|----------|-------|
| Blocking bugs | 4 |
| Blocking suggestions | 1 |
| Test bugs | 1 (covers 2 files) |
| Non-blocking | 5 |
| Praise | 2 |

**Most critical fixes before merge:**
1. `reset_session()` — wrap `start_session()` in try/except
2. `_pending_session_id` — guard against silent overwrite (log at minimum, deque ideally)
3. `process_loop` — guard `resume_session()` in try/except so message isn't dropped
4. Two test files — fix `[0][0]` → `[0][1]` and `asyncio.get_event_loop().create_task()`
5. `discord_inbound` — make DM/thread callback mutually exclusive

---

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, verified):** 7 findings
**Deferred (issue created):** 1 finding → #597
**Skipped:** 0
**Failed:** 0

### Auto-Applied
- [applied] `issue`: `start_session()` bare in `reset_session()` — wrapped in try/except (91%)
- [applied] `issue`: `_pending_session_id` overwrite silently — log guard in processor (85%)
- [applied] `issue`: `process_loop` `resume_session()` unguarded — wrapped in try/except (82%)
- [applied] `suggestion(blocking)`: discord_inbound dual `_session_update_fn` — made mutually exclusive with elif (80%)
- [applied] `issue`: test arg index `[0][0]` → `[0][1]` in discord + telegram negative tests (92%)
- [applied] `issue`: `asyncio.get_event_loop().create_task()` deprecated → `asyncio.create_task()` (89%)
- [applied] `suggestion(non-blocking)`: stale RED-phase docstrings removed (97%)

### Applied via 1b1
- [applied] `suggestion(non-blocking)`: `middleware_submit` path 1 now returns `ResumeStatus.SKIPPED` after queuing (83%)
- [applied] `suggestion(non-blocking)`: added `test_process_loop_fires_pending_session_id` — verifies process_loop consumes `_pending_session_id` (95%)

### Deferred
- `suggestion(non-blocking)`: `increment_resume_count` not called for pending-path resumes → [#597](https://github.com/Roxabi/lyra/issues/597) (87%)

---

PR #711, opened 2026-04-15
feat(nats): validate contract_version on inbound envelopes

MickaelV0:
## Summary
- Add `check_contract_version()` in `src/lyra/nats/_version_check.py`; wire it into `NatsAdapterBase._validate_envelope` alongside the existing `schema_version` check.
- Prevents a stale hub from silently processing envelopes whose `contract_version > hub's CONTRACT_VERSION` after an ADR-044 bump — drops + counts + rate-limited ERROR log, mirroring schema-mismatch semantics.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #707: feat(nats): validate contract_version on inbound envelopes | Open |
| Implementation | 1 commit on `worktree-707-validate-contract-version` | Complete |
| Verification | Lint ✅ · Typecheck ✅ · Tests ✅ (13 new, 237 passed) | Passed |

## Test Plan
- [ ] `uv run pytest tests/nats/` — all pass (237/237).
- [ ] Payload with `contract_version: "2"` and `schema_version: 1` hits `_validate_envelope` → dropped, `_drop_count["InboundMessage"] == 1`, ERROR log emitted.
- [ ] Payload without `contract_version` field still accepted (legacy compat).
- [ ] Malformed values (`None`, `"v1"`, `1.5`, `True`, `"0"`, `"-1"`) are dropped.

Closes #707

---
Generated with [Claude Code](https://claude.com/claude-code) via `/pr`

Issue #523, opened 2026-04-04
feat(nats): wire nkey authentication in Python clients


MickaelV0:
Both `hub_standalone.py` and `adapter_standalone.py` call `nats.connect(url)` with no `nkeys_seed_file`, `tls`, or `tls_ca_file`. The server config (`nats.conf`) requires nkey auth but Python clients never provide credentials. Works only because `nats-local.conf` (no auth) is deployed.

**Touches:** bootstrap files, run scripts, `.env`, deploy docs.
**Needs:** real NATS server testing with auth enabled.
**Related:** #461 (JWT/account auth) is the broader auth story; this issue wires the existing nkey infra.

PR #714, opened 2026-04-15
feat(nats): per-role nkeys + subject ACLs in auth.conf

### LINKED ISSUE #706: feat(nats): per-role nkeys + subject ACLs in nats-server config
Define per-role nkeys in `nats-server` configuration with subject-prefix allowlists.

## Roles

| Role | Allowed subjects |
|---|---|
| hub | `lyra.inbound.>`, `lyra.outbound.>`, `lyra.voice.*.request` |
| tts-worker | `lyra.voice.tts.>` |
| stt-worker | `lyra.voice.stt.>` |
| plugin | project-scoped namespace only (TBD) |

## Why

Without per-role ACLs, any plugin that installs `roxabi-nats` and calls `nats_connect()` inherits the process nkey. A buggy or malicious plugin could publish to `lyra.inbound.*` and inject messages into the hub as if from a legitimate adapter.

## References

- `src/lyra/nats/connect.py:111` — `nats_connect()` public entry point
- `src/lyra/nats/__init__.py:28` — exported surface
- Security review: Broken Access Control (OWASP 5), Medium severity, confidence 75%

Blocks roxabi-nats SDK extraction.

MickaelV0:
## Summary
- Add per-identity publish/subscribe allow-lists + `allow_responses: true` to nats-server `auth.conf` for 7 identities (hub, telegram-adapter, discord-adapter, tts-adapter, stt-adapter, llm-worker, monitor). Before: every authenticated nkey had full bus access — any process (including future plugins via ADR-045) could publish to `lyra.inbound.*` and forge adapter traffic. After: each identity is scoped to exactly the subjects its role needs.
- Extend `deploy/nats/gen-nkeys.sh` with a matrix-driven `auth.conf` generator, `--template-only` mode for CI (sudo-less), and `--regenerate --yes` with atomic backup of both `auth.conf` and `~/.lyra/nkeys/` for safe rollback. Wire `NATS_NKEY_SEED_PATH` into `lyra_hub.conf` / `lyra_telegram.conf` / `lyra_discord.conf` supervisor configs.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | [#706](../../issues/706): feat(nats): per-role nkeys + subject ACLs in nats-server config | OPEN |
| Frame | [artifacts/frames/706-per-role-nkeys-acls-frame.mdx](artifacts/frames/706-per-role-nkeys-acls-frame.mdx) | Approved |
| Spec | [artifacts/specs/706-per-role-nkeys-acls-spec.mdx](artifacts/specs/706-per-role-nkeys-acls-spec.mdx) | Approved |
| Plan | [artifacts/plans/706-per-role-nkeys-acls-plan.mdx](artifacts/plans/706-per-role-nkeys-acls-plan.mdx) | Approved |
| Implementation | 1 commit on `feat/706-per-role-nkeys-acls`, 19/19 micro-tasks complete | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (2748 passed, 1 new shell test) | Passed |

## Notable design choices

- **`_INBOX.>` in hub subscribe allow-list.** `NatsLlmDriver.stream()` uses a manual inbox subscription (`new_inbox()` + `subscribe(inbox)`) rather than `nc.request()`, so `allow_responses: true` alone does NOT cover it. Hub needs `_INBOX.>` explicitly in its subscribe list. Other identities rely on `allow_responses: true` (responder pattern via `msg.respond()`).
- **`lyra.llm.health.*` single-token wildcard** (not `>`) matches the current `NatsLlmDriver` subscribe pattern in `src/lyra/llm/drivers/nats_driver.py:26-34`. Tightens future drift.
- **Plugin ACL intentionally deferred** until ADR-045 (roxabi-nats SDK extraction) lands. The `auth.conf` header comment references ADR-045; no plugin identity is generated.
- **Rollback atomicity.** `--regenerate` rotates ALL seeds, so restoring only `auth.conf.bak` leaves old pubkeys with no matching seeds. The runbook restores both backup pairs atomically.

## Spec deviation (documented)

The spec's AC said the auth.conf header should contain the literal string "plugin ACL deferred", but the V1 test assertion (e) forbids the word `plugin` anywhere in the generated conf. The generator uses `ACL for roxabi-nats SDK identities intentionally omitted — deferred until ADR-045 lands` in the header — same intent, satisfies assertion (e). The `# plugin` identity block is absent as specified.

## T3.2 documented skip

The V3 positive-case test (simulate a `Permissions Violation` log entry and assert `check-nats-acls.sh` exits 1) was skipped: `journalctl -u <unit>` filters on `SYSTEMD_UNIT`, but `systemd-cat --identifier=X` only sets `SYSLOG_IDENTIFIER`. Reliable injection requires root-owned transient units (`systemd-run`), out of scope for this PR. The spec explicitly accepts a documented skip when `-u` filtering is ambiguous. The negative case (no violations → exit 0) is verified by the T3.3 gate.

## Test Plan
- [ ] Machine 1: `sudo ./deploy/nats/gen-nkeys.sh --regenerate --yes` — verify 7 seeds in `~/.lyra/nkeys/`, `/etc/nats/nkeys/auth.conf` with 7 `users[]` blocks, timestamped backups at `auth.conf.bak.{epoch}` and `~/.lyra/nkeys.bak.{epoch}/`.
- [ ] Machine 1: `sudo systemctl reload nats.service` — `systemctl status nats.service` shows active, no `Permissions Violation` in `journalctl -u nats.service --since "5 minutes ago"`.
- [ ] Machine 1: reconnect in order `make telegram reload && make discord reload && make tts reload && make stt reload && make lyra reload` — `make ps` shows all RUNNING, `make lyra errors` / `make tts errors` / `make stt errors` have no permission violations.
- [ ] Machine 1: `scripts/check-nats-acls.sh --since "$(date -Iseconds)" --window 90 | tee rollout-evidence.txt` exits 0. Attach `rollout-evidence.txt` to this PR before merge.
- [ ] Rollback drill (non-destructive): restore `auth.conf.bak` + seeds backup on a test host, reload, verify processes reconnect with old credentials.

Closes #706

---


MickaelV0:
## Code Review — PR #711

**Verdict: Approve with comments.** No blockers. Two high-confidence issues around shared rate-limit / counter state between the schema and contract drop paths are worth a follow-up fix before an ADR-044 contract bump actually lands. Other findings are non-blocking polish.

### Warnings (non-blocking issues)

**issue: Shared rate-limit key silences schema logs when contract drops flood, and vice versa**
- `src/lyra/nats/_version_check.py:182` — security-auditor · tester (95%)
- Root cause: `_last_log_ts` is keyed on `envelope_name` only, so any kind's drop within the 60 s window suppresses all kinds' ERROR logs for that envelope. An attacker (or misbehaving plugin) producing a steady contract-mismatch stream would blind operators to concurrent schema failures. No test exercises cross-kind silencing.
- Solutions:
  1. Key `_last_log_ts` on `f"{envelope_name}:{kind}"` so the two rate-limit windows are independent. (recommended)
  2. Separate module-level dicts: `_last_schema_log_ts` / `_last_contract_log_ts`.

**issue: `_drop_count` telemetry cannot distinguish schema drops from contract drops**
- `src/lyra/nats/_version_check.py:174` — security-auditor · backend-dev (90%)
- Root cause: Same `envelope_name` key is used for both kinds, so `_drop_count["InboundMessage"]` aggregates across failure modes. Alert thresholds calibrated against one kind will fire for the other.
- Solutions:
  1. Key the counter as `f"{envelope_name}:{kind}"` (e.g. `"InboundMessage:contract"`). Update assertions in `test_adapter_base.py` / `test_version_check.py`. (recommended)
  2. Separate per-kind counter dicts on `NatsAdapterBase`.

**issue: Integration test suite never exercises the schema-fails-first short-circuit**
- `tests/nats/test_adapter_base.py:161` — tester (90%)
- Root cause: Both new integration tests use `schema_version=1` (valid). A regression removing the `if not check_schema_version` guard would not be caught.
- Solutions:
  1. Add `test_invalid_schema_version_short_circuits`: payload with `schema_version=2, contract_version="1"` → `_drop_count == {"InboundMessage": 1}` (exactly one increment, from schema). (recommended)

**issue: Malformed hub constant is caught per-message instead of at startup**
- `src/lyra/nats/_version_check.py:146` — backend-dev (88%)
- Root cause: `int(expected)` runs on every received envelope. A typo in `CONTRACT_VERSION` produces silent per-message drops (rate-limited after 60 s), not a crash. This is hub config, not wire data.
- Solutions:
  1. Add `assert CONTRACT_VERSION.isdigit() and int(CONTRACT_VERSION) > 0` at module scope in `adapter_base.py`. Remove the per-message `try/except`. (recommended)
  2. Log at CRITICAL and bypass rate-limit for the hub-config branch.

**issue: `check_contract_version` accepts bare `int` payloads — asymmetric with schema check**
- `src/lyra/nats/_version_check.py:134` — backend-dev (85%)
- Root cause: No actual producer stamps `contract_version` as int; all sites use the string `CONTRACT_VERSION`. The int-acceptance branch is dead-tolerant lenience and creates asymmetry with `check_schema_version` (which rejects non-int).
- Solutions:
  1. Remove the `isinstance(raw, int)` branch; keep str-only. Flip `test_int_payload_accepts` to assert False. (recommended)
  2. Document the intentional asymmetry in the docstring if kept.

**issue: `test_malformed_hub_expected_drops` asserts drop but not log — format-string regressions invisible**
- `tests/nats/test_version_check.py:435` — tester (85%)
- Root cause: Sister schema tests all assert `caplog` content; this one relies on counter only. A future typo that drops the `kind` arg would pass silently.
- Solutions:
  1. Add `caplog` fixture and assert `"NATS contract version mismatch"` in message. (recommended)

**issue: `TestLogRateLimit` only exercises schema path; contract rate-limit untested**
- `tests/nats/test_version_check.py` (TestLogRateLimit) — tester (80%)
- Root cause: If `kind="contract"` were dropped in a future refactor, rate-limit tests would still pass while the contract log path silently broke.
- Solutions:
  1. Add `test_contract_drops_are_rate_limited` mirroring `test_subsequent_drops_within_interval_silent`. (recommended)

**issue: `int()` accepts `"1_000"`, `"+1"`, and unicode digits — out of ADR-044 wire spec**
- `src/lyra/nats/_version_check.py:138` — security-auditor (78%)
- Root cause: Python `int()` is lenient; the wire contract is plain decimal. `"+1"` silently equals `"1"` today.
- Solutions:
  1. Pre-validate with `re.fullmatch(r'[1-9][0-9]*', raw)` before `int()`. (recommended)
  2. Round-trip check `str(int(raw)) == raw` (weaker — still accepts `+1` via reconstruction).

### Suggestions / nits

**suggestion: Add startup assert on `CONTRACT_VERSION`**
- `src/lyra/nats/adapter_base.py:32` — backend-dev (80%)
- `_EXPECTED_CONTRACT_VERSION = int(CONTRACT_VERSION)` at module scope — converts config errors to import-time crashes.

**suggestion: `TestCheckContractVersion` acceptance tests omit AAA comments used throughout `TestCheckSchemaVersion`**
- `tests/nats/test_version_check.py:290` — tester (75%)

**nitpick: `test_out_of_range_drops` parametrize cases don't assert log emission**
- `tests/nats/test_version_check.py:392` — tester (70%)

**nitpick: `# noqa: PLR0913` on `_drop` is a growing-arity smell**
- `src/lyra/nats/_version_check.py:164` — backend-dev (78%)
- Accept for now; introduce `_DropContext` dataclass if a third check type arrives.

**nitpick: `_validate_envelope` emits one log line per failed check — becomes noisy at ≥3 checks**
- `src/lyra/nats/adapter_base.py:97` — backend-dev (72%)

**question: Heartbeat payloads carry `contract_version` but no consumer validates them in this PR**
- `src/lyra/nats/adapter_base.py:118` — backend-dev (75%)
- Explicitly document heartbeat validation as out of scope; open follow-up issue.

### Praise
- Rate-limited ERROR log pattern reused cleanly via the `kind`-parameterized `_drop`.
- Forward-compat semantics (accept equal/lesser, reject strictly greater) match the existing `schema_version` rule exactly.
- 100% coverage on `_version_check.py` after the change.

---
Generated by `/code-review` (dev-core) — security-auditor · tester · backend-dev (3 fresh agents)

MickaelV0:
## Review Fixes Applied

Architect second-pass review surfaced 3 findings the first-pass missed (**C1** adapter `system.ready` audit — no forge surface; **C2** `default_permissions` deny-all; **C3** missing subscribe coverage). Applied the long-term structural fixes where they belonged in this PR; deferred design changes that would push out of F-lite into F-full.

**Auto-applied (commit 538076d):** 3 findings
**Applied via architect-validated 1b1 (commit 942aafa):** 9 findings (including architect's 3 additions)
**Deferred to follow-up issues:** 4 findings
**Skipped (architect rejected first-pass finding):** 1 finding
**Deferred to rollout:** 1 finding (B6 needs Machine 1 access)

---

### Auto-Applied (C≥80, commit 538076d)
- [applied] **B1** issue(blocking): `NATS_UNIT` default `nats-server.service` → `nats.service` (C=95) — the first finding would have silently defeated the rollout gate.
- [applied] **NB1** suggestion: `--regenerate` non-TTY bypass → now requires explicit `--yes` (C=72).
- [applied] **NB2** nitpick: `nk` binary SHA-256 check failure → `error` instead of `warn` (C=78).

### Applied via architect-validated 1b1 (commit 942aafa)
- [applied] **B2 + C3** issue(blocking) [C=95]: test harness rewritten for full matrix coverage (7 identities × {pub, sub}) with set equality (catches missing AND over-privilege), brace-depth anchored block extraction (no awk range leaks), new assertion (g) for `default_permissions` fallback.
- [applied] **B4** issue(blocking) [C=80]: atomic backup via `ERR/INT/TERM` trap — if generation fails after the `rm` but before the new auth.conf is written, backups auto-restore and the script aborts with an error. Trap cleared on successful write.
- [applied] **C2** (architect): `default_permissions { publish: {deny: [">"]}, subscribe: {deny: [">"]} }` inside `authorization{}` — any future user added without an explicit `permissions{}` block defaults to deny-all instead of NATS's implicit allow-all.
- [applied] **NB3** [C=65]: explicit `chmod 0700` on backup dir + `0600` on backed-up seeds after `cp -a`.
- [applied] **NB4** [C=62]: `check-nats-acls.sh` validates `--since` is a parseable past timestamp; future or invalid values fail with exit 2.
- [applied] **NB7** [C=75]: extracted `render_auth_conf` shared function — `--template-only` and `--regenerate` paths now call the same renderer instead of duplicating the `authorization{}` scaffold.
- [applied] **NB12** [C=90]: idempotency warn now parses the existing auth.conf's identity count and flags when < 7 (surfaces the silent upgrade path for operators).
- [applied] **NB14** [C=70]: `docs/DEPLOYMENT.md` runbook fixed — lyra Makefile targets are `lyra-tts` / `lyra-stt`, not `tts` / `stt`.
- [applied] **NB16** [C=70]: test CWD guard — assert `gen-nkeys.sh` is executable after `cd`, fail loudly otherwise.

### Deferred to follow-up issues
- #715 — **B3** issue(blocking): migrate `NatsLlmDriver.stream()` from manual inbox to `nc.request()` so `_INBOX.>` drops from the hub ACL (security hardening, architect classified as F-full scope). Inline code comment at `SUB_ALLOW[hub]` references #715.
- #716 — **B5 + NB15** issue(blocking) + suggestion: CI integration test for T3.2 using a real `nats-server` binary, replacing the documented skip. `# SKIP T3.2: deferred to #716` comment added to the test file.
- #717 — **NB5 + NB6**: `deploy/nats/acl-matrix.json` as single source of truth consumed by `gen-nkeys.sh`, the test harness, and the spec-matrix renderer — eliminates the 3-way drift (spec / bash arrays / supervisor confs).
- #718 — **NB11**: event-driven `check-nats-acls.sh` using `journalctl -f` + `timeout --preserve-status`.

### Skipped (architect rejected first-pass finding)
- **NB8** suggestion: remove `allow_responses: true` from subscriber-only identities. Architect pushback was correct — `msg.respond()` in tts-adapter / stt-adapter / llm-worker (responders) requires that permission to publish on the ephemeral reply subject. Without it, request/reply breaks. Only `monitor` is truly subscriber-only and its marginal risk (1 reply per request, 2m TTL) is not worth a special-case branch in the generator. Left as-is.

### Deferred to rollout (needs Machine 1)
- **B6** issue(blocking): `rollout-evidence.txt` can only be produced on Machine 1. Operator decision pending: run rollout against this branch pre-merge, or merge first and capture evidence within 24h as an AC amendment.

### TODO(ADR-045) markers
Added in `deploy/nats/gen-nkeys.sh` at the matrix section and in the `render_auth_conf` header so the plugin ACL reminder surfaces when ADR-045 lands.

### Diff since initial review
```
 deploy/nats/gen-nkeys.sh               | ~200 lines restructured + hardened
 docs/DEPLOYMENT.md                     |   1 line (runbook Makefile target fix)
 scripts/check-nats-acls.sh             |  ~20 lines (NATS_UNIT default + --since guard)
 tests/nats/test_gen_nkeys_acls.sh      | ~110 lines (full matrix, set equality, block scoping)
```

CI reruns automatically on push. Ready for re-review.

MickaelV0:
## Code Review (iteration 2)

**Verdict: APPROVE with follow-ups** — 3 new blockers found and fixed in this iteration (commit 4f81df2). 7 non-blockers deferred or already tracked.

Reviewers: security-auditor, architect, devops, tester (all fresh, no impl context).

---

### 🛑 New blockers (found in iter 2, all fixed in 4f81df2)

**B7** issue(blocking) · C=95 · security-auditor + architect
  `deploy/nats/gen-nkeys.sh` — `restore_backups` trap handler called `error()` which invokes `exit 1` under `set -e`, re-triggering the ERR trap recursively → infinite loop on any real regen failure. Fix: clear the trap on entry (`trap - ERR INT TERM`), then `echo` + direct `exit 1`.

**B9** issue(blocking) · C=92 · security-auditor + tester (mutation-tested)
  `tests/nats/test_gen_nkeys_acls.sh` — `extract_block` awk exited on the first `}` at `depth == 1`, which is the file's final `}`. The range leaked from the identity comment to end of file (measured: **63 lines instead of 8**). Tests passed by accident because the caller's `head -1` took the first matching `publish:`/`subscribe:` line (hub's own). Fix: record `entry_depth` when the identity anchor is seen; exit on the `}` that returns depth to `entry_depth`. Verified: all 7 identities now extract to exactly 7 lines each.

**NB17** nitpick · C=90 · devops
  `docs/DEPLOYMENT.md:305` — the original NB14 fix (`make tts reload` → `make lyra-tts reload`) didn't persist to the initial fix commit. Re-applied in 4f81df2.

### ⚠ Contested — deferred to rollout validation

**B8** issue · C=85 · security-auditor
  `deploy/nats/gen-nkeys.sh` — security-auditor claimed `default_permissions: { ... }` colon syntax is invalid for NATS. Contested because `permissions: { ... }` uses the same colon pattern throughout the existing users[] block (established since #523). NATS config is HOCON-ish and accepts both `:` and `=` as assignments before brace blocks. Machine 1 rollout (B6 gate) will empirically validate — if NATS rejects the config on reload, `rollout-evidence.txt` surfaces it within the 90s window.

### ⚠ Non-blockers deferred

- **NB18** (architect): `restore_backups` "state restored from ${epoch}" message is misleading when the machine had no pre-existing state to restore. Cosmetic, low impact.
- **NB19** (architect): NB12 grep pattern `^[[:space:]]*# [a-z][a-z0-9-]*$` could false-positive on any future lowercase `# word` comment — defer to #717 (typed identity labels eliminate the risk).
- **NB20** (tester): assertion (a) counts identity comment labels, not `nkey:` blocks — a regression that emits `# name` but drops `nkey:` would pass (a). Low probability; defer to #716 (real nats-server integration test fails to parse).
- **NB21** (tester): no supervisor-conf coverage for `llm-worker`/`monitor` — these are quadlet-only or reserved. Defer to #717 (`seed_consumer` field + `--validate-supervisor` subcommand).
- **NB22** (tester): `extract_block` inline comment previously claimed "brace-depth anchored" which was inaccurate. Fixed in B9 with an accurate `entry_depth` description.
- **NB23** (architect): TODO(ADR-045) marker absent from spec §Out-of-scope. The spec already documents the deferral; a marker is redundant. Skip.

### Deferred since iteration 1
- #715 — B3 `_INBOX.>` eavesdropping → migrate `NatsLlmDriver.stream()` to `nc.request()`
- #716 — T3.2 + real nats-server CI integration test
- #717 — acl-matrix.json as single source of truth
- #718 — event-driven check-nats-acls.sh

### ✅ Praise

- **Security-auditor**: NB1 non-TTY guard — no env-var bypass exists; `AUTO_YES` only set via `--yes` flag.
- **Architect**: `render_auth_conf` extraction via bash nameref is clean; both call sites confirmed; header comments rendered once.
- **Devops**: NB4 `--since` guard accepts both ISO-8601 and `YYYY-MM-DD HH:MM:SS` formats; silent 0-match scenario blocked.
- **Tester**: mutation testing confirmed set-equality — swapping inbound/outbound on discord, injecting `lyra.admin.>` on tts-adapter, removing `_INBOX.>` from hub all correctly fail.

---

### Summary

| Category | Count |
|---|---|
| Iter-2 blockers (fixed 4f81df2) | 3 |
| Contested, deferred to rollout | 1 |
| Non-blockers deferred to follow-ups | 6 |
| Praise | 4 |

**Verdict:** **APPROVE with follow-ups.** Iteration 2 caught 3 real blockers the first review missed (trap recursion, awk depth leak, docs regression). All fixed. Remaining items are non-blocking and either tracked in follow-up issues or deferred to the Machine 1 rollout gate (B6).

Re-review loop cap reached (iteration 2 = max). Next step is merge with B6 evidence.

PR #724, opened 2026-04-15
feat(nats): extract roxabi-nats SDK as uv workspace subpackage (#721)

MickaelV0:
## Summary

Extract 9 transport-primitive files from `src/lyra/nats/` into `packages/roxabi-nats/` as a uv workspace member. External plugins (voiceCLI, roxabi-vault, imageCLI) can now consume it via `{ git=..., subdirectory="packages/roxabi-nats", tag=... }`.

**Cohort A (moved to package):**
- adapter_base.py, connect.py, circuit_breaker.py, readiness.py
- _validate.py, _version_check.py, _sanitize.py, _serialize.py, _tts_constants.py

**Cohort B (stays in lyra, imports from roxabi_nats):**
- nats_bus.py, queue_groups.py, render_event_codec.py
- nats_channel_proxy.py, nats_tts_client.py, nats_stt_client.py

## Test plan

- [x] `uv run pytest tests/` — 2673 passed, 60 skipped, 83% coverage
- [x] `uv run pytest packages/roxabi-nats/tests/` — 135 passed, 7 skipped
- [x] `uv run pyright` — clean (voicecli optional dep ignored)
- [x] `uv run ruff check .` — clean
- [x] `uv run lyra --help` — boots without import errors
- [x] No `from lyra.nats.{adapter_base,connect,...}` imports remaining in lyra src/tests
- [ ] Tag `roxabi-nats/v0.1.0` after merge

## References

- ADR: docs/architecture/adr/045-roxabi-nats-sdk-uv-workspace-extraction.mdx
- Issue: #721

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #774, opened 2026-04-17
refactor(roxabi-nats): per-adapter type_registry, delete global registry

### LINKED ISSUE #729: refactor(roxabi-nats): replace global type-checking registry with explicit per-adapter init param
## Context

`roxabi_nats._serialize.register_type_checking_import(module_path, type_name)` maintains `_TYPE_CHECKING_IMPORTS`, a process-global mutable registry. At deserialization time, `_get_hints()` calls `importlib.import_module()` on each registered module path.

Surfaced during review of PR #724 (roxabi-nats SDK extraction). Both the architect and security-auditor agents independently flagged the global registry as an architectural smell worth refactoring before external plugins (voiceCLI, roxabi-vault, imageCLI) start consuming the SDK.

Short-term mitigation landing in PR #724: demote the helper to `_register_type_checking_import` (remove from `__all__`, remove from `__init__.py` re-export) to make the hub-internal nature explicit.

## Problem

- **Implicit global state**: any consumer importing `roxabi_nats` can mutate the registry. Order-dependent behaviour; hard to reason about across adapter instances.
- **Parallel test isolation**: tests that rely on different type-hint sets interfere through the shared registry.
- **Security surface** (as SDK scope grows): `importlib.import_module` on a caller-controlled string is a code-execution sink. Internally-trusted today, but the blast radius grows with every external consumer.
- **API discoverability**: initialization via side effect — nothing in `NatsAdapterBase.__init__` tells you the adapter needs type hints registered somewhere else.

## Proposal

Replace the global registry with an explicit per-adapter constructor parameter:

```python
# Before
register_type_checking_import("lyra.core.commands.command_parser", "CommandContext")
adapter = NatsAdapterBase(...)

# After
adapter = NatsAdapterBase(
    ...,
    type_registry=[
        ("lyra.core.commands.command_parser", "CommandContext"),
    ],
)
```

The adapter constructs a local `_TypeHintResolver` from that list and passes it into every `deserialize()` call it makes. No module-level mutable state.

## Acceptance criteria

- [ ] `NatsAdapterBase.__init__` accepts `type_registry: Sequence[tuple[str, str]] | None = None`
- [ ] `serialize()`/`deserialize()` accept an optional `resolver` parameter; fall back to a default (empty) resolver if unset — no global lookup
- [ ] All lyra callers migrated to pass `type_registry` at adapter construction time
- [ ] `_register_type_checking_import` marked `# Deprecated — remove in v0.2.0` and emits a `DeprecationWarning` if called
- [ ] `_TYPE_CHECKING_IMPORTS` module-level dict removed
- [ ] Unit tests cover: empty registry, duplicate entries, non-existent module path (raise `ValueError` at adapter init, not at deserialize time — fail fast)
- [ ] ADR-045 Consequences section updated to remove the "known design debt" note

## Target release

`roxabi-nats/v0.2.0` — breaking change in the adapter constructor signature. External consumers (voiceCLI, roxabi-vault, imageCLI) should pin to v0.1.x until they migrate.

## References

- PR #724 (extraction)
- ADR-045 (extraction rationale)
- Code review agent consensus: architect + security-auditor

MickaelV0:
## Summary

- Replace `roxabi_nats._serialize._TYPE_CHECKING_IMPORTS` (process-global mutable registry mutated via `_register_type_checking_import`) with a `_TypeHintResolver` instance passed through `type_registry=` on `NatsAdapterBase.__init__` and `resolver=` on `serialize`/`deserialize`/`deserialize_dict`.
- Clean break in `v0.2.0` — no deprecation shim, no `DeprecationWarning`; the hub-internal helper + the module-level list are deleted outright. External `v0.1.x` consumers that imported the private symbols hit a loud `ImportError` on upgrade (documented trade-off over silent degraded type coercion).
- Lyra-side wiring funnels through a single `TYPE_REGISTRY_RESOLVER` module constant in `src/lyra/nats/type_registry.py`; every NATS consumer (NatsBus, NatsRenderEventCodec, NatsChannelProxy, NatsOutboundListener, InboundCache, nats_envelope_handlers, stt/tts adapters) defaults to it via keyword-only kwarg.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #729: refactor(roxabi-nats): replace global type-checking registry with explicit per-adapter init param | OPEN |
| Frame | [729-replace-global-type-registry-frame.mdx](artifacts/frames/729-replace-global-type-registry-frame.mdx) | Approved |
| Spec | [729-replace-global-type-registry-spec.mdx](artifacts/specs/729-replace-global-type-registry-spec.mdx) | Approved |
| Plan | [729-replace-global-type-registry-plan.mdx](artifacts/plans/729-replace-global-type-registry-plan.mdx) | 22 tasks across V1–V3 |
| Implementation | 3 commits on `feat/729-replace-global-type-registry` (V1 SDK core · V2 adapter base · V3 lyra migration + release) | Complete |
| Verification | Lint ✅ · Typecheck ✅ · Tests ✅ (2972/76 skipped, 84% cov) · ruff-format ✅ · check-file-length ✅ | Passed |

## Test Plan

- [ ] `uv run pytest packages/roxabi-nats/` → 178 pass, 7 skipped (SDK core + adapter base + resolver unit tests)
- [ ] `uv run pytest tests/nats/` → green, includes new `test_nats_bus_defaults_to_type_registry_resolver` and `test_nats_bus_round_trips_inbound_with_non_empty_resolver` (SC-tests lyra-side NatsBus)
- [ ] Boot Lyra via `make lyra` with the unified NATS mode — verify inbound `CommandContext`-bearing payloads deserialize with typed fields (trust_level enum, timestamp datetime, attachments bytes)
- [ ] Grep guards: `grep -rE "_TYPE_CHECKING_IMPORTS|_register_type_checking_import" packages/roxabi-nats/src/ src/lyra/` returns zero production matches (documentation + guard-test references only)
- [ ] `packages/roxabi-nats/pyproject.toml` shows `version = "0.2.0"`; `packages/roxabi-nats/CHANGELOG.md` has a `[0.2.0]` Breaking entry with migration snippet
- [ ] ADR-045 Consequences: "known design debt" entry replaced with "Resolved in #729 (v0.2.0 — clean break, no deprecation shim)"

### Drive-by fixes

- `packages/roxabi-nats/tests/test_readiness.py` — pre-existing `from .conftest import` issue left over from PR #724; inlined the `requires_nats_server` marker so test collection works across rootdirs.
- `src/lyra/core/commands/command_router.py` + `tests/bootstrap/test_hub_builder.py` — ruff-format whitespace/grouping cleanups incidental to the migration.
- Two files trimmed (docstring compression in `nats_bus.py`, module docstring in `command_router.py`) to stay at/under the 300-line limit after the resolver threading.

Closes #729

---

### DIFF
--- a/docs/architecture/adr/045-roxabi-nats-sdk-uv-workspace-extraction.mdx
+++ b/docs/architecture/adr/045-roxabi-nats-sdk-uv-workspace-extraction.mdx
@@ -182,7 +182,7 @@ Both blocking prerequisites are resolved; the `roxabi-nats/v0.1.0` tag is unbloc
 
 ### Known design debt
 
-- **`_register_type_checking_import` global registry** — `_serialize.py` maintains `_TYPE_CHECKING_IMPORTS` as a process-global mutable list populated at lyra startup. This solves the TYPE_CHECKING-hint-resolution problem but imposes implicit global state on the SDK. The helper is kept hub-internal (underscore prefix, not in `__all__`) until a per-adapter explicit init param replaces it. Tracked in issue #729.
+- **Resolved in #729 (v0.2.0 — clean break, no deprecation shim).** The process-global `_TYPE_CHECKING_IMPORTS` list and its `_register_type_checking_import` mutator were deleted outright. Adapters now pass TYPE_CHECKING-only types via an explicit keyword-only `type_registry` argument on `NatsAdapterBase.__init__`; direct `serialize`/`deserialize` consumers pass a `resolver=` keyword argument. Lyra owns a single `TYPE_REGISTRY_RESOLVER` module constant in `src/lyra/nats/type_registry.py`. External `v0.1.x` consumers importing the removed private symbols hit a loud `ImportError` at load — the chosen failure mode over a silent `DeprecationWarning + no-op` shim that would have masked degraded type coercion.
 - **Public vs private submodule asymmetry** — lyra (as workspace host) imports `roxabi_nats._serialize`, `_sanitize`, etc. directly; external consumers are restricted to `__all__`. Python's underscore convention is social, not enforced — the contract is maintained via README and ADR documentation, not runtime guard.
 
 ### Neutral
--- a/packages/roxabi-nats/CHANGELOG.md
+++ b/packages/roxabi-nats/CHANGELOG.md
@@ -1,0 +1,87 @@
+# Changelog — roxabi-nats
+
+All notable changes to the `roxabi-nats` package are documented here.
+
+The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.2.0] — 2026-04-17
+
+### Breaking
+
+- **`NatsAdapterBase.__init__` gains a keyword-only `type_registry` parameter.**
+  Adapter subclasses must declare their TYPE_CHECKING-only type hints at construction:
+  ```python
+  adapter = MyAdapter(
+      subject="lyra.inbound.tg.main",
+      queue_group="lyra-hub",
+      envelope_name="InboundMessage",
+      schema_version=1,
+      type_registry=[
+          ("lyra.core.commands.command_parser", "CommandContext"),
+      ],
+  )
+  ```
+  Adapters that do not need TYPE_CHECKING resolution can pass `type_registry=None` (the default).
+- **`serialize`, `deserialize`, `deserialize_dict` gain a keyword-only `resolver` parameter.**
+  Callers that decode dataclasses carrying TYPE_CHECKING-only annotations MUST pass a configured `_TypeHintResolver`. Bare calls default to an empty resolver and will fail to coerce types whose hints are only visible under `TYPE_CHECKING`.
+- **`_register_type_checking_import` removed.**
+  The process-global `_TYPE_CHECKING_IMPORTS` registry is gone. Consumers must construct a `_TypeHintResolver` and pass it explicitly.
+- **`_TYPE_CHECKING_IMPORTS` module-level list removed** from `roxabi_nats._serialize`.
+
+### Added
+
+- `_TypeHintResolver` class with fail-fast validation. Invalid module paths or missing attributes raise `ValueError` at resolver construction, not at deserialize time.
+- `_EMPTY_RESOLVER` module-level singleton — immutable null-object used as the default resolver for bare `serialize`/`deserialize` calls.
+
+### Migration
+
+Before (v0.1.x):
+
+```python
+from roxabi_nats._serialize import _register_type_checking_import
+
+_register_type_checking_import(
+    "my.package.types", "MyTypeCheckingOnlyType"
+)
+
+adapter = MyAdapter(...)
+```
+
+After (v0.2.0):
+
+```python
+adapter = MyAdapter(
+    ...,
+    type_registry=[
+        ("my.package.types", "MyTypeCheckingOnlyType"),
+    ],
+)
+```
+
+Direct `serialize`/`deserialize` consumers:
+
+```python
+# Before
+from roxabi_nats._serialize import deserialize
+msg = deserialize(data, MyDataclass)
+
+# After — build one resolver at module scope, reuse it
+from roxabi_nats._serialize import _TypeHintResolver, deserialize
+
+_RESOLVER = _TypeHintResolver([
+    ("my.package.types", "MyTypeCheckingOnlyType"),
+])
+msg = deserialize(data, MyDataclass, resolver=_RESOLVER)
+```
+
+External `v0.1.x` consumers (voiceCLI, roxabi-vault, imageCLI) that imported the private `_register_type_checking_import` helper will hit `ImportError` on upgrade. Either migrate to `type_registry=` at adapter construction or pin `roxabi-nats = ">=0.1.0,<0.2.0"` until the migration is complete.
+
+## [0.1.0] — 2026-04-14
+
+### Added
+
+- Initial extraction from the Lyra monorepo as a uv workspace subpackage. See `docs/architecture/adr/045-roxabi-nats-sdk-uv-workspace-extraction.mdx`.
+- `NatsAdapterBase` — ABC lifecycle host for NATS request-reply adapters.
+- `nats_connect` — seed-auth-aware NATS connection helper.
+- Internal helpers: `_serialize` (type-aware JSON codec), `_sanitize`, `_validate`, `_version_check`, `_tts_constants`.
+- Full test suite covering adapter lifecycle, circuit breaker, readiness probe, serialization round-trip, and version-gate drop handling.
--- a/packages/roxabi-nats/CHANGELOG.md
+++ b/packages/roxabi-nats/CHANGELOG.md
@@ -23,15 +23,17 @@ The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and 
   ```
   Adapters that do not need TYPE_CHECKING resolution can pass `type_registry=None` (the default).
 - **`serialize`, `deserialize`, `deserialize_dict` gain a keyword-only `resolver` parameter.**
-  Callers that decode dataclasses carrying TYPE_CHECKING-only annotations MUST pass a configured `_TypeHintResolver`. Bare calls default to an empty resolver and will fail to coerce types whose hints are only visible under `TYPE_CHECKING`.
+  Callers that decode dataclasses carrying TYPE_CHECKING-only annotations MUST pass a configured `TypeHintResolver`. Bare calls default to an empty resolver and will fail to coerce types whose hints are only visible under `TYPE_CHECKING`.
 - **`_register_type_checking_import` removed.**
   The process-global `_TYPE_CHECKING_IMPORTS` registry is gone. Consumers must construct a `_TypeHintResolver` and pass it explicitly.
 - **`_TYPE_CHECKING_IMPORTS` module-level list removed** from `roxabi_nats._serialize`.
 
 ### Added
 
-- `_TypeHintResolver` class with fail-fast validation. Invalid module paths or missing attributes raise `ValueError` at resolver construction, not at deserialize time.
-- `_EMPTY_RESOLVER` module-level singleton — immutable null-object used as the default resolver for bare `serialize`/`deserialize` calls.
+- **Public `TypeHintResolver`** exported from the package root (`from roxabi_nats import TypeHintResolver`). Wraps the internal `_TypeHintResolver` class; external consumers should use the public alias rather than reaching into `_serialize` / `_resolver`.
+- `_TypeHintResolver` class with fail-fast validation. Invalid module paths or missing attributes raise `ValueError` at resolver construction, not at deserialize time. Duplicate `type_name` with different `module_path` also fails loud (was silently last-wins internally during development).
+- Internal singleton `_EMPTY_RESOLVER` — immutable null-object used as the default resolver for bare `serialize`/`deserialize` calls. `resolved` is a `MappingProxyType` so attempted mutation raises `TypeError`.
+- Resolver instances carry a per-instance monotonic `_uid`; the hint cache pairs `(dc_type, resolver._uid)` so GC'd resolvers cannot poison cache entries of new resolvers that happen to land at the same `id()` address.
 
 ### Migration
 
@@ -66,9 +68,10 @@ from roxabi_nats._serialize import deserialize
 msg = deserialize(data, MyDataclass)
 
 # After — build one resolver at module scope, reuse it
-from roxabi_nats._serialize import _TypeHintResolver, deserialize
+from roxabi_nats import TypeHintResolver
+from roxabi_nats._serialize import deserialize
 
-_RESOLVER = _TypeHintResolver([
+_RESOLVER = TypeHintResolver([
     ("my.package.types", "MyTypeCheckingOnlyType"),
 ])
 msg = deserialize(data, MyDataclass, resolver=_RESOLVER)
--- a/packages/roxabi-nats/pyproject.toml
+++ b/packages/roxabi-nats/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "roxabi-nats"
-version = "0.1.0"
+version = "0.2.0"
 description = "NATS transport SDK for Lyra and Roxabi plugin ecosystem"
 readme = "README.md"
 license = { text = "MIT" }
--- a/packages/roxabi-nats/src/roxabi_nats/__init__.py
+++ b/packages/roxabi-nats/src/roxabi_nats/__init__.py
@@ -9,11 +9,13 @@ contract. ``_``-prefixed submodules (``_serialize``, ``_sanitize``,
 may change without notice — external consumers MUST NOT import them.
 """
 
+from ._serialize import _TypeHintResolver as TypeHintResolver
 from .adapter_base import CONTRACT_VERSION, NatsAdapterBase
 from .connect import nats_connect
 
 __all__ = [
     "CONTRACT_VERSION",
     "NatsAdapterBase",
+    "TypeHintResolver",
     "nats_connect",
 ]
--- a/packages/roxabi-nats/src/roxabi_nats/_resolver.py
+++ b/packages/roxabi-nats/src/roxabi_nats/_resolver.py
@@ -1,0 +1,81 @@
+"""`_TypeHintResolver` — per-instance registry of TYPE_CHECKING-only types.
+
+Extracted from `_serialize.py` so that the serializer module stays within the
+300-line cap and the resolver has a single-responsibility home.  The public
+alias is re-exported from the package root as ``TypeHintResolver``
+(see `__init__.py`).
+"""
+
+from __future__ import annotations
+
+import importlib
+import itertools
+import types
+from collections.abc import Sequence
+from typing import Any
+
+# Per-instance monotonic counter — used as part of the `_hints_cache` key in
+# `_serialize._get_hints` so two resolvers never collide even if CPython
+# recycles `id()` after GC.  Plain `itertools.count()` is thread-safe for the
+# purpose of minting unique ints because of the GIL.
+_uid_counter = itertools.count()
+
+
+class _TypeHintResolver:
+    """Per-instance registry of TYPE_CHECKING-only types for deserialization.
+
+    Constructed at adapter/consumer init time. Eagerly imports every
+    (module_path, type_name) entry and caches the resolved type object so the
+    hot deserialization path never calls ``importlib``.  Non-existent
+    modules or attributes raise ``ValueError`` immediately — fail-fast at
+    construction, not on first message.
+
+    Each instance carries a monotonic ``_uid`` that callers pair with the
+    dataclass type to form the serializer's hint-cache key.  The UID is
+    unforgeable and never reused, so GC of a resolver cannot produce a stale
+    cache hit against a new resolver at the same ``id()`` address.
+    """
+
+    __slots__ = ("_uid", "entries", "resolved")
+
+    def __init__(self, entries: Sequence[tuple[str, str]]) -> None:
+        seen: set[tuple[str, str]] = set()
+        deduped: list[tuple[str, str]] = []
+        resolved: dict[str, type] = {}
+        for module_path, type_name in entries:
+            key = (module_path, type_name)
+            if key in seen:
+                continue
+            seen.add(key)
+            deduped.append(key)
+            if type_name in resolved:
+                prev_module = next(
+                    m for m, n in deduped if n == type_name and (m, n) != key
+                )
+                raise ValueError(
+                    f"type_registry: duplicate type_name {type_name!r} from "
+                    f"{module_path!r} conflicts with earlier entry "
+                    f"from {prev_module!r}"
+                )
+            try:
+                mod = importlib.import_module(module_path)
+            except ImportError as exc:
+                raise ValueError(
+                    f"type_registry: cannot import {module_path}"
+                ) from exc
+            if not hasattr(mod, type_name):
+                raise ValueError(
+                    f"type_registry: {module_path} has no attribute {type_name}"
+                )
+            resolved[type_name] = getattr(mod, type_name)
+        self.entries: tuple[tuple[str, str], ...] = tuple(deduped)
+        self.resolved: types.MappingProxyType[str, type] = types.MappingProxyType(
+            resolved
+        )
+        self._uid: int = next(_uid_counter)
+
+    def localns(self) -> dict[str, Any]:
+        return dict(self.resolved)
+
+
+_EMPTY_RESOLVER = _TypeHintResolver(())
--- a/packages/roxabi-nats/src/roxabi_nats/_resolver.py
+++ b/packages/roxabi-nats/src/roxabi_nats/_resolver.py
@@ -42,20 +42,20 @@ class _TypeHintResolver:
         seen: set[tuple[str, str]] = set()
         deduped: list[tuple[str, str]] = []
         resolved: dict[str, type] = {}
+        # name_to_module: type_name → module_path for the entry that first
+        # claimed it. O(1) lookup for the duplicate-name error; keeps the
+        # predicate unambiguous regardless of `seen`-set evolution.
+        name_to_module: dict[str, str] = {}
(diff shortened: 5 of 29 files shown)


MickaelV0:
## Code Review

**Reviewers (parallel, fresh context):** architect · backend-dev · product-lead · security-auditor · tester
**Diff:** 23 files, 1407 lines · **Spec compliance:** all 17 success criteria met
**Verdict:** 🔄 **Request changes** — 2 blocking findings, 1 non-blocking cache-key concern, several suggestions

---

### Blockers

#### issue(blocking): `NatsRenderEventCodec.decode` is `@staticmethod` — instance `self._resolver` is dead code

  `src/lyra/nats/render_event_codec.py:43-44, 59-66, 92` · `src/lyra/adapters/nats_stream_decoder.py:92`
  -- architect, product-lead
  **Root cause:** `__init__` now takes a `resolver` kwarg and stores `self._resolver`, but `decode()` kept the old `@staticmethod` decorator plus its own `resolver=` default. The only caller (`nats_stream_decoder.py:92`) invokes `NatsRenderEventCodec.decode(event_type, payload, counter=counter)` on the class — no instance, no forwarded resolver. `self._resolver` is stored but never read. Functionally safe because the static default is `TYPE_REGISTRY_RESOLVER`, but the instance-resolver injection is silently non-functional — future callers who try to inject a custom resolver via `NatsRenderEventCodec(resolver=alt)` will get ignored.
  **Solutions:**
    1. Drop `@staticmethod`, convert to `def decode(self, event_type, payload, *, counter=None)`, use `self._resolver` internally, remove the redundant `resolver` kwarg on the method. Update the call site to `codec.decode(...)` on a constructed instance (recommended).
    2. Keep `decode` static and remove `__init__` + `self._resolver` entirely — the method's `resolver` kwarg becomes the full API.
  **Confidence:** 95%

#### issue(blocking): `_TypeHintResolver` (private `_`-prefixed) now leaks into the public API surface of ~7 lyra files

  `src/lyra/nats/type_registry.py:24` plus every lyra consumer with `from roxabi_nats._serialize import _TypeHintResolver` (nats_bus, render_event_codec, nats_channel_proxy, nats_outbound_listener, _inbound_cache, nats_envelope_handlers)
  -- architect
  **Root cause:** ADR-045 allowed lyra (workspace host) to import from `_`-prefixed SDK submodules as a documented exception. That exception held when only `_serialize.deserialize_dict` was imported. Now `_TypeHintResolver` appears in public type annotations on constructors (`resolver: _TypeHintResolver = …`), so a `_`-prefixed type is effectively part of the public contract of ~7 lyra classes. This contradicts the SDK's own convention (PR #724 explicitly said "external consumers MUST NOT import `_`-prefixed submodules"). Since the type is now a first-class public API point, it should shed the underscore.
  **Solutions:**
    1. Add `from roxabi_nats._serialize import _TypeHintResolver as TypeHintResolver` to `packages/roxabi-nats/src/roxabi_nats/__init__.py`, append to `__all__`, and update the 7 lyra imports to use the public alias. One-line addition to the SDK, 7 import tweaks in lyra (recommended).
    2. Keep private, document in ADR-045 as an expanded workspace-host exception. No code change, but the API contract becomes inconsistent with the extraction story.
  **Confidence:** 82%

---

### Non-blocking issues

#### issue: `id(resolver)` cache key is vulnerable to CPython address reuse after GC

  `packages/roxabi-nats/src/roxabi_nats/_serialize.py:28, 146`
  -- security-auditor, architect, backend-dev (three independent flags)
  **Root cause:** The plan's Notes & Risks section explicitly called this out ("Fix: key the cache on `(dc_type, id(resolver))`") and T2 implemented exactly that, but `id()` is the memory address — a GC'd resolver can have its slot reused by a new, differently-configured resolver. The cache would then serve stale hints. In practice every resolver in this codebase is either a module-level singleton or `self._resolver` on a long-lived adapter, so the window is practically closed — but short-lived test resolvers rotate through id space. Zero production exposure today; reviewer-quality concern.
  **Solutions:**
    1. Add `_uid_counter = itertools.count()` at module scope and `self._uid = next(_uid_counter)` in `__init__`, then key cache on `(dc_type, self._uid)`. Two-line change, closes the loop (recommended).
    2. Use `WeakKeyDictionary[_TypeHintResolver, dict[type, dict]]` — more complex but self-cleaning on GC.
    3. Accept and document the "resolver lifetime ≥ cache entry lifetime" invariant in the `_get_hints` docstring.
  **Confidence:** 78%

#### issue: `type_registry=[]` takes the `_EMPTY_RESOLVER` fast path via truthy check, not an explicit `None` check

  `packages/roxabi-nats/src/roxabi_nats/adapter_base.py:81`
  -- backend-dev
  **Root cause:** `_TypeHintResolver(type_registry) if type_registry else _EMPTY_RESOLVER` — for `type_registry=[]` this hands back `_EMPTY_RESOLVER` without constructing a resolver. Correct outcome, surprising path. The annotation is `Sequence[tuple[str, str]] | None`, implying `[]` is a valid empty sequence. A future refactor that needs to distinguish "no entries explicitly configured" vs "not configured" will stumble.
  **Solutions:**
    1. `_TypeHintResolver(type_registry) if type_registry is not None else _EMPTY_RESOLVER` — explicit None guard; `_TypeHintResolver([])` is cheap (recommended).
    2. Keep current; add comment: `# truthy check intentionally collapses None and [] to _EMPTY_RESOLVER`.
  **Confidence:** 92%

#### issue: `_TypeHintResolver` silently clobbers on duplicate `type_name` with different `module_path`

  `packages/roxabi-nats/src/roxabi_nats/_serialize.py:44-61`
  -- backend-dev
  **Root cause:** `[("pkg_a", "Foo"), ("pkg_b", "Foo")]` — the `seen` set keeps both (distinct tuples), so `entries == (("pkg_a","Foo"), ("pkg_b","Foo"))`, but `resolved["Foo"]` gets overwritten to `pkg_b.Foo`. Silent last-write-wins — no caller in this codebase hits it today, but a multi-domain registry would produce subtle wrong-type coercion.
  **Solutions:**
    1. Detect before overwrite: `if type_name in resolved: raise ValueError(f"type_registry: duplicate type_name {type_name!r} from {module_path!r} conflicts with earlier entry from {prev_path!r}")` (recommended).
    2. Document last-wins as intentional.
  **Confidence:** 85%

#### issue: `_test_stub_module.py` ships inside the production wheel

  `packages/roxabi-nats/src/roxabi_nats/_test_stub_module.py`
  -- security-auditor
  **Root cause:** `[tool.hatch.build.targets.wheel] packages = ["src/roxabi_nats"]` includes every `.py` under `src/roxabi_nats/` — so `StubInner` is importable by any downstream consumer as `roxabi_nats._test_stub_module`. Harmless class today, but the pattern makes it easy to accidentally ship future test-only stubs with mutable state or internal hooks.
  **Solutions:**
    1. Move stub to `packages/roxabi-nats/tests/stubs/stub_module.py`; register it on `sys.path` via a `tests/conftest.py` fixture or add the path so the resolver's import of `"roxabi_nats._test_stub_module"` → `"stubs.stub_module"` still works (recommended).
    2. Add `[tool.hatch.build.targets.wheel] exclude = ["src/roxabi_nats/_test_stub_module.py"]` — fastest; keeps the file colocated with prod code.
  **Confidence:** 80%

#### issue: clean-break guard tests weaker than spec literal + cache-isolation regression not covered

  `packages/roxabi-nats/tests/test_type_hint_resolver.py:48-58, 127-142, 150-153`
  -- tester
  **Root cause:**
  - `test_resolver_resolves_stub_type` uses `_StubOuter(name="x", inner=None)` — never exercises `StubInner` resolution; `_decode_union` short-circuits on None before touching the resolver. Test passes with `_EMPTY_RESOLVER` too.
  - `test_empty_resolver_singleton_is_module_level` asserts `entries == ()` only — doesn't verify `resolved` is a `MappingProxyType` nor that it's empty. A future refactor swapping `MappingProxyType` for a plain `dict` wouldn't be caught.
  - No test catches the `id(resolver)` cache-key regression the plan flagged. Reverting `_hints_cache` to `dict[type, ...]` passes the entire suite today.
  **Solutions:**
    1. (a) Serialize `_StubOuter(name="x", inner=StubInner())` and assert `isinstance(result.inner, StubInner)`. (b) Add `assert isinstance(_EMPTY_RESOLVER.resolved, types.MappingProxyType)` + `with pytest.raises(TypeError): _EMPTY_RESOLVER.resolved["x"] = 1`. (c) Add a cache-isolation test: deserialize with `_EMPTY_RESOLVER` first, then the non-empty resolver — assert the results differ (one has typed `inner`, other doesn't). All three land in the same test file (recommended).
    2. Drop (c); the runtime correctness is already implemented, only the regression guard is weak.
  **Confidence:** 90%

---

- `nitpick:` `serialize()` accepts `resolver=` but ignores it. Docstring covers this; renaming to `_resolver_unused` would be legible at the call site but noisy. Not worth it. (backend-dev, 70%)
- `nitpick:` `NatsChannelProxy` threads resolver through encode-path `serialize()` calls as pure overhead. Spec-approved for API symmetry; no action. (architect, backend-dev, 80%)
- `suggestion:` `_get_hints` globalns filter removes from `localns` if name is in `globalns`, regardless of whether the `globalns` entry is a type. PascalCase convention makes real collisions near-impossible. Narrowing to `isinstance(globalns[type_name], type)` would harden it. (backend-dev, 65%)
- `suggestion:` `ValueError` on resolver construction leaks `module_path` via `from exc` chain. Zero-risk today (all callers use hard-coded literals); forward-looking mitigation if any future caller ever builds `module_path` from config. (security-auditor, 65%)
- `thought:` `_EMPTY_RESOLVER` immutability: `MappingProxyType` locks the mapping but `object.__setattr__(resolver, 'resolved', ...)` can still replace the slot. Spec claim is a minor overstatement; not exploitable, not practically reachable. (security-auditor, product-lead)
- `suggestion:` `test_nats_bus_defaults_to_type_registry_resolver` passes `nc=None  # type: ignore` — use `MagicMock(spec=NATS)` instead to future-proof against `__init__` adding any `nc`-dependent validation. (tester, 75%)
- `suggestion:` Add a direct `NatsBus.handler()` coverage test that decodes a serialized `InboundMessage` through the instance resolver path, not just the module-level `deserialize` helper. Existing tests verify `bus._resolver is TYPE_REGISTRY_RESOLVER` but don't exercise the actual handler's call site. (tester, 85%)

---

### Praise

- `praise:` Genuine security improvement — the global `_TYPE_CHECKING_IMPORTS` list is gone, `importlib.import_module` runs exactly once per entry at a single deterministic call site, no caller-controlled strings flow through the hot path. (security-auditor)
- `praise:` `NatsAdapterBase` fail-fast validation at construction is a meaningful upgrade over the old deserialize-time failure mode. Pattern matches the existing `assert CONTRACT_VERSION.isdigit()` precedent. (security-auditor)
- `praise:` `__slots__` + `MappingProxyType` on `_TypeHintResolver` is the right shape — more precise than `@dataclass(frozen=True)` for a class with a mapping attribute, and avoids `NamedTuple._replace()` escape hatch. (architect)
- `praise:` No import cycle risk in the default-kwarg chain `NatsBus(resolver=TYPE_REGISTRY_RESOLVER)` — verified. (architect)
- `praise:` `_EMPTY_RESOLVER` as a module-level null-object singleton: zero-iteration constructor, identity-checkable, cheap. (backend-dev)
- `praise:` All 17 spec Success Criteria met; release signals correct (pyproject 0.2.0, CHANGELOG `[0.2.0] Breaking` + Migration, ADR-045 Consequences updated with clean-break narrative). (product-lead)

---

### Verdict

**🔄 Request changes** — two blocking issues:
1. `NatsRenderEventCodec.decode` is still a `@staticmethod` while its class now carries a `_resolver` field — dead code + inconsistent design surface.
2. `_TypeHintResolver` as a `_`-prefixed type in ~7 lyra public method signatures contradicts the SDK's own convention and the PR #724 extraction story — promote to public alias in `roxabi_nats/__init__.py`.

After fixes, the other `issue:` items (cache-key GC safety, `type_registry=[]` semantics, duplicate type_name handling, test stub packaging, test-coverage gaps) can land either in this PR or as follow-ups — none of them are regressions and all have documented low real-world exposure.

MickaelV0:
## Review Fixes Applied

**Auto-applied (review-driven):** 7 findings
**Deferred (kept with docstring/pragma):** 2 suggestions
**Follow-ups created:** 0 — all findings in scope for this PR

### Applied

1. **[applied] issue(blocking)** — `NatsRenderEventCodec.decode` is now an instance method using `self._resolver`. The `@staticmethod` + duplicate-`resolver=` kwarg pattern made `__init__` dead code. Single call site (`nats_stream_decoder.py`) updated to construct the codec once and reuse it. `95%` confidence (architect + product-lead).
2. **[applied] issue(blocking)** — `TypeHintResolver` promoted to public SDK alias in `roxabi_nats/__init__.py`. Every lyra consumer switched from `from roxabi_nats._serialize import _TypeHintResolver` to `from roxabi_nats import TypeHintResolver`. Internal SDK keeps the `_`-prefixed class in `_resolver.py`. `82%` confidence (architect).
3. **[applied] issue(F3)** — `_hints_cache` key replaced `id(resolver)` → `resolver._uid` (monotonic counter from `itertools.count()`). No more GC-address-reuse hazard. Unanimous A1 pick across architect + devops + security-auditor consensus. `78–92%` confidence across 3 reviewers.
4. **[applied] issue(F6)** — `_test_stub_module.py` moved out of `src/roxabi_nats/` entirely. New home: `packages/roxabi-nats/tests/_stub_fixture.py`. Conftest registers it under `sys.modules["roxabi_nats_test_stub"]` at collection time via `importlib.util.spec_from_file_location`. Hatch wheel target `packages = ["src/roxabi_nats"]` now physically cannot ship it — zero wheel surface. Unanimous B1 pick across 3-PoV consensus.
5. **[applied] issue(F4)** — `type_registry=[]` now takes the resolver-construction path explicitly (`if type_registry is not None`). Empty list no longer silently aliases to `_EMPTY_RESOLVER`. Semantic ambiguity closed.
6. **[applied] issue(F5)** — Duplicate `type_name` with different `module_path` raises `ValueError` at construction. Previously last-write-wins silently clobbered `resolved[name]`. Fail-loud invariant with test coverage.
7. **[applied] issue(tests)** — Four test gaps closed:
   - `test_resolver_resolves_stub_type` now builds `_StubOuter(name="x", inner=StubInner())` and asserts `isinstance(result.inner, StubInner)` — exercises the resolver-driven coercion path that the previous `inner=None` assertion skipped.
   - `test_empty_resolver_singleton_is_module_level` now asserts `MappingProxyType` + `TypeError` on mutation.
   - New `test_hints_cache_isolated_across_resolvers` defends the monotonic-UID cache key against regression.
   - New `test_type_registry_duplicate_type_name_rejected` defends F5.

### Architectural bonus (consensus-driven, not in review)

- **`_TypeHintResolver` extracted to `_resolver.py`.** `_serialize.py` had crept to 350 lines, over the project's hard 300-line cap. The split aligns the module boundary with the logical boundary — resolver is a standalone component, serializer uses it. Unanimous C1 pick across 3-PoV consensus. `_serialize.py` now at 300, `_resolver.py` at 81.

### Deferred (with rationale)

- **nitpick — `serialize()` accepts `resolver=` but ignores it.** Docstring already notes it's for API symmetry. Renaming to `_resolver_unused` would be noisy at call sites. Keeping as-is per the spec's N3 design note. `70%` confidence — not a correctness concern.
- **suggestion — `_get_hints` globalns filter could narrow to `isinstance(..., type)`.** PascalCase convention makes collision near-impossible; current filter is correct for the stated intent. `65%` confidence — hypothetical only.

### Gates after fixes

- `uv run pytest` → **2974 passed, 76 skipped** (added 2 new tests vs. pre-fix)
- `uv run pyright` → **0 errors, 0 warnings**
- `uv run ruff check .` → **All checks passed**
- `find … -name '*.py' -exec wc -l {} + | awk '$1 > 300'` → **0 matches**

All blocking findings addressed. Ready for re-review or merge.

MickaelV0:
## Code Review (iteration 2/2)

**Reviewers (parallel, fresh context):** architect · backend-dev · tester · security-auditor
**Scope:** fix commit `1c8f2ad` verifying B1, B2, F3, F4, F5, F6, C1 + new tests
**Verdict:** ✅ **Approve with comments** — all iter-1 blockers landed correctly; 5 non-blocking structural/polish items

---

### Verification of claimed fixes

| Fix | Claim | Verdict |
|---|---|---|
| B1 | `NatsRenderEventCodec.decode` instance method | ✅ Landed (architect, backend-dev, security-auditor) — `self._resolver` threaded through `deserialize()`; no static-default bypass remains |
| B2 | `TypeHintResolver` public alias | ✅ Landed (architect) — 7 lyra files switched; `_`-prefixed internal class preserved |
| F3 | `resolver._uid` monotonic counter | ✅ Landed (architect, backend-dev, security-auditor) — `itertools.count()` on a module-level singleton; GIL-atomic `next()`; GC-reuse window closed structurally |
| F4 | `type_registry is not None` explicit | ✅ Landed (backend-dev) |
| F5 | duplicate `type_name` fail-loud | ✅ Landed (backend-dev, security-auditor) — prevents silent type-confusion |
| F6 | stub out of `src/` | ✅ Landed (security-auditor) — zero `_test_stub_module` references in `src/`; conftest registration idempotent + safe across xdist workers |
| C1 | `_serialize.py` split into `_resolver.py` | ✅ Landed (architect) — `_serialize.py` 300, `_resolver.py` 81. No file exceeds 300 lines. |
| Tests | 2 new + 2 updated | ✅ Landed; 2 have structural weaknesses (see below) |

No regressions. `pytest 2974/76`, `pyright 0 errors`, `ruff clean`.

---

### Non-blocking findings

#### issue: `deduped.append(key)` ordering in `_resolver.py` duplicate-name check

  `packages/roxabi-nats/src/roxabi_nats/_resolver.py:50-58`
  -- architect
  **Root cause:** `deduped.append(key)` runs on line 50 (after the `seen` check) BEFORE the `type_name in resolved` conflict check on line 51. When the generator `next(m for m, n in deduped if n == type_name and (m, n) != key)` runs to produce `prev_module`, it searches a list that already contains the current `key`. The `(m, n) != key` filter rescues correctness today — the current entry is excluded so the earlier entry surfaces. Currently harmless because the `seen` guard makes the same `(module, name)` pair unreachable, but the predicate is fragile: any future relaxation of the `seen` set would cause `next()` to match the current entry before the prior one and return the wrong module in the error message.
  **Solutions:**
    1. Move `deduped.append(key)` AFTER `resolved[type_name] = getattr(mod, type_name)` — `deduped` only contains committed entries when searched; predicate becomes unambiguous (recommended).
    2. Track `type_name → module_path` in a separate `seen_names: dict[str, str]` for O(1) lookup without generator scan.
  **Confidence:** 92%

#### issue: `_hints_cache` unbounded growth over process lifetime

  `packages/roxabi-nats/src/roxabi_nats/_serialize.py:31`
  -- backend-dev
  **Root cause:** `_hints_cache: dict[tuple[type, int], dict]` accumulates one entry per `(dc_type, resolver._uid)` pair with no eviction. UIDs increase monotonically; cache is never pruned. In production the resolver count is bounded (one per adapter, process-lifetime singletons), so growth is `O(adapter_count × dataclass_count)` — benign. In a long-running test harness that churns resolvers, it is a measurable-but-not-fatal leak.
  **Solutions:**
    1. Add a comment/docstring noting the invariant: "production resolvers are process-lifetime singletons; cache growth is bounded by `adapter_count × dataclass_count`". Zero code change (recommended for this PR).
    2. Register `weakref.finalize(resolver, _evict, resolver._uid)` to auto-evict on GC — higher complexity, only worth it if resolver churn is measured as a real problem.
  **Confidence:** 80%

#### issue: `test_hints_cache_isolated_across_resolvers` misses the canonical poisoning order

  `packages/roxabi-nats/tests/test_type_hint_resolver.py:191-203`
  -- tester
  **Root cause:** The test deserializes with `r_non_empty` FIRST, then `r_empty`. The canonical cache-poisoning scenario is: resolver A caches a result for `dc_type`, resolver B queries `dc_type`, wrong cached result returned. Under the bug (`dc_type`-only cache key), the CURRENT ordering (non-empty first) would still catch it (empty resolver reads non-empty hints, gets `StubInner`, assertion fails). But the mirror ordering (empty first, non-empty second) is the more commonly documented regression scenario. Not a test that's wrong — a test that happens to work in both directions but documents only one.
  **Solutions:**
    1. Add a mirror test that runs empty-then-non-empty, asserting the non-empty resolver still coerces `StubInner` after the empty resolver cached `{}` (recommended).
    2. Add a docstring note that the current ordering catches the regression because the assertions are sensitive to both directions; leave as-is.
  **Confidence:** 85%

#### nitpick: alias-not-rename — tracebacks still show `_TypeHintResolver`

  `packages/roxabi-nats/src/roxabi_nats/__init__.py:12`
  -- architect
  **Root cause:** `from ._serialize import _TypeHintResolver as TypeHintResolver` keeps `__name__ = "_TypeHintResolver"` and `__qualname__ = "_TypeHintResolver"` on the class object. Pyright hovers and Python tracebacks display the private name even though users import the public alias. Two-hop alias chain (`_resolver → _serialize → __init__`) adds navigation overhead in IDEs.
  **Solutions:**
    1. Rename the class to `TypeHintResolver` directly in `_resolver.py`; keep underscore on `_uid_counter` and `_EMPTY_RESOLVER`. Tracebacks + pyright both show the public name. One-shot rename across the file (recommended).
    2. Set `__name__ = "TypeHintResolver"` / `__qualname__ = "TypeHintResolver"` on the class body — cosmetic fix without a rename.
  **Confidence:** 75%

#### nitpick: `test_type_registry_duplicate_type_name_rejected` fragile second entry

  `packages/roxabi-nats/tests/test_type_hint_resolver.py:208-213`
  -- tester
  **Root cause:** Second entry is `("nonexistent.path", "StubInner")`. The duplicate check fires BEFORE `importlib.import_module` is called on `nonexistent.path`, so the non-existent path never needs to resolve. Relies on a specific check-ordering in `_TypeHintResolver.__init__`. If someone reorders to import-before-dedup-check, the test breaks with `ValueError("cannot import nonexistent.path")` instead of `"duplicate type_name"` — misleading failure.
  **Solutions:**
    1. Use an importable second module (add a second class to `_stub_fixture.py` or use a stdlib path), so both entries are always valid for import and only the duplicate-name check can fire (recommended).
    2. Add a comment documenting the check-ordering dependency.
  **Confidence:** 78%

#### nitpick: `NatsRenderEventCodec` allocated per-stream in `nats_stream_decoder.py`

  `src/lyra/adapters/nats_stream_decoder.py:61`
  -- architect
  **Root cause:** `_codec = NatsRenderEventCodec()` inside `decode_stream_events` means a fresh codec is constructed per stream invocation. Codec constructor is cheap (single attribute assignment to `TYPE_REGISTRY_RESOLVER`), but a module-level `_CODEC = NatsRenderEventCodec()` would make the singleton sharing explicit and eliminate the allocation.
  **Solutions:**
    1. Hoist to module scope: `_CODEC = NatsRenderEventCodec()` at top of `nats_stream_decoder.py`; use `_CODEC.decode(...)` in the generator (recommended).
    2. Keep as-is — allocation cost is negligible, codec is stateless beyond the resolver reference.
  **Confidence:** 80%

---

### Praise

- All 7 claimed fixes from `1c8f2ad` landed correctly. No regressions, no new security findings, no file exceeds 300 lines. (security-auditor, architect, backend-dev)
- `_resolver.py` split is a clean module-boundary-meets-logical-boundary refactor; `_uid_counter` placement is safe (GIL-atomic `next()`, no ordering hazard). (architect)
- Conftest `sys.modules` registration is idempotent, robust across `pytest -p no:cacheprovider` and `pytest-xdist` workers, and uses the precise `importlib.util.spec_from_file_location` + `exec_module` pattern. (architect)
- `F5` fail-loud duplicate check is a real threat-model improvement (prevents silent type-confusion); `F6` stub relocation physically excludes the test fixture from every built wheel regardless of future `pyproject.toml` edits. (security-auditor)
- Test strengthening for `_EMPTY_RESOLVER` (MappingProxyType assertion + TypeError-on-mutation) verifies the real immutability contract against the source. (tester)

---

### Verdict

**✅ Approve with comments** — iter-1 blockers resolved, no remaining correctness issues. The 5 polish items are either cosmetic (`_hints_cache` growth comment, alias rename, test second-entry, per-stream codec allocation) or structural fragility that's currently harmless (deduped-append ordering, cache-isolation test direction).

Loop cap (iter 2/2) reached — any further polish should either land in this PR directly or as a follow-up issue. Not a re-review candidate.

PR #802, opened 2026-04-19
chore(ci): add workflow_dispatch + concurrency to ci.yml

### LINKED ISSUE #796: chore(ci): add workflow_dispatch + concurrency to ci.yml
## Context

During PR #793 (#768 pytest include + fixture provenance) a push to the feature branch silently failed to dispatch a CI run. The commit was ~45 minutes stale, the branch was pushed and visible on origin, the PR was synchronized (GitHub acknowledged the new head SHA), the Actions workflow was active — yet no workflow run was queued. Closing + reopening the PR did not re-trigger it either.

The only reliable escape hatch was an empty commit to force a new SHA.

## Root cause (hypothesis)

GitHub occasionally drops \`pull_request.synchronize\` events under load or when a \`push\` event for the same SHA coalesces with the \`pull_request.synchronize\` and one "wins." With no \`workflow_dispatch\` trigger and no concurrency group, there is no way to re-queue the dropped run short of generating a new SHA.

## Proposed change

\`\`\`yaml
# .github/workflows/ci.yml

on:
  push:
    branches: [main, staging]
  pull_request:
    branches: [main, staging]
  workflow_dispatch: {}        # manual re-trigger: gh workflow run ci.yml --ref <branch>

concurrency:
  group: ci-\${{ github.ref }} # per-ref de-duplication
  cancel-in-progress: true     # force-push or fast follow-up cancels stale runs
\`\`\`

## Why

- **workflow_dispatch** — adds \`gh workflow run ci.yml --ref feat/<slug>\` as a one-liner escape hatch. Next time an event is dropped we never need the close/reopen dance or the empty commit hack.
- **concurrency.cancel-in-progress** — when a force-push or rapid follow-up lands, the in-flight run is cancelled and only the latest SHA runs. Reduces wasted minutes and sidesteps the coalescing bug by making the scheduling intent explicit instead of relying on GH's implicit de-duplication.

## Acceptance Criteria

- [ ] \`.github/workflows/ci.yml\` gains \`workflow_dispatch: {}\` at the top-level \`on:\` map
- [ ] \`concurrency\` block added with \`group: ci-\${{ github.ref }}\` and \`cancel-in-progress: true\`
- [ ] Verified: \`gh workflow run ci.yml --ref <feature-branch>\` dispatches a run
- [ ] Verified: pushing two commits in rapid succession cancels the first run and only the latest completes

## References

- Surfaced during PR #793 / issue #768 (lost \`pull_request.synchronize\` event)

MickaelV0:
## Summary
- Add `workflow_dispatch: {}` to `on:` — manual re-queue escape hatch (`gh workflow run ci.yml --ref <branch>`) for dropped `pull_request.synchronize` events.
- Add `concurrency: { group: ci-${{ github.ref }}, cancel-in-progress: true }` — per-ref de-duplication; force-push/fast follow-up cancels the in-flight run and only the latest SHA completes.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #796: chore(ci): add workflow_dispatch + concurrency to ci.yml | OPEN |
| Implementation | 1 commit on `feat/796-ci-workflow-dispatch-concurrency` | Complete |
| Verification | Lint ✅ Typecheck ✅ YAML parses ✅ | Passed |

## Test Plan
- [ ] `gh workflow run ci.yml --ref feat/796-ci-workflow-dispatch-concurrency` dispatches a run
- [ ] Pushing two commits in rapid succession cancels the first run; only the latest completes
- [ ] Normal `push` / `pull_request` triggers still fire as before

Closes #796

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,12 @@ on:
     branches: [main, staging]
   pull_request:
     branches: [main, staging]
-  workflow_dispatch: {}
+  workflow_dispatch: {}  # dispatch resolves from default branch — per-branch re-triggers require the branch to contain a compatible ci.yml
 
 concurrency:
   group: ci-${{ github.ref }}
-  cancel-in-progress: true
+  # Cancel only on feature branches; protected branches (main, staging) run to completion so every merge-commit SHA has a recorded status.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/staging' }}
 
 jobs:
   ci:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,11 @@ on:
     branches: [main, staging]
   pull_request:
     branches: [main, staging]
+  workflow_dispatch: {}
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   ci:


MickaelV0:
## Code Review

**Scope:** 5 added lines in `.github/workflows/ci.yml` — `workflow_dispatch: {}` trigger + workflow-level `concurrency` group.

### ⚠ Blockers (1)

**issue:** `cancel-in-progress: true` on `push` to `main` / `staging` leaves superseded SHAs with a stuck cancelled status
- `.github/workflows/ci.yml:15`
- **-- devops**
- **Root cause:** `cancel-in-progress: true` is safe on feature branches where the head commit always supersedes. On protected integration branches every commit is independently valuable — cancelling the in-flight run for an older SHA records a failed/cancelled status against that commit. If branch protection requires CI, the older SHA is permanently stuck.
- **Solutions:**
  1. Exempt `main`/`staging` pushes from cancellation using a conditional: `cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/staging' }}` — keeps cancellation active for PRs and feature-branch dispatches while protecting merge commits. (recommended)
  2. Accept the risk — direct pushes to protected branches are infrequent and always supersede. Document the assumption in a comment above `cancel-in-progress`.
- **Confidence:** 90%

### Suggestions (4, non-blocking)

**suggestion:** Make concurrency-key intent explicit by including event type
- `.github/workflows/ci.yml:14`
- **-- devops**
- **Root cause:** `github.ref` is technically correct — `push` and `workflow_dispatch` on the same branch share a slot (intentional), while `pull_request` runs use `refs/pull/<N>/merge` and are already isolated per PR. But the current key does not document this grouping logic.
- **Solutions:**
  1. Tighten to `ci-${{ github.event_name }}-${{ github.ref }}` so a `workflow_dispatch` does not cancel a concurrent `push` run (auditable intent).
  2. Keep `github.ref` — it already produces correct behaviour for this trigger set.
- **Confidence:** 85%

**suggestion(non-blocking):** Workflow-level `cancel-in-progress` can preempt the `secrets` (TruffleHog) job mid-scan
- `.github/workflows/ci.yml:92-103`
- **-- devops / security-auditor** (consolidated)
- **Root cause:** `secrets` runs in parallel with `ci` (no `needs:`). Under workflow-level concurrency, a rapid second push cancels all in-flight jobs for that ref, including a still-running TruffleHog scan. Cancelled scans emit no status — so a secret pushed then rapidly superseded could slip through the gap (the second run does re-scan the full history with `fetch-depth: 0`, mitigating this in practice).
- **Solutions:**
  1. Move `secrets` to a separate workflow file (`secret-scan.yml`) with its own group and `cancel-in-progress: false` to decouple security scanning from CI lifecycle.
  2. Add `needs: [ci]` to the `secrets` job so it sequences after `ci` completes (reduces the cancellation window).
  3. Accept — the next push triggers a fresh full scan.
- **Confidence:** 75% (merged 65% + 80% → midpoint)

**nitpick:** `workflow_dispatch` scope constraint worth documenting
- `.github/workflows/ci.yml:11`
- **-- devops**
- **Root cause:** GitHub resolves `workflow_dispatch` targets using the workflow file from the **default branch** (`main`). A branch-only version of `ci.yml` cannot be dispatched via `gh workflow run ci.yml --ref <feature-branch>` until the file exists on `main` in a compatible form. This is a GitHub platform constraint, not a bug.
- **Solutions:**
  1. Add an inline comment next to `workflow_dispatch: {}`: `# resolves from default branch; for per-branch re-triggers the file must be on main`.
  2. No change — users will discover the constraint if they hit it.
- **Confidence:** 95%

### Praise (2)

**praise:** `workflow_dispatch: {}` adds no user-input surface, `permissions: contents: read` is correctly scoped, all external actions remain SHA-pinned, and `github.ref`-keyed concurrency groups are PR-isolated — no cross-PR collision risk. **-- security-auditor**

**praise:** Change matches the issue body verbatim; minimal diff; issue #796 acceptance criteria (both trigger + concurrency block) satisfied. **-- code-review**

---

### Verdict: **Request changes** (one `issue:` blocker on protected-branch cancellation)

The single blocker is a real concern for `main` / `staging` pushes. Contributor options:
- **Fix now (`/fix`)** — apply Solution 1 to the blocker (conditional `cancel-in-progress` excluding protected branches), optionally address suggestions.
- **Merge as-is** — accept the trade-off documented in the issue (user explicitly chose this shape); protected-branch pushes are rare and are expected to supersede cleanly.

_Agents: security-auditor, devops (architect/product-lead/tester/frontend-dev/backend-dev skipped — out of scope)_

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 2 findings (bundle B, long-term recommended)
**Deferred (issues created):** 1 finding → #803
**Skipped:** 1 finding
**Failed:** 0

### Applied (1b1)
- [applied] **issue** `cancel-in-progress: true` on protected branches (90%) — now conditional: `${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/staging' }}`. Feature branches still cancel-supersede; `main`/`staging` push runs complete so every merge-commit SHA records a real status.
- [applied] **nitpick** `workflow_dispatch` default-branch constraint (95%) — inline comment added.

### Deferred
- **suggestion** workflow-level cancel preempts TruffleHog `secrets` job mid-scan (75%) → #803 (split into standalone `secret-scan.yml` with `cancel-in-progress: false`).

### Skipped
- **suggestion** tighten concurrency key with `github.event_name` (85%) — current `github.ref`-only key is correct; collapsing `push` + `workflow_dispatch` on the same branch is the intended behaviour. Splitting them would duplicate CI minutes on manual dispatch.

Commit: f1f016a

Issue #728, opened 2026-04-15
feat(infra): containerize voicecli workers (GPU passthrough) — supersedes #609


MickaelV0:
## Parent

Part of #605 (Containerize production — Podman + Quadlet)

## Supersedes

**#609** (closed) — "stt-adapter.container + tts-adapter.container with voiceCLI socket mounts" modeled voice as **lyra adapters** delegating to host voiceCLI daemons via Unix sockets. That premise is obsolete after:

- **#691** (refactor) — voice path cutover to `voicecli nats-serve {stt,tts}` workers; `lyra_stt` / `lyra_tts` satellites deleted (#690), `voicecli` dep dropped from `pyproject.toml` (#691)
- **#724** (PR) + ADR-045 — `roxabi-nats` SDK extracted as uv workspace subpackage; voicecli workers consume it directly as first-class NATS participants

The remaining question is no longer "containerize the lyra voice adapters" but **"containerize the voicecli workers themselves"** — a different architectural problem (GPU passthrough, VRAM management, model cache volumes).

## Blocked by

- #691 — voicecli nats-serve must be verified in prod before we containerize it
- #727 — podman bootstrap on Machine 1

## Open design questions

1. **GPU passthrough** — use `--device nvidia.com/gpu=all` via CDI (Container Device Interface), or `--gpus all` legacy runtime hook?
2. **VRAM sharing** — voicecli STT + voicecli TTS + imageCLI + future LLM worker all want the same GPU. Does containerization make this worse, better, or neutral vs current bare-metal scheduling?
3. **Model cache** — voicecli downloads large models (Chatterbox TTS, Whisper, etc.). Quadlet volume mounted from `~/.cache/voicecli/` or named volume?
4. **Boot time** — GPU warmup + model load takes tens of seconds. How does `Restart=always` + systemd `RestartSec` interact? Any impact on `make voice-smoke` SLA?
5. **Image ownership** — do we build a `localhost/voicecli:latest` image in the voiceCLI repo (mirror of lyra's `#610`), or does this repo build a composite image? ADR needed.
6. **Scope boundary** — this repo or `voiceCLI` repo? Probably voiceCLI repo with lyra providing only the Quadlet glue; depends on #727 + #610 landing first.

## Proposed flow

1. **ADR** — document voicecli container strategy (answers to Qs 1–5 above), pick scope boundary (Q6)
2. **Dockerfile** — in voiceCLI repo, CUDA base image + `voicecli nats-serve` as entrypoint
[...]

Issue #661, opened 2026-04-12
chore(ci): automate container digest bumps via Renovate/Dependabot


MickaelV0:
## Context

Follow-up from PR #656 code review. Two manual pain points around container image pinning:

1. **NATS** is pinned by **manifest-list** digest (\`sha256:b83efa…6bb927\`). Per-platform (\`linux/amd64\` — roxabituwer is fixed arch) is a stronger pin. Today the bump is manual: re-resolve via registry API and open a PR.
2. **localhost/lyra** is not pinned at all (#657 tracks the CI tag-by-SHA strategy).

Long-term, this isn't a human decision — it's automation.

## Goal

Upstream digest bumps arrive as reviewable PRs, not as \"oops the tag moved\" surprises.

## Proposed approach

- Add **Renovate** (\`renovate.json\`) or **Dependabot** (\`.github/dependabot.yml\`) with a Podman/Docker ecosystem grouping:
  - Watch \`deploy/quadlet/*.container\` \`Image=\` lines
  - Pin to per-platform digest (\`linux/amd64\`) instead of manifest-list
  - Auto-open a PR with the new digest + changelog link when upstream releases
  - Group minor/patch bumps; route major bumps to a separate PR for human judgement
- Ties in with #657: once CI emits \`localhost/lyra:sha256-…\` tags, the same automation scheme covers lyra itself.

## Scope

1. Pick Renovate vs Dependabot (Renovate has richer container support; Dependabot is native to GH).
2. Configure container ecosystem.
3. Re-resolve the NATS digest to per-platform and update \`nats.container\` in the first Renovate-opened PR (validates end-to-end).
4. Document the bump review workflow in \`docs/\` or CONTRIBUTING.

## Acceptance

- [ ] Renovate or Dependabot config committed
- [ ] A test PR has been opened by the bot for a container digest bump
- [ ] NATS digest in \`nats.container\` is per-platform (amd64), not manifest-list
- [ ] Bump review workflow documented

Issue #657, opened 2026-04-12
chore(ci): tag lyra image by git SHA for digest pinning in Quadlet units


MickaelV0:
## Context

Follow-up from #652 / PR #656. The Quadlet hardening landed digest pinning for the upstream NATS image but deferred `localhost/lyra:latest` because local-build images have no registry digest — and even if we resolve the digest post-build, coupling every rebuild to a manual unit-file commit is the wrong ergonomic.

## Goal

Traceable, reproducible Quadlet deployments for `hub.container` and `lyra-adapter@.container` without requiring every image rebuild to touch the unit files.

## Proposed approach

1. CI workflow on push to `staging` / `main`: build image, tag as `localhost/lyra:${GIT_SHA_SHORT}` (and optionally `:latest` alias).
2. Write the resolved tag/digest into a deploy descriptor (`deploy/quadlet/lyra.image.env` or a systemd drop-in), committed by CI.
3. Quadlet units reference `Image=localhost/lyra:\${LYRA_IMAGE_TAG}` (Environment-driven) or consume the digest directly.
4. `systemctl --user daemon-reload && systemctl --user restart lyra.service` on deploy.

## Acceptance

- [ ] Every deployed hub/adapter container is traceable back to a specific git SHA via `podman inspect`
- [ ] Rolling back = editing one descriptor file + restart, not rebuilding
- [ ] Quadlet units no longer reference `:latest`

## Open questions

- Local-only registry vs pushing to GHCR — trade off between air-gap (roxabituwer is a home server) and CI simplicity
- How to wire tag interpolation into Quadlet `Image=` (Environment, drop-in, generator script)

Issue #611, opened 2026-04-08
chore(infra): supervisord cutover — validate Quadlet stack, remove supervisord, update docs


MickaelV0:
## Parent

Part of #605 (Containerize production — Podman + Quadlet)

## Blocked by

- #610 — Makefile targets must work first

## Context

Final cutover: run Quadlet stack in parallel with supervisord, validate for 48h, then remove supervisord. Update all docs that reference supervisord.

## Cutover procedure

```
1. Deploy Quadlet stack on Machine 1 (different ports during parallel run)
2. Run both stacks simultaneously for 48h — compare logs, error rates
3. Switch NATS_URL + bot tokens to point to Quadlet containers
4. Confirm Lyra is fully operational via Telegram/Discord smoke test
5. Stop supervisord: systemctl --user stop lyra.service
6. Disable lyra.service: systemctl --user disable lyra.service
7. Remove supervisor/ directory (or archive to deploy/legacy/)
```

## Acceptance Criteria

- [ ] Quadlet stack runs stable for 48h on Machine 1 alongside supervisord
- [ ] All adapters respond correctly in Telegram + Discord during parallel run
- [ ] `systemctl --user disable --now lyra.service` — supervisord stopped
- [ ] `supervisor/` directory archived or removed
- [ ] `GETTING-STARTED.md` — replace supervisord setup steps with `make quadlet-install`
- [ ] `DEPLOYMENT.md` — replace all supervisord references with Quadlet equivalents
- [ ] `CLAUDE.md` — update "Production entry points" table to reference Quadlet units
- [ ] `docs/ARCHITECTURE.md` — update deployment section
- [ ] `local/machines.md` — update Machine 1 service list

## Key Files

- `GETTING-STARTED.md` — update
- `DEPLOYMENT.md` — update
- `CLAUDE.md` — update production entry points
- `docs/ARCHITECTURE.md` — update deployment section
- `supervisor/` — archive or remove

MickaelV0:
Cross-reference: #701 (host migration to new `roxabi-ops` repo, part of epic #693 / ADR-043) is blocked by this issue. Rationale: if #611 removes supervisord entirely via Quadlet, #701 shrinks to installing the autodeploy timer + moving `~/projects/Makefile` — no supervisord migration needed. If #701 lands first, supervisord is migrated only to be deleted here → wasted work. So #611 should go first.

PR #812, opened 2026-04-20
chore(infra): bootstrap podman + user socket on Machine 1

### LINKED ISSUE #727: chore(infra): bootstrap podman + user socket on Machine 1 (prereq #610)
## Parent

Part of #605 (Containerize production — Podman + Quadlet)

## Context

Lane B has produced Quadlet unit files (#606 ✅ #607 ✅ #608 ✅ #612 ✅ #652 ✅) but **podman is not installed on Machine 1**. #610 (Makefile Podman targets + image pipeline) assumes `podman build` / `podman save` work. Nothing in the chain bootstraps podman itself — this was an implicit dependency.

## Current state (verified on `roxabituwer`)

- `command -v podman` → not found
- `~/.config/containers/systemd/` → missing
- `loginctl show-user mickael` → `Linger=yes` ✅ (prereq for user systemd services — already set)
- NATS currently runs bare-metal (`/usr/local/bin/nats-server -c /etc/nats/nats.conf`, system unit)

## Acceptance Criteria

- [ ] `podman` installed on Machine 1 (apt package or static binary — pick based on Ubuntu 24.04 version currency)
- [ ] `podman info` succeeds as user `mickael` (no root)
- [ ] `podman run --rm docker.io/library/hello-world` succeeds
- [ ] `~/.config/containers/systemd/` exists (will be populated by `make quadlet-install` from #610)
- [ ] `systemctl --user daemon-reload` picks up future Quadlet generators cleanly
- [ ] Optional: `podman.socket` user unit enabled (for Podman API if any plugin needs it)
- [ ] Machine 1 notes in `local/machines.md` updated with podman version + install method

## Blocks

- #610 — Makefile Podman targets cannot be validated without podman
- transitively: #611 (supervisord cutover), and any voicecli container work (tracked separately)

## Notes

This could have been an implicit step inside #610. Splitting it out makes the dep-graph honest and lets the install be verified independently of the Makefile changes.

MickaelV0:
## Summary
- Add idempotent Podman section to `deploy/provision.sh` — apt install (`podman` + `uidmap` + `fuse-overlayfs` + `slirp4netns`), user-scope systemd socket (`podman.socket`), Quadlet config dir (`~/.config/containers/systemd/`), and a guarded `hello-world` smoke test.
- Unblocks #610 (Makefile Podman targets + image pipeline) and everything downstream in Lane B (#611 cutover, #728 voicecli containers, #657 digest tagging, #661 Renovate).

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #727: bootstrap podman + user socket on Machine 1 | Open |
| Implementation | 1 commit on `worktree-727-bootstrap-podman` | Complete |
| Verification | Bash syntax ✅ · Ruff ✅ · Installed + smoke-tested on `roxabituwer` (Machine 1, Ubuntu 24.04) | Passed |

## Test Plan
- [ ] Re-running `provision.sh` on Machine 1 prints `podman already installed` (idempotency guard).
- [ ] `sudo -u mickael XDG_RUNTIME_DIR=/run/user/1000 systemctl --user is-enabled podman.socket` → `enabled`.
- [ ] `ls ~/.config/containers/systemd/` exists.
- [ ] `podman run --rm docker.io/library/hello-world` prints the greeting.

Closes #727

---

### DIFF
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -27,10 +27,11 @@ AGENT_USER="${AGENT_USER:-lyra}"
 USER_RE='^[a-z_][a-z0-9_-]{0,31}$'
 [[ "$ADMIN_USER" =~ $USER_RE ]] || error "Invalid ADMIN_USER: $ADMIN_USER"
 [[ "$AGENT_USER" =~ $USER_RE ]] || error "Invalid AGENT_USER: $AGENT_USER"
-# Resolve admin home from passwd — do not assume /home/$ADMIN_USER.
+# Resolve admin home + UID from passwd — do not assume /home/$ADMIN_USER.
 ADMIN_HOME=$(getent passwd "$ADMIN_USER" | cut -d: -f6)
 [[ -n "$ADMIN_HOME" && -d "$ADMIN_HOME" ]] || error "Could not resolve home directory for $ADMIN_USER."
-info "Running setup for admin: $ADMIN_USER (home: $ADMIN_HOME), agent: $AGENT_USER"
+ADMIN_UID=$(id -u "$ADMIN_USER")
+info "Running setup for admin: $ADMIN_USER (uid: $ADMIN_UID, home: $ADMIN_HOME), agent: $AGENT_USER"
 
 # ── System packages ──────────────────────────────────────────────────────────
 
@@ -79,15 +80,21 @@ else
 fi
 
 # Verify rootless uid/gid ranges — required for user namespace mapping.
-# Missing when user was created via `useradd` without -U/-m (adduser adds them).
-if ! grep -q "^$ADMIN_USER:" /etc/subuid || ! grep -q "^$ADMIN_USER:" /etc/subgid; then
-  warn "subuid/subgid ranges missing for $ADMIN_USER — adding 100000-165535."
+# Missing or too-small ranges (podman requires ≥65536 IDs) when user was created
+# via `useradd` without -U/-m, or via a manual `usermod --add-subuids` with a
+# smaller range. The grep-only "presence" check would accept `user:100000:100`.
+has_sufficient_subids() {
+  awk -F: -v u="$ADMIN_USER" '$1 == u && $3 >= 65536 {found=1} END {exit !found}' "$1"
+}
+if ! has_sufficient_subids /etc/subuid || ! has_sufficient_subids /etc/subgid; then
+  warn "subuid/subgid ranges missing or too small for $ADMIN_USER — adding 100000-165535."
   sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 "$ADMIN_USER"
   # Re-run migrate in case podman already has stale rootless state.
-  sudo -u "$ADMIN_USER" podman system migrate 2>/dev/null || true
-  info "subuid/subgid added for $ADMIN_USER."
+  sudo -u "$ADMIN_USER" HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
+    podman system migrate 2>/dev/null || true
+  info "subuid/subgid added for $ADMIN_USER (≥65536 IDs)."
 else
-  info "subuid/subgid already configured for $ADMIN_USER."
+  info "subuid/subgid already configured for $ADMIN_USER (≥65536 IDs)."
 fi
 
 # Ensure user-scope container + systemd config dirs exist.
@@ -95,14 +102,25 @@ sudo -u "$ADMIN_USER" mkdir -p "$ADMIN_HOME/.config/containers/systemd"
 sudo -u "$ADMIN_USER" mkdir -p "$ADMIN_HOME/.config/systemd/user"
 info "Container dirs present: ~/.config/containers/systemd/ and ~/.config/systemd/user/"
 
-# Enable linger now so /run/user/$UID persists for systemctl --user calls below
-# (also re-enabled at the Users section — both idempotent).
-loginctl enable-linger "$ADMIN_USER" 2>/dev/null || true
+# Enable linger so /run/user/$UID persists across logout for systemctl --user
+# calls below. Hard-fail if this doesn't work — every downstream systemctl --user
+# call depends on it; silent failure would only surface as cryptic "Failed to
+# connect to bus" errors later.
+loginctl enable-linger "$ADMIN_USER"
+info "Linger enabled for $ADMIN_USER."
+
+# Poll for /run/user/$ADMIN_UID — systemd-logind may create it asynchronously
+# after enable-linger, and on a fresh headless boot it may not exist yet.
+# Bounded wait (≤10s); bail out loud if it never materialises.
+for _ in $(seq 10); do
+  [[ -d "/run/user/$ADMIN_UID" ]] && break
+  sleep 1
+done
+[[ -d "/run/user/$ADMIN_UID" ]] || error "/run/user/$ADMIN_UID never appeared — logind/linger not functional."
 
 # Enable + start the rootless Podman API socket.
 # XDG_RUNTIME_DIR is required when running systemctl --user via sudo;
 # without it systemd cannot locate the user's dbus session.
-ADMIN_UID=$(id -u "$ADMIN_USER")
 if sudo -u "$ADMIN_USER" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
      systemctl --user is-enabled podman.socket &>/dev/null; then
   info "podman.socket already enabled for $ADMIN_USER."
@@ -126,13 +144,13 @@ else
 fi
 
 if sudo -u "$ADMIN_USER" podman images --format '{{.Repository}}' \
-     | grep -q "^docker.io/library/hello-world$"; then
+     | grep -q "^quay.io/podman/hello$"; then
   info "hello-world image already pulled, skipping smoke test."
 else
-  if sudo -u "$ADMIN_USER" podman run --rm docker.io/library/hello-world > /dev/null; then
+  if sudo -u "$ADMIN_USER" podman run --rm quay.io/podman/hello > /dev/null; then
     info "podman hello-world smoke OK"
   else
-    warn "hello-world pull failed — check network / docker.io reachability."
+    warn "hello-world pull failed — check network / quay.io reachability."
   fi
 fi
 
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -22,7 +22,15 @@ section() { echo -e "\n${GREEN}=== $1 ===${NC}"; }
 ADMIN_USER="${ADMIN_USER:-$(whoami)}"
 # Agent user (defaults to lyra, override with AGENT_USER=anotherame)
 AGENT_USER="${AGENT_USER:-lyra}"
-info "Running setup for admin: $ADMIN_USER, agent: $AGENT_USER"
+# Validate usernames — reject shell metachars since values are env-driven (curl|bash).
+# Matches POSIX NAME_REGEX used by useradd: [a-z_][a-z0-9_-]* (max 32 chars).
+USER_RE='^[a-z_][a-z0-9_-]{0,31}$'
+[[ "$ADMIN_USER" =~ $USER_RE ]] || error "Invalid ADMIN_USER: $ADMIN_USER"
+[[ "$AGENT_USER" =~ $USER_RE ]] || error "Invalid AGENT_USER: $AGENT_USER"
+# Resolve admin home from passwd — do not assume /home/$ADMIN_USER.
+ADMIN_HOME=$(getent passwd "$ADMIN_USER" | cut -d: -f6)
+[[ -n "$ADMIN_HOME" && -d "$ADMIN_HOME" ]] || error "Could not resolve home directory for $ADMIN_USER."
+info "Running setup for admin: $ADMIN_USER (home: $ADMIN_HOME), agent: $AGENT_USER"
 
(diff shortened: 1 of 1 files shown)


MickaelV0:
## Code Review

Multi-domain review by `security-auditor` + `devops`. Verdict: **Approve with comments** — no blockers, 4 real robustness issues worth addressing before the script is re-run on a fresh machine.

### Warnings

**`podman run hello-world` aborts script on network failure** — `deploy/provision.sh:104-105` (devops, 92%)
Under `set -euo pipefail`, `podman run ... && info` will exit non-zero if docker.io is unreachable (no DNS on a fresh machine, rate limit, registry down). The comment claims the smoke test is "guarded" but the guard only skips when the image is already pulled. On first run it pulls unconditionally.
- **Fix:** `podman run ... > /dev/null && info "smoke OK" || warn "hello-world pull failed — check network"`

**`podman info` silent abort under `set -e`** — `deploy/provision.sh:97-98` (security-auditor + devops, 90%)
Same pattern: `podman info ... && info "OK"`. If `podman info` fails (e.g. storage not initialized, subuid missing) the script aborts with no message before SSH hardening / ufw run, leaving the machine half-provisioned.
- **Fix:** explicit `if/else` with `warn` in the failure branch, or append `|| warn "..."` to the compound statement.

**`systemctl --user` depends on linger which is enabled later** — `deploy/provision.sh:82-93` (devops, 88%)
`loginctl enable-linger "$ADMIN_USER"` is at line 249, after Security/GRUB/Users sections. On a fresh machine without linger, `/run/user/$ADMIN_UID` only exists during an active login session. Running `systemctl --user enable --now podman.socket` works during that session but the socket stops when admin logs out.
- **Fix:** move `loginctl enable-linger "$ADMIN_USER"` to immediately before the Podman `systemctl --user` calls, or rely on the fact that the Admin runs the script in an ssh session (document this).

**Hardcoded `/home/$ADMIN_USER` path** — `deploy/provision.sh:74-75` (devops, 85%)
Breaks for LDAP users, admins created with `useradd -d /var/lib/$USER`, or any non-standard homedir. Creates config in the wrong place while the user's actual `~/.config/` remains absent.
- **Fix:** `sudo -u "$ADMIN_USER" bash -c 'mkdir -p ~/.config/containers/systemd ~/.config/systemd/user'` — lets the target user's shell expand `~`. Or `ADMIN_HOME=$(getent passwd "$ADMIN_USER" | cut -d: -f6)`.

- **`subuid`/`subgid` verification** (devops, 78%) — rootless Podman requires entries in `/etc/subuid`/`/etc/subgid`. Ubuntu auto-creates them for `useradd` accounts, but LDAP/SSSD users won't have them. `podman info` will fail silently without a clear error.
- **`local/machines.md` automation** (devops, 97%) — AC item 7 isn't automated. Add a `warn` at section end: `warn "Update local/machines.md: podman $(podman --version | awk '{print $3}')"`.
- **`$ADMIN_USER` input validation** (security-auditor, 88%) — defense in depth. `sudo` already rejects invalid usernames, but an explicit regex guard (`[[ "$ADMIN_USER" =~ ^[a-z_][a-z0-9_-]{0,31}$ ]] || error ...`) would fail fast with a clear message and protect the pre-existing `$AGENT_USER` path construction at line 173 too.
- **Version comment precision** (devops, 83%) — "podman 4.9.x" → "podman 4.9.3+ds1" (apt-cache policy podman).

### Praise

- **XDG_RUNTIME_DIR comment** — the two-line explanation for why the env var is needed when crossing user context via `sudo -u` is exactly the kind of documentation that saves a future engineer a 30-minute debugging session. (100%)
- **Idempotency pattern consistency** — `command -v` / `is-enabled` / `podman images | grep` guards match the existing script conventions (nvidia-smi, moviepy venv, ufw status). (100%)

### Summary

The changes work on the machine where they were verified (Machine 1, user `mickael`, linger already enabled, standard `/home/mickael` path). The findings are about **resilience on a fresh machine or non-standard admin account** — which is the whole point of a provisioning script. Shipping as-is is fine given #727's AC is met; the robustness fixes can land in a follow-up.

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 3 finding(s)
**Skipped:** 0
**Failed:** 0

### Auto-Applied
- [applied] suggestion: Enable linger before systemctl --user (ordering fix)
- [applied] issue: Guard `podman info` with if/else soft-fail + warn
- [applied] issue: Guard `podman run hello-world` with if/else soft-fail + warn
- [applied] nitpick: Add machines.md reminder after podman install

### Applied (1b1)
- [applied] suggestion: Resolve `ADMIN_HOME` via `getent passwd`; stop assuming `/home/$ADMIN_USER`
- [applied] issue: Validate `ADMIN_USER` / `AGENT_USER` against NAME_REGEX — blocks shell injection via `curl|bash` env vars
- [applied] suggestion: Verify subuid/subgid ranges; auto-add `100000-165535` via `usermod` if missing (with `podman system migrate` fallback)

### Also updated
- Comment `Ubuntu 24.04 / podman 4.9.x` → `Ubuntu 26.04 LTS / podman 5.x` (prod host upgraded; apt ships 5.x natively, no pinning needed).

MickaelV0:
## Code Review (iteration 2, post-fix)

Fresh pass by `security-auditor` + `devops` on commit `1bb3c61`. Re-review after 7 fixes applied in iteration 1. Verdict below.

---

### Blockers (6)

**issue(blocking): `loginctl enable-linger || true` silently swallows fatal error** — `deploy/provision.sh:100`
*-- devops, 92%*
`|| true` masks linger failure (PAM denial, D-Bus down, policy). All downstream `systemctl --user` calls then hit "Failed to connect to bus" with no clear cause.
Fix: drop `|| true`; let it hard-fail with `error`.

**issue(blocking): Duplicate `loginctl enable-linger` at :100 and :282 — both silent-fail** — `deploy/provision.sh:100,282`
*-- devops, 90%*
Two silent calls create false confidence. The only user-visible `info` log is the second one, but it's on the duplicate.
Fix: consolidate to one hard-fail call at :100; remove :282.

**issue(blocking): `/run/user/$ADMIN_UID` may not exist immediately after linger** — `deploy/provision.sh:106-117`
*-- devops + security, 88%*
`loginctl enable-linger` writes `/var/lib/systemd/linger/$USER` but systemd-logind creates `/run/user/$UID` asynchronously. On a headless first-run with no active session, `systemctl --user enable --now podman.socket` fails with "Failed to connect to bus".
Fix: poll for `/run/user/$UID` existence (≤10s) after enable-linger, or use `machinectl shell "$USER"@` to force a proper logind session.

**issue(blocking): `podman images` fast-path may fail before rootless stack is operational** — `deploy/provision.sh:128-129`
*-- devops, 80%*
If subuid was just added + migrate ran, the kernel hasn't re-mapped the uid range for `sudo -u` processes without a new login. `podman images` errors confusingly instead of falling through.
Fix: wrap the `podman images` check in the same soft-fail pattern as `podman run`.

**issue(blocking): subuid range `100000-165535` hardcoded, ignores existing entries** — `deploy/provision.sh:85`
*-- security, 82%*
If another user already holds that band (or real UIDs fall in it — LDAP/container hosts can place UIDs ≥100k), the mapping silently aliases another account's UID inside the namespace. Rootless → privilege escalation path.
Fix: compute next free range: `awk -F: '{print $2+$3}' /etc/subuid | sort -n | tail -1`, start at `max(65536, hwm+1)`, allocate 65536.

**suggestion(blocking): subuid grep checks presence, not count** — `deploy/provision.sh:83`
*-- devops, 85%*
A manual entry like `user:100000:100` passes the grep but gives only 100 IDs (podman needs ≥65536). Runtime fails with cryptic "invalid mapping".
Fix: `awk -F: '$1 == ADMIN_USER && $3 >= 65536' /etc/subuid`; only skip usermod if count is sufficient.

**suggestion(blocking): `hello-world` pulled by tag from docker.io — supply-chain exposure during provisioning** — `deploy/provision.sh:132`
*-- security, 78%*
Mutable tag → registry hijack / compromised mirror → arbitrary code under ADMIN_USER namespace before SSH hardening runs.
Fix: either pin by digest or (cleaner) drop the network pull entirely — `podman info --format '{{.Host.CgroupsVersion}}'` is a sufficient smoke test with zero network.

---

### Warnings (3)

**suggestion(non-blocking): `podman system migrate` can silently remap existing rootless storage** — `deploy/provision.sh:87`
*-- security, 75%*
`2>/dev/null || true` swallows migration errors. If the user has prior rootless containers (partial re-run), they may be orphaned silently.
Fix: check `podman images -q` first; if non-empty, require explicit `FORCE_MIGRATE=1` flag.

**suggestion(non-blocking): `podman system migrate` missing `HOME` + `XDG_RUNTIME_DIR`** — `deploy/provision.sh:87`
*-- devops, 75%*
`sudo -u` without env vars lets podman resolve from root's environment. On fresh user this is a no-op, but worth fixing for correctness.
Fix: add `HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID"` to the migrate invocation.

**suggestion(non-blocking): Replace `docker.io/library/hello-world` with `quay.io/podman/hello`** — `deploy/provision.sh:132`
*-- devops, 88%*
Docker Hub unauth rate-limits (100 pulls/6h/IP). `quay.io/podman/hello` is the official podman smoke image — no rate limit, maintained by Red Hat.
Fix: swap the image string in both the `podman images` check and the `podman run`.

---

### Nitpicks (3)

**nitpick: `LC_ALL=C` not set — locale-dependent `[a-z]` regex** — `deploy/provision.sh:27`
*-- security, 68%*
UTF-8 locales can collate accented letters into `[a-z]`. Low practical risk, but `LC_ALL=C` at top of script pins it cleanly.

**nitpick: `uidmap` / `fuse-overlayfs` / `slirp4netns` may be redundant on podman 5.x** — `deploy/provision.sh:77`
*-- devops, 70%*
`uidmap` is a hard dep of the podman package since 4.5+. `slirp4netns` deprecated in favor of `pasta` in 5.x. `fuse-overlayfs` only needed when kernel overlayfs is unavailable (it is, on 26.04). Not harmful, just noise.

**nitpick: Escape ADMIN_USER in grep pattern (belt-and-braces)** — `deploy/provision.sh:83`
*-- security, 72%*
`USER_RE` already blocks metachars, but `grep -qF "^${ADMIN_USER}:"` (fixed-string) removes the whole class of concern. Cheap fix.

---

### Praise (1)

**praise: `ADMIN_HOME` extraction handles GECOS colons correctly** — `deploy/provision.sh:31`
*-- devops, 97%*
`cut -d: -f6` reads column 6; GECOS is column 5, so colons inside GECOS don't affect the result. Clean fix from iteration 1.

---

### Summary

| | Count |
|---|---|
| Blockers | **6** |
| Warnings | 3 |
| Nitpicks | 3 |
| Praise | 1 |
| **Verdict** | **Request changes** |

Main thread: rootless-podman ordering + first-run assumptions are fragile. `linger → systemctl --user` needs a real logind session (or explicit `/run/user/$UID` setup); subuid range logic needs to check availability and sufficiency, not just presence; smoke test should avoid unpinned network pulls during provisioning.

MickaelV0:

**Auto-applied (Q_auto, C≥80 + 2 agents):** 3 finding(s) — commit `85269bd`
**Applied via 1b1 (3-man consensus: architect + devops + security-auditor):** 3 finding(s) — commit `8ea1a7c`
**Deferred (issues filed):** 2 finding(s)
**Skipped:** 1 finding
**Failed:** 0

### Auto-Applied
- [applied] issue(blocking, 88%): `/run/user/$ADMIN_UID` race post-linger — hard-fail linger + poll loop up to 10s (F3)
- [applied] suggestion(blocking, 82%): subuid range presence-only check → awk-based count check (≥65536 IDs) (F5)
- [applied] suggestion(blocking, 88%): `docker.io/library/hello-world` → `quay.io/podman/hello` (no rate limit, maintained by Red Hat) (F7)

*Cascade effects of auto-apply:*
- F1 (`loginctl || true` swallow) effectively fixed at line 109 via F3's hard-fail refactor
- F10 (`podman migrate` missing `HOME`/`XDG_RUNTIME_DIR`) effectively fixed via F5's cascade
- F13 (grep metachar via `ADMIN_USER`) moot — grep replaced by awk in F5

### Applied (1b1, consensus 3/3 or 2/3)
- [applied] issue(blocking, 72%): remove duplicate silent-fail `loginctl enable-linger` at end of Users section (F2)
- [applied] nitpick (68%): `export LC_ALL=C` at top — pin locale so `[a-z]` regex is ASCII-only (F11)
- [applied] nitpick (70%): comment update on `uidmap fuse-overlayfs slirp4netns` dep list — reflect podman 5.x reality (F12)

### Deferred (follow-up issues)
- suggestion (68%): subuid range `100000-165535` hardcoded → compute next-free range — **#825** (F6)
- suggestion (75%): `podman system migrate` can silently remap rootless storage → `FORCE_MIGRATE=1` guard — **#826** (F8)

### Skipped (agent consensus)
- issue(blocking→55%): `podman images` fast-path — verifier C=55, else-branch already soft-fails (F4)

### Consensus methodology
Spawned 3 fresh agents (architect, devops, security-auditor) with identical payload for 6 Q_1b1 items. Each returned APPLY/DEFER/SKIP per finding with ≤40-word long-term rationale. Final actions matched majority vote (3/3 on F2/F4/F6/F11; split-resolved for F8 → defer, F12 → apply comment-only).

### Next
CI re-running on `8ea1a7c`. Loop cap hit — third review cycle not allowed; merge-as-is decision next.

PR #869, opened 2026-04-22
chore(provision): compute next-free subuid/subgid range

### LINKED ISSUE #825: chore(provision): compute next-free subuid/subgid range instead of hardcoding 100000-165535
## Context

Split out from PR #812 (iteration 2 review, finding F6).

`deploy/provision.sh:86` currently hardcodes:

```bash
sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 "$ADMIN_USER"
```

`usermod --add-subuids` performs no overlap check (man page is explicit). If the script is ever run on a host where another user already holds that band — or where real UIDs fall in `100000+` (LDAP, container hosts, corporate systems) — the rootless namespace silently aliases another account's UID, which is a privilege-escalation path from inside any rootless container.

## Why deferred from #727

On `roxabituwer` (single-admin personal server, no LDAP, `lyra` agent user has no subuid entries) the collision probability is effectively zero. 3-man agent consensus (architect + devops + security-auditor) all recommended DEFER rather than gold-plate the S-tier scope of #727.

Compute next-free range from current high-water mark:

\`\`\`bash
hwm=\$(awk -F: '{print \$2+\$3}' /etc/subuid 2>/dev/null | sort -n | tail -1)
start=\$(( \${hwm:-0} > 65535 ? hwm + 1 : 65536 ))
sudo usermod --add-subuids "\${start}-\$((start+65535))" \\
              --add-subgids "\${start}-\$((start+65535))" "\$ADMIN_USER"
\`\`\`

Revisit when:
- Machine 1 gets a second human user with rootless containers.
- provision.sh starts being used on more than one host class.
- Lyra's agent pool grows such that lyra (AGENT_USER) needs rootless too.

## Links

- PR #812 (closes parent #727)
- Epic #605 (Lane B — containerize)

MickaelV0:
## Summary
- `usermod --add-subuids` performs no overlap check — hardcoded `100000-165535` silently aliases another account's UID on hosts with pre-existing entries (LDAP, multi-user container hosts), opening a privilege-escalation path from inside any rootless container.
- Derive `start` from the high-water mark across `/etc/subuid` and `/etc/subgid`, floor at 65536 to stay clear of the normal UID space.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #825: chore(provision): compute next-free subuid/subgid range | Open |
| Implementation | 1 commit on \`chore/825-compute-next-free-subuid\` | Complete |
| Verification | Lint ✅ Typecheck ✅ bash -n ✅ | Passed |

## Test Plan
- [ ] Dry-run `next_free_subid_start` on a host with multiple `/etc/subuid` entries — asserts start > last allocation end.
- [ ] Empty `/etc/subuid`/`/etc/subgid` → falls through to 65536 floor.
- [ ] `usermod --add-subuids "${start}-${end}"` runs against \$ADMIN_USER without collision warnings.

Closes #825

---

### DIFF
--- a/artifacts/analyses/825-subuid-review-followups-consensus.mdx
+++ b/artifacts/analyses/825-subuid-review-followups-consensus.mdx
@@ -1,0 +1,74 @@
+---
+title: "subuid/subgid review follow-ups — Expert Consensus"
+issue: 825
+status: consensus-reached
+date: 2026-04-22
+panel: architect, security-auditor, devops
+confidence: high
+---
+
+## Problem
+
+PR #869 (issue #825) replaces a hardcoded `100000-165535` subuid/subgid range in `deploy/provision.sh` with an HWM-computed next-free range. Two blockers from the initial review were auto-applied in PR #869: per-namespace HWM, `usermod` exit check, 32-bit UID space overflow guard.
+
+Four non-blocking findings remained; panel was convened to recommend apply / defer / skip per finding on this security-hardening PR whose original scope was explicitly deferred to a single-admin machine (`roxabituwer`).
+
+## Panel
+
+| α | Focus |
+|---|-------|
+| architect | arch soundness, maintainability, long-term invariants |
+| security-auditor | attack surface, threat model for the deployment context |
+| devops | ops payback for a run-once-per-install provisioning script |
+
+## Findings and votes
+
+| # | Finding | Architect | Security | Devops | Consensus |
+|---|---------|-----------|----------|--------|-----------|
+| 1 | Non-contiguous overlap detection (C=72%) | apply | defer | skip | **Defer → #875** |
+| 2 | TOCTOU between HWM read and usermod write (C=70%) | skip | skip | skip | **Skip (wontfix, documented inline)** |
+| 3 | `awk ... 2>/dev/null` swallows perm errors (C=78%) | apply | apply | defer | **Apply in PR #869** |
+| 4 | Re-run idempotency warn (C=75%) | defer | defer | defer | **Defer → #876** |
+
+## Consensus
+
+### F1 — Non-contiguous overlap: Defer (→ #875)
+
+**Rationale:** HWM protects against the single highest-ending entry. Out-of-order / manually-edited entries could hide an overlap. On a single-admin host where `/etc/subuid` is managed exclusively by `usermod`, the malformed-file vector requires manual misconfiguration that doesn't exist today. Architect's argument (closes HWM invariant gap) is valid long-term but security + devops both cite no current payback. Ship follow-up issue to capture the improvement.
+
+### F2 — TOCTOU: Skip (wontfix)
+
+**Rationale:** Unanimous. Reviewer's own admission: shadow-utils does not honor external `flock`. Any shell-level fix is security theater that gives a false safety signal. On a single-admin machine with one provisioner, no concurrent `usermod` exists. Better to record a wontfix comment in code than to file a ticket that invites a future implementer to add an ineffective flock.
+
+### F3 — awk error swallow: Apply now
+
+**Rationale:** 2-of-3 apply (devops agrees on substance, disagrees on timing). This is the only finding with a concrete silent-failure path on the target machine: an unreadable `/etc/subuid` (chmod drift, SELinux denial) currently falls through to the 65536 floor silently and could alias uid 0. Fix is a 3-line preflight: `[[ -e && ! -r ]] → error`. Consistency of intent — the whole PR exists to prevent silent wrong-range assignment.
+
+### F4 — Re-run idempotency warn: Defer (→ #876)
+
+**Rationale:** Unanimous. `has_sufficient_subids` is intentionally permissive to preserve idempotency. A badly-placed prior entry still satisfying count is an edge that surfaces via the `podman info` smoke test already present downstream in `provision.sh`. Warn-only overlap scan is legitimate but adds scope; revisit when `AGENT_USER` subuid provisioning is added.
+
+## Alternatives considered
+
+| ω | Proposed by | Rejected because |
+|---|-------------|------------------|
+| Apply all 4 now | architect (F1) | Over-engineers for scenarios the issue body explicitly deferred to "another host class" / multi-user. |
+| TOCTOU flock fix | (none — security flagged it, but panel rejected) | Shadow-utils ignores external flock; false safety signal. |
+| Hard-fail idempotency on overlap | (rejected) | Breaks idempotent re-provision for no benefit on single-admin. Warn is the correct signal level, deferred. |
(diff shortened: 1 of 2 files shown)


MickaelV0:
## Code Review

**Verdict:** Approve with comments (2 blocking, 5 non-blocking).

The HWM approach closes the hardcoded-range privilege-escalation risk for the typical case. Reviewers flagged two issues that partially reopen or mask it: unchecked `usermod` exit + cross-namespace HWM. Recommend addressing both before merge; the rest are worth capturing as follow-ups.

### Blocking

**issue(blocking): `usermod` exit code not checked — provisioning silently continues on allocation failure**
- `deploy/provision.sh:115`
- -- security-auditor
- Root cause: No `|| { error; exit 1; }` guard on the `usermod` call. If the range is rejected (32-bit overflow, malformed entry, perm error), the script prints the success `info` on line 119 and continues. The rootless env is misconfigured; the PE risk is silently deferred, not closed.
- Solutions:
  1. (recommended) `sudo usermod ... || { error "Failed to configure subuid/subgid for $ADMIN_USER"; exit 1; }`
  2. Re-run `has_sufficient_subids` after `usermod` and abort if still false.
- Confidence: 92%

**issue(blocking): HWM computed across both /etc/subuid and /etc/subgid — cross-namespace pollution**
- `deploy/provision.sh:104`
- -- security-auditor / devops (both flagged independently, C=85/82)
- Root cause: `awk -F: '{print $2 + $3}' /etc/subuid /etc/subgid` merges both namespaces into a single stream. UID and GID subordinate namespaces are independent. Result: if subgid extends further than subuid (or vice versa), the new allocation is pushed unnecessarily high and both `--add-subuids`/`--add-subgids` are co-allocated at the same range by accident — safe, but brittle and UID-space-wasteful.
- Solutions:
  1. (recommended) Two HWM passes, one per file. Pass potentially different ranges to `--add-subuids` and `--add-subgids`.
  2. Keep the combined approach but add a comment documenting the intentional co-allocation invariant and add an upper-bound guard.
- Confidence: 85%

### Non-blocking

**issue(non-blocking): no upper-bound guard before usermod — 32-bit UID space overflow**
- `deploy/provision.sh:113` · -- security-auditor · C=88%
- Add `(( end > 4294967295 )) && { error "subuid space exhausted"; exit 1; }` before the `usermod` call.

**issue(non-blocking): HWM-only does not detect overlap with non-contiguous existing ranges**
- `deploy/provision.sh:104` · -- security-auditor · C=72%
- Add an awk verification pass confirming `[start, start+65536)` does not intersect any existing interval.

**issue(non-blocking): TOCTOU — HWM read and usermod write are not atomic**
- `deploy/provision.sh:112-115` · -- security-auditor · C=70%
- Wrap read+write in `flock /etc/subuid /etc/subgid` (reduces but does not eliminate window — shadow-utils uses its own locking).

**issue(non-blocking): `awk ... 2>/dev/null` swallows legitimate errors (perm denied, corrupt file)**
- `deploy/provision.sh:104` · -- devops · C=78%
- Prefer `sudo awk ...` without stderr redirect, or add `[[ -r /etc/subuid ]]` preflight to distinguish "empty" from "unreadable".

**issue(non-blocking): re-run idempotency — `has_sufficient_subids` accepts a pre-existing badly-placed range**
- `deploy/provision.sh:111-121` · -- devops · C=75%
- Gate checks count (`>= 65536`) not placement. Prior mis-provisioning won't be detected. Add optional overlap-warn scan.

### Praise

**praise: `has_sufficient_subids` awk combines presence + size check correctly**
- `deploy/provision.sh:92-94, 111` · -- devops · C=95%
- Avoids the grep-presence-only false-positive called out in the pre-existing comment.

---
*Reviewed by `dev-core:security-auditor` + `dev-core:devops`. 1 file, 20+/2-.*

MickaelV0:
## Review Fixes Applied

Panel consensus (architect + security-auditor + devops) — see [artifact](../blob/chore/825-compute-next-free-subuid/artifacts/analyses/825-subuid-review-followups-consensus.mdx).

**Auto-applied (C≥80, 2+ agents):** 3 findings in commit `3d7bc28`
**Applied via 1b1 + consensus:** 1 finding in commit `76d572c`
**Deferred (follow-up issues):** 2
**Skipped (wontfix, inline note):** 1
**Praise:** 1

### Auto-Applied (`3d7bc28`)
- [applied] issue(blocking): `usermod` exit code not checked — now aborts via `error` helper.
- [applied] issue(blocking): HWM computed across both files — now per-namespace.
- [applied] issue(non-blocking): 32-bit UID space overflow — now guarded before `usermod`.

### Applied via 1b1 + Consensus (`76d572c`)
- [applied] issue(non-blocking): `awk ... 2>/dev/null` swallow — preflight `[[ -e && ! -r ]] → error` added. Silent fall-through to floor 65536 was a real bug; panel 2-of-3 apply (devops agreed on substance).

### Skipped (wontfix)
- issue(non-blocking): TOCTOU — shadow-utils does not honor external `flock`; any mitigation is security theater on a single-admin host. Documented inline as a `NOTE (TOCTOU wontfix)` with revisit condition.

### Deferred (follow-up issues)
- issue(non-blocking): non-contiguous overlap detection → **#875**
- issue(non-blocking): re-run idempotency warn → **#876**

Both scoped to "revisit when provision.sh runs on a second host class or AGENT_USER gets its own rootless block," consistent with issue #825's original deferral rationale.

MickaelV0:
## Code Review (iteration 2)

**Verdict:** Approve with comments (1 low-confidence finding, defense-in-depth only).

All 4 blockers and 3 non-blocking items from iteration 1 are correctly addressed — verified against current diff.

### Non-blocking (defense-in-depth)

**issue(non-blocking): `awk` arithmetic on unvalidated field content**
- `deploy/provision.sh` — `next_free_subid_start` line ~109
- -- security-auditor (fresh agent, iter 2) · C=72%
- Root cause: `awk '{print $2 + $3}' "$file"` does no field validation. A non-numeric or oversized value (garbage line, truncated write, IEEE-754 precision loss above 2^53) could yield a `hwm` that passes `-gt 65535` but breaks `$((hwm + 65535))` or produces a float bash rejects — silently falling back to the 65536 floor.
- Threat model: `/etc/subuid` is root-owned 644; writing garbage requires root, which neutralizes the exploit vector. Realistic trigger is benign corruption (truncated write, manual edit).
- Solutions:
  1. Add a regex guard before arithmetic: `[[ "$hwm" =~ ^[0-9]+$ ]] || hwm=""` (recommended — 1 line).
  2. Harden `awk`: `$2 !~ /^[0-9]+$/ || $3 !~ /^[0-9]+$/ {next}`.
  3. Defer — fold into #875 (both are `/etc/subuid` integrity concerns).

### Loop Cap

Per `/fix` skill: max 2 fix→review iterations. A 3rd fix round is refused. Options at this point: merge + optional inline 1-line guard, or defer to #875.

---
*Re-review by fresh `dev-core:security-auditor`. All prior findings correctly resolved.*

PR #870, opened 2026-04-22
chore(provision): guard podman system migrate against silent storage remap

### LINKED ISSUE #826: chore(provision): guard podman system migrate against silent storage remap
## Context

Split out from PR #812 (iteration 2 review, finding F8).

`deploy/provision.sh:93` currently runs:

\`\`\`bash
sudo -u "\$ADMIN_USER" HOME="\$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/\$ADMIN_UID" \\
  podman system migrate 2>/dev/null || true
\`\`\`

On a clean first-run this is a safe no-op. On a partial re-run (admin user already had prior rootless containers or named volumes), the migrate step can silently remap or orphan existing storage. `2>/dev/null || true` swallows any resulting error, so data loss would go undetected.

## Why deferred from #727

3-man agent consensus was split (architect SKIP, devops DEFER, security-auditor APPLY). Reasoning to defer rather than apply now:

- Scenario (partial re-run with existing rootless state) is not a documented use case — script runs once per freshly provisioned machine.
- A `FORCE_MIGRATE=1` guard introduces a re-run contract the rest of the script doesn't have.
- No orphaned-volume incidents observed yet on Machine 1.

Security-auditor's concern is valid long-term; re-evaluate if rootless state loss is observed or if provision.sh starts being re-run routinely.

\`\`\`bash
if sudo -u "\$ADMIN_USER" HOME="\$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/\$ADMIN_UID" \\
     podman images -q 2>/dev/null | grep -q .; then
  if [[ "\${FORCE_MIGRATE:-0}" != "1" ]]; then
    warn "Rootless storage for \$ADMIN_USER already has images; skipping migrate."
    warn "Set FORCE_MIGRATE=1 to re-run migrate (may orphan existing volumes)."
  else
    sudo -u "\$ADMIN_USER" HOME="\$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/\$ADMIN_UID" \\
      podman system migrate
  fi
else
  sudo -u "\$ADMIN_USER" HOME="\$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/\$ADMIN_UID" \\
    podman system migrate 2>/dev/null || true
fi
\`\`\`

## Links

- PR #812 (closes parent #727)
- Epic #605 (Lane B — containerize)

MickaelV0:
## Summary
- Detect pre-existing rootless images before re-running `podman system migrate` in `deploy/provision.sh`; skip unless `FORCE_MIGRATE=1` is set so partial re-runs cannot silently orphan existing volumes.
- Fresh-machine path unchanged (still tolerant `|| true`).

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #826: chore(provision): guard podman system migrate against silent storage remap | OPEN |
| Implementation | 1 commit on `feat/826-guard-podman-system-migrate` | Complete |
| Verification | Lint ✅ Typecheck ✅ (bash -n) / pre-commit passed | Passed |

## Test Plan
- [ ] Fresh machine (no prior rootless images): migrate runs, provision unchanged.
- [ ] Machine with existing rootless images + `FORCE_MIGRATE` unset: migrate is skipped, warning printed.
- [ ] Machine with existing rootless images + `FORCE_MIGRATE=1`: migrate runs, any failure surfaces (no `|| true`).

Closes #826

---

### DIFF
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -110,8 +110,11 @@ if ! has_sufficient_subids /etc/subuid || ! has_sufficient_subids /etc/subgid; t
         podman system migrate
     fi
   else
+    # No prior rootless state detected — migrate is safe to run loud so real
+    # failures (e.g. broken graph driver, missing newuidmap) surface under
+    # `set -euo pipefail` instead of being silently swallowed.
     sudo -u "$ADMIN_USER" HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
-      podman system migrate 2>/dev/null || true
+      podman system migrate
   fi
   info "subuid/subgid added for $ADMIN_USER (≥65536 IDs)."
 else
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -96,8 +96,23 @@ if ! has_sufficient_subids /etc/subuid || ! has_sufficient_subids /etc/subgid; t
   warn "subuid/subgid ranges missing or too small for $ADMIN_USER — adding 100000-165535."
   sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 "$ADMIN_USER"
   # Re-run migrate in case podman already has stale rootless state.
-  sudo -u "$ADMIN_USER" HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
-    podman system migrate 2>/dev/null || true
+  # Guard against silent storage remap on partial re-runs: if the admin user
+  # already has rootless images, `podman system migrate` can orphan existing
+  # volumes, and `2>/dev/null || true` would swallow the failure. Require an
+  # explicit FORCE_MIGRATE=1 opt-in when prior state is detected.
+  if sudo -u "$ADMIN_USER" HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
+       podman images -q 2>/dev/null | grep -q .; then
+    if [[ "${FORCE_MIGRATE:-0}" != "1" ]]; then
+      warn "Rootless storage for $ADMIN_USER already has images; skipping migrate."
+      warn "Set FORCE_MIGRATE=1 to re-run migrate (may orphan existing volumes)."
+    else
+      sudo -u "$ADMIN_USER" HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
+        podman system migrate
+    fi
+  else
+    sudo -u "$ADMIN_USER" HOME="$ADMIN_HOME" XDG_RUNTIME_DIR="/run/user/$ADMIN_UID" \
+      podman system migrate 2>/dev/null || true
+  fi
   info "subuid/subgid added for $ADMIN_USER (≥65536 IDs)."
 else
   info "subuid/subgid already configured for $ADMIN_USER (≥65536 IDs)."
--- a/deploy/provision.sh
+++ b/deploy/provision.sh
@@ -97,14 +97,18 @@ if ! has_sufficient_subids /etc/subuid || ! has_sufficient_subids /etc/subgid; t
   sudo usermod --add-subuids 100000-165535 --add-subgids 100000-165535 "$ADMIN_USER"
   # Re-run migrate in case podman already has stale rootless state.
   # Guard against silent storage remap on partial re-runs: if the admin user
-  # already has rootless images, `podman system migrate` can orphan existing
-  # volumes, and `2>/dev/null || true` would swallow the failure. Require an
-  # explicit FORCE_MIGRATE=1 opt-in when prior state is detected.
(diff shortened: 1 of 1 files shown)


MickaelV0:
## Code Review

**Scope:** 1 file, `deploy/provision.sh` — guard around `podman system migrate`.
**Agents:** security-auditor, devops (architect skipped: |Δ|=1; product-lead skipped: no spec; tester skipped: infra-only).

### Blockers

**issue:** `FORCE_MIGRATE=1` unset path exits 0 with no signal — operator can't tell migrate was needed but skipped.
  `deploy/provision.sh:105-107`
  -- devops
  Root cause: When images are detected and `FORCE_MIGRATE` is unset, two `warn` lines fire and execution falls through to `info "subuid/subgid added"`. Exit code is 0. In `curl | bash` runs the operator sees a clean exit even though the subuid change was not reconciled against existing rootless state. On the next container operation podman may still hit the stale mapping.
  Solutions:
    1. Exit non-zero (or call `error`) after the two `warn` lines so the operator must make an explicit decision. (recommended)
    2. Add a distinct `SKIP_MIGRATE_OK=1` flag that must also be set to suppress the non-zero exit — intentional skip becomes explicit.
  Confidence: 85%

### Suggestions

**suggestion:** `podman images -q` is too narrow a signal for "has prior rootless state".
  `deploy/provision.sh:103-104`
  -- devops
  Root cause: Rootless state also includes named volumes (`~/.local/share/containers/storage/volumes/`) and stopped containers, neither of which appear in `podman images -q`. A partial re-run where images were pulled then removed (failed service start, cleanup) but volumes persist would read as "no state" → migrate runs unconditionally → volumes orphaned. That is exactly the scenario issue #826 worries about.
  Solutions:
    1. Broaden: `podman images -q | grep -q . || podman ps -aq | grep -q . || podman volume ls -q | grep -q .` — any non-empty result trips the guard. (recommended)
    2. Filesystem-level: `[[ -n "$(ls -A $ADMIN_HOME/.local/share/containers/storage/volumes/ 2>/dev/null)" ]]` — no podman invocation, survives a broken graph driver.
  Confidence: 90%

**nitpick:** Fresh-run `else` branch still carries `2>/dev/null || true` — silent-failure concern from #826 is not fully addressed.
  `deploy/provision.sh:113-114`
  -- devops
  Root cause: The `FORCE_MIGRATE=1` branch correctly drops the `|| true` (fails loud). The `else` branch (no prior state) still swallows errors. A broken graph driver or missing `newuidmap` binary would pass unnoticed even on a fresh machine.
  Solutions:
    1. Drop `2>/dev/null || true` from the `else` branch; let `podman system migrate` fail loud. (recommended)
    2. Capture stderr to a temp file and `warn` on failure — preserves "don't crash" intent while surfacing the message.
  Confidence: 95%

### Security

No security issues found. `ADMIN_USER`/`ADMIN_UID` validated upstream (lines 29-31). `FORCE_MIGRATE` only used as a boolean gate, never interpolated. `podman images -q` output piped to `grep -q .`, not shell-evaluated.

### Summary

- Blockers: 1
- Warnings: 0
- Suggestions: 2 (1 suggestion, 1 nitpick)
- Praise: 0

**Verdict:** Request changes — address the exit-code signalling before merge. The two suggestions broaden the state-detection and close the silent-failure gap on the fresh-run path; worth considering in the same iteration.

---
Generated by `/code-review` — dev-core

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, panel-verified):** 1 finding
**Applied via 1b1 (panel consensus):** 2 findings

### Auto-Applied
- `[applied]` nitpick: drop `|| true` on fresh-run migrate path — devops 95%, architect verifier 88%. Commit `28a7a43`.

### Applied (1b1, after devops↔architect panel)
Both findings were initially rejected by the verifier (C=30 and C=35). On rebuttal, the panel converged:

- `[applied]` suggestion: broaden state check to include named volumes — devops 45% concede, architect 72% concede. `podman images -q || podman volume ls -q`. Volume data keeps old subuid ownership after namespace rebase → same orphaning hazard even without images present. Dropped `ps -aq` (adds marginal value).
- `[applied]` issue: hard-fail when `FORCE_MIGRATE` unset + prior state exists — devops 80% hold, architect 78% concede. The hunk runs only when subuid ranges *just changed*, so continuing with exit 0 layers services on a known-inconsistent mapping; `curl | bash` warnings scroll off screen. `error` exits non-zero; `FORCE_MIGRATE=1` remains the knowing opt-in.

Both applied in commit `90f2406`.

### Panel notes
- Verifier's initial "migrate only remaps image layers" was incorrect — volume files inside `~/.local/share/containers/storage/volumes/` retain UID ownership across the namespace rebase.
- Verifier's initial "warn-and-continue is intentional idempotency" did not survive the branch-scoped context (the hunk is specifically the *mutation* path where state just became inconsistent).

---
Generated by `/fix` — dev-core

PR #887, opened 2026-04-22
feat(core): PoolManager bounded + LRU eviction + iteration lock

MickaelV0:
## Summary
- Add `max_pools` to `HubConfig` (default 500) for bounded pool count
- Convert `PoolManager.pools` to `OrderedDict` with `threading.Lock` for thread-safe LRU tracking
- Implement `_evict_lru()` for capacity-based eviction (pop leftmost on hit cap)
- Add validation: `max_pools <= 0` raises `ValueError` at Hub init
- Add tests: lock safety (100 threads), LRU eviction, max_pools bound

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #859: PoolManager bounded + LRU eviction | OPEN |
| Frame | [859-poolmanager-bounded-lru-eviction-frame.mdx](../artifacts/frames/859-poolmanager-bounded-lru-eviction-frame.mdx) | Approved |
| Spec | [859-poolmanager-bounded-lru-eviction-spec.mdx](../artifacts/specs/859-poolmanager-bounded-lru-eviction-spec.mdx) | Approved |
| Implementation | 1 commit on `feat/859-poolmanager-bounded-lru-eviction` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (8 new) | Passed |

## Test Plan
- [ ] `uv run pytest tests/core/test_pool_manager.py` — all 20 tests pass
- [ ] `uv run pyright src/lyra/core/` — 0 errors
- [ ] Verify LRU eviction: create 3 pools, touch middle, create 4th → oldest evicted

Closes #859

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/pr`

PR #903, opened 2026-04-23
test(core): cover set_cancel_on_new_message + document PoolManager threading model

### LINKED ISSUE #888: test(core): add tests for set_cancel_on_new_message + document threading model
## Context

Deferred from PR #887 review (consensus panel decision).

## Tasks

- [ ] Add `TestSetCancelOnNewMessage` test class mirroring `TestSetDebounceMs` pattern
- [ ] Add threading model documentation to `docs/ARCHITECTURE.md` (hybrid threading.Lock + asyncio.create_task)

## Acceptance Criteria

- `set_cancel_on_new_message()` has test coverage
- ARCHITECTURE.md has section explaining PoolManager threading model

## Related

- #859 — PoolManager bounded + LRU eviction
- #887 — PR with review findings

MickaelV0:
## Summary
- Add `TestSetCancelOnNewMessage` mirroring `TestSetDebounceMs` (hub default, live pools, new pools).
- Add a `PoolManager threading model` subsection to `docs/ARCHITECTURE.md` explaining the hybrid `threading.Lock` + `asyncio.create_task` approach, invariants, and the "never await under the sync lock" rule.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #888: tests for set_cancel_on_new_message + document threading model | OPEN |
| Implementation | 1 commit on `feat/888-cancel-tests-threading-doc` | Complete |
| Verification | Lint ✅  Typecheck ✅  Tests ✅ (3 new) | Passed |

## Test Plan
- [ ] `uv run pytest tests/core/test_pool_manager.py::TestSetCancelOnNewMessage -o addopts=""` → 3 passed
- [ ] `uv run pytest tests/core/test_pool_manager.py -o addopts=""` → 23 passed (full file regression)
- [ ] Render `docs/ARCHITECTURE.md` and confirm the new `PoolManager threading model` section sits under `Discussion Pools`.

Closes #888

Deferred from PR #887 review (consensus panel decision).

---


MickaelV0:
## Code Review

Domain agents: `tester`, `architect`. 2 files, +66/-1.

### Blockers / Issues

**issue**: Invariant 2 implies eviction always schedules a flush, but the LRU eviction path silently skips it
  `docs/ARCHITECTURE.md` (new Invariant 2)
  -- architect
  Root cause: `_evict_lru_locked` (pool_manager.py:105) calls `self._pools.popitem(last=False)` to remove the entry, then calls `_evict_pool_locked(pid, reason="lru")`. Inside `_evict_pool_locked`, `self._pools.pop(pool_id, None)` returns `None` (pool already removed) and the method returns early at line 112–113 — no `create_task` is issued. Only the TTL/stale eviction path reaches the `create_task` branch. The invariant as written implies every eviction schedules a flush.
  Solutions:
    1. Tighten the invariant: "Eviction via TTL drops the pool from `_pools` before scheduling `flush_session`; LRU eviction (capacity cap) does not schedule a flush — the pool is silently discarded." (recommended)
    2. Fix the implementation so LRU eviction also schedules a flush (out of scope for this doc review).
  Confidence: 97%

**issue**: GC cancellation claim is inaccurate — Python's GC does not cancel asyncio Tasks
  `docs/ARCHITECTURE.md` (Invariant 3 prose)
  -- architect
  Root cause: CPython's asyncio does not cancel a Task when it is garbage-collected — it emits a `RuntimeWarning` ("Task was destroyed but it is pending") but the coroutine is not cancelled. The actual guarantee `_memory_tasks` provides is keeping a strong reference so the task is not eligible for GC at all.
  Solutions:
    1. Replace with: "strong refs keep tasks ineligible for GC; without this, CPython would destroy the pending task and emit a RuntimeWarning." (recommended)
    2. Replace "cancelling" with "destroying" — marginally more accurate but still imprecise.
  Confidence: 93%

**issue**: `test_updates_hub_default` asserts only the hub attribute, not the `_pool_config` snapshot
  `tests/core/test_pool_manager.py:163–169`
  -- tester
  Root cause: `set_cancel_on_new_message` makes two writes: `hub._cancel_on_new_message` **and** `self._pool_config = replace(...)`. The test verifies the first write but not the second. If the `replace()` line broke, no pool created *after* the call would get the updated default, but this test would still pass. (`TestSetDebounceMs` has the same gap — new test mirrors it.)
  Solutions:
    1. Add `assert hub._pool_manager._pool_config.cancel_on_new_message is True` after the setter call — directly verifies the snapshot that `get_or_create_pool` passes to new `Pool` instances. (recommended)
    2. Drop the dedicated test and rely on `test_new_pools_use_updated_value` only — acceptable but loses the direct invariant assertion.
  Confidence: 90%

### Suggestions

**suggestion**: Invariant 4 overstates atomicity — hub-level field mutations precede lock acquisition
  `docs/ARCHITECTURE.md` (Invariant 4)
  -- architect
  Root cause: `set_debounce_ms` writes `self._hub._debounce_ms` and `self._pool_config` at lines 154–155, then acquires `_lock` at line 156. A concurrent reader of `self._hub._debounce_ms` between lines 155 and 156 would see the new config value before any pool has been updated.
  Solutions:
    1. Qualify: "the per-pool iteration is atomic, but the hub-level snapshot update is written before the lock is taken; a narrow window exists where hub config and pool config diverge." (recommended)
    2. Drop "atomically" and replace with "under `_lock`" (scoped only to the pool iteration).
  Confidence: 95%

**suggestion**: extend `_make_hub` to accept `cancel_on_new_message` instead of duplicating setup boilerplate
  `tests/core/test_pool_manager.py:163–194`
  -- tester
  Root cause: `_make_hub` only wraps `pool_ttl` and `debounce_ms`, so `TestSetCancelOnNewMessage` inlines `Hub(cancel_on_new_message=False)` + `inbound_bus.register(...)` in every method (3× two-line blocks).
  Solutions:
    1. Add `cancel_on_new_message: bool | None = None` to `_make_hub` and forward to `Hub(...)` — mirrors the existing `debounce_ms` pattern. All three methods collapse to one-liners. (recommended)
    2. Use a `@pytest.fixture` — heavier machinery; not worth it here.
  Confidence: 95%

**suggestion**: `create_task` safety claim omits the required precondition — must run on the event loop thread
  `docs/ARCHITECTURE.md` ("Why asyncio.create_task is safe…" paragraph)
  -- architect
  Root cause: `asyncio.create_task` requires a running event loop on the calling thread. Current call sites satisfy this implicitly; the doc does not state it.
  Solutions:
    1. Add: "safe here because `_evict_pool_locked` is invoked from within a running coroutine (event loop thread). A background-thread caller would need `loop.call_soon_threadsafe` instead." (recommended)
    2. Short parenthetical: "(safe only because this code path runs on the event loop thread)."
  Confidence: 88%

### Nitpicks

**nitpick**: `test_updates_hub_default` redundantly pre-asserts the initial value
  `tests/core/test_pool_manager.py:166`
  -- tester
  Root cause: `assert hub._cancel_on_new_message is False` re-checks the constructor argument from the line above.
  Solutions:
    1. Remove the pre-assert; begin the test body with the setter call. (recommended)
    2. Keep as defensive guard — low cost.
  Confidence: 80%

### Praise

**praise**: docs section is accurate on the core claims — `threading.Lock` rationale, `flush_pool` drops lock before await, strong-ref task pattern. Line-number references (`flush_pool` at line 142) are exact.

---

### Summary

- 3 issues (97/93/90%) — 2 doc inaccuracies that obscure real behavior + 1 missing test assertion
- 3 suggestions (95/95/88%) — atomicity/precondition qualifiers + helper extension
- 1 nitpick, 1 praise

**Verdict: Request changes.** The three issues are concrete factual gaps: Invariant 2 hides a real LRU/flush gap in the code, the GC claim is technically wrong, and the hub-default test would pass even if `_pool_config` replacement broke.

MickaelV0:
## Review Fixes Applied

**Applied inline (user accepted all findings):** 6 finding(s) + 1 nitpick
**Failed:** 0

### Applied
- [applied] issue: Invariant 2 LRU silent-skip clarified (97%) — `docs/ARCHITECTURE.md`
- [applied] issue: Invariant 3 GC wording corrected (strong refs vs cancellation) (93%) — `docs/ARCHITECTURE.md`
- [applied] issue: `_pool_config.cancel_on_new_message` assertion added to `test_updates_hub_default` (90%) — `tests/core/test_pool_manager.py`
- [applied] suggestion: Invariant 4 atomicity qualified — hub-level write precedes lock (95%) — `docs/ARCHITECTURE.md`
- [applied] suggestion: `_make_hub` extended with `cancel_on_new_message`; test boilerplate collapsed (95%) — `tests/core/test_pool_manager.py`
- [applied] suggestion: `create_task` safety — event-loop-thread precondition added (88%) — `docs/ARCHITECTURE.md`
- [applied] nitpick: removed redundant pre-assert of initial hub default (80%) — `tests/core/test_pool_manager.py`

Commit: a0c54d2
Gate: ruff ✅ · test_pool_manager.py 23 passed

PR #910, opened 2026-04-24
refactor(scripts): extract deploy-lib.sh (ADR-055 D5) + QUADLET-ADOPTION guide

MickaelV0:
## Summary
- Extract reusable Quadlet deploy logic into `scripts/deploy-lib.sh` (356 L) per ADR-055 D5; downstream projects (voiceCLI, imageCLI, llmCLI, 2ndBrain) source it via `${LYRA_DEPLOY_LIB:-$HOME/.local/lib/roxabi/deploy-lib.sh}` and supply per-project variables.
- Refactor `scripts/deploy-quadlet.sh` from 205 → 45 lines as a thin wrapper — behavior preserved (fail-file, hub-readiness gating, env-file 0600 check, image-digest tamper check, tags summary).
- Add `make quadlet-install-deploy-lib` + `make quadlet-upgrade-lib` — install lib to `~/.local/lib/roxabi/` with commit SHA pinned in header.
- Add `docs/QUADLET-ADOPTION.md` — project-agnostic adoption guide covering prereqs, unit templates, credentials, deploy-script skeleton, Makefile skeleton, bootstrap checklist, coordination triggers, troubleshooting. Cross-links ADR-055 / ADR-053 / ADR-054.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | ADR-055 Follow-up #4 (`docs/architecture/adr/055-quadlet-ecosystem-conventions.mdx` §Implementation issues) | Accepted 2026-04-24 |
| Spec | ADR-055 D5 (lines 210–251) defines interface, versioning, bootstrap-copy model | Complete |
| Implementation | 4 commits on `worktree-deploy-lib-extract` | Complete |
| Verification | Lint ✅ · `bash -n` both scripts ✅ · dry-run nothing-changed path exits 0 ✅ | Passed |

## Judgment calls flagged for review

1. **`git reset --hard` kept in `check_repo` rollback path** — global rule says ¬`--hard`, but this is deploy-side rollback on a remote checkout, not git-history manipulation. Preserving original behavior; open to switching to `--merge` or SHA re-checkout on request.
2. **Tests run on primary repo only** — formalized from the original (voiceCLI extra-repo never ran its own test suite). Configurable via `PROJECT_TEST_CMD`.
3. **`EXPECTED_IMAGE_ID` as lib-level global** — mirrors the original script-level global; callers can preset it if they pre-build outside the lib.
[...]

PR #916, opened 2026-04-24
fix(scripts): apply B7 pushd/popd discipline to build_image (follow-up to #910)

MickaelV0:
## Summary
One-line correctness fix missed by PR #910 merge. Re-review flagged a bare `cd "$PROJECT_DIR"` in `build_image` that mutates the caller's `$PWD` — same anti-pattern that B7 fixed in `check_repo`. Applied the same `pushd`/`popd` + `RETURN` trap for consistency.

## Why it matters
`build_image` currently leaves the shell in `$PROJECT_DIR` after it returns. Lyra's happy path doesn't hit this (subsequent ops use absolute paths), but downstream callers of the shared lib can't assume side-effect-free PWD semantics.

## Test Plan
- [x] `bash -n scripts/deploy-lib.sh` passes
- [x] Lyra deploy path uses absolute paths after `build_image` — no regression expected
- [ ] Verify on M₁ next deploy cycle

Context: #910 re-review iteration 2.

### DIFF
--- a/scripts/deploy-lib.sh
+++ b/scripts/deploy-lib.sh
@@ -178,7 +178,8 @@ build_image() {
     local dockerfile="${DOCKERFILE:-Dockerfile}"
     local digest_dir="$HOME/.${PROJECT}"
 
-    cd "$PROJECT_DIR"
+    pushd "$PROJECT_DIR" > /dev/null
+    trap 'popd > /dev/null 2>&1 || true' RETURN
 
     log "Tagging current image as rollback..."
     podman tag "${IMAGE}" "${IMAGE%:*}:rollback" 2>/dev/null \


MickaelV0:
## Panel Review (2 architect + 2 devops)

**Verdict: ✅ Approve**

| Reviewer | Finding |
|---|---|
| architect (design) | B7 pattern applied faithfully; no nested RETURN trap conflict. Small DRY nit — extract `_enter_dir`/`_leave_dir` helper in a follow-up |
| architect (ecosystem) | Pure internal fix, no caller contract change. Two lightweight doc additions worth doing before voiceCLI adopts: one line in `QUADLET-ADOPTION.md` §5 ("functions are PWD-clean") + one in `deploy-lib.sh` header ("library contract: use pushd/trap RETURN, never bare cd") |
| devops (operational) | Operationally correct. Minor theoretical gap: `set -e` exit bypasses RETURN trap, leaving pushd stack unbalanced — but the process is dying, no impact |
| devops (deploy impact) | Boring and fine. Next normal timer tick on M₁ verifies |

No blockers. The 2 doc suggestions are worth a ~10-line follow-up PR once voiceCLI consumer lands.

PR #918, opened 2026-04-24
fix(quadlet): bind-mount ~/.claude.json into lyra-hub

MickaelV0:
## Summary
- Add inline \`Volume=%h/.claude.json:/home/lyra/.claude.json:z\` to lyra-hub.container
- Required because the Claude CLI stores top-level config at \`~/.claude.json\` outside \`~/.claude/\`
- Uses same pattern as config.toml (inline Volume=, not .volume unit)

## Test plan
- [ ] Deploy to M₁ and verify hub container starts cleanly
- [ ] Confirm CLI inside container reads expected settings

### DIFF
--- a/deploy/quadlet/lyra-hub.container
+++ b/deploy/quadlet/lyra-hub.container
@@ -37,7 +37,10 @@ Environment=LYRA_LOG_DIR=/home/lyra/.local/state/lyra/logs
 Volume=%h/.lyra/config.toml:/app/config.toml:ro,z
 # claude CLI credentials — host ~/.claude is bind-mounted read-write so the
 # CLI can refresh tokens. UserNS=keep-id maps UID 1500 → host UID 1000 (mickael).
+# ~/.claude.json is a separate top-level config file the CLI looks up outside
+# of ~/.claude/ — must be bind-mounted inline (same pattern as config.toml).
 Volume=%h/.claude:/home/lyra/.claude:z
+Volume=%h/.claude.json:/home/lyra/.claude.json:z
 # Health endpoint — probed by lyra-monitor via host loopback (localhost:8443).
 # Bind to 0.0.0.0 inside container so PublishPort forwarding reaches uvicorn;
 # host exposure is restricted to 127.0.0.1 by PublishPort below.


MickaelV0:
## Code Review

Single-file, +4-line diff on `deploy/quadlet/lyra-hub.container` adding an inline bind-mount of `~/.claude.json` into the container at `/home/lyra/.claude.json`. Reviewed by **security-auditor** + **devops**.

### Blockers

**issue(blocking): missing host-existence guard — Podman auto-creates `~/.claude.json` as a directory if absent**
  `deploy/quadlet/lyra-hub.container:43`
  -- devops + security-auditor (converged, C=92%)
  Root cause: On a host where `~/.claude.json` does not yet exist (fresh prod machine, post-wipe), Podman bind-mount semantics auto-create the source as a **directory**. The container starts "successfully" but the Claude CLI hits `IsADirectoryError` silently. Same failure mode the comment block at lines 34–36 calls out for `.volume` units — applies here too. The `config.toml` mount on line 37 is mitigated by being required (container crashes loudly without valid config); `.claude.json` is optional, so the directory trap goes unnoticed longer.
  Solutions:
    1. Add `ExecStartPre=/bin/bash -c 'test -f %h/.claude.json || touch %h/.claude.json'` in `[Service]` (recommended)
    2. Add `ConditionPathExists=%h/.claude.json` in `[Unit]` — fails cleanly before Podman evaluates the mount
  Confidence: 92%

**issue(blocking): read-write mount exposes live OAuth account metadata to the container**
  `deploy/quadlet/lyra-hub.container:43`
  -- security-auditor
  Root cause: `~/.claude.json` holds `oauthAccount` (accountUuid, emailAddress, organizationUuid, billingType, workspaceRole) and `userID`. Mounted RW (no `:ro`), any write path inside the container — compromised dep, prompt-injection, CLI bug — can overwrite or corrupt the host identity file. The sibling `~/.claude` dir mount shares this exposure, but `.claude.json` at home-root contains persistent identity fields.
  Solutions:
    1. Add `:ro` — `Volume=%h/.claude.json:/home/lyra/.claude.json:ro,z`. Verify first with `strace`/`inotifywait` that token refresh does not write `.claude.json` itself (likely writes only under `~/.claude/`). (recommended)
    2. Copy into a named volume via `ExecStartPre`, mount that RW, sync back on stop — keeps host file out of direct write reach.
  Confidence: 88%

**issue(blocking): `:z` relabels a home-root file host-wide**
  `deploy/quadlet/lyra-hub.container:43`
  -- security-auditor
  Root cause: `:z` (lowercase) runs `chcon` on the bind source, setting `container_file_t`. On `~/.claude.json` at `$HOME/`, this changes SELinux context of a home-root file host-wide. Any other confined process that depended on `user_home_t` loses access silently. Low-impact on Ubuntu (AppArmor, not SELinux enforcing by default), higher on SELinux-enforcing hosts.
  Solutions:
    1. Use `:Z` (uppercase) — private per-container label, does not relabel host-visible context. (recommended — but verify host CLI still works afterward, since `:Z` isolates)
    2. Pre-copy into named volume, eliminates relabel concern.
  Confidence: 75%
  Note: pre-existing on the `~/.claude:z` mount (line 42) — not introduced by this PR, but worth addressing together.

### Suggestions

**suggestion: `~/.claude` directory mount is RW over 1 200+ session dirs**
  `deploy/quadlet/lyra-hub.container:42`
  -- security-auditor
  Root cause: `~/.claude/sessions/` holds full conversation history incl. MEMORY.md files and any secrets typed into Claude Code. Token-refresh justification targets credential files, not session history. Container-compromise blast radius extends to exfil/corruption of all past conversations.
  Solutions:
    1. Audit which paths CLI writes during refresh (likely `~/.claude/.credentials.json`); replace coarse dir mount with scalpel mounts: `:ro,z` on bulk + `:rw,z` on credentials only. (recommended)
    2. Make whole dir `:ro,z`; pre-refresh credentials at deploy time.
  Confidence: 82%
  Note: pre-existing, not introduced by this PR.

**suggestion: `.claude.json` is a config file, not credentials — `:ro` likely sufficient**
  `deploy/quadlet/lyra-hub.container:43`
  -- devops
  Root cause: `.claude.json` stores model prefs, project list, telemetry opt-out. The CLI does write to it (updates last-used project, command history), so `:ro` may break functionality. Verify empirically before choosing.
  Confidence: 78% (overlaps with blocker #2 — same fix)

**question: should `.claude.json` be symlinked under `~/.claude/` instead?**
  `deploy/quadlet/lyra-hub.container:43`
  -- devops
  Root cause: Podman does not follow symlinks escaping the bind-mount root, so the current two-mount approach is correct. A comment explaining why the symlink approach does not work would prevent a future "simplification" that silently breaks things.
  Solutions:
    1. Add one-line comment: `# symlink inside ~/.claude/ does not work — Podman does not follow symlinks escaping bind-mount root` (recommended)
  Confidence: 85%

**nitpick: comment gap — `keep-id` mapping applies to both mounts**
  `deploy/quadlet/lyra-hub.container:38-41`
  -- devops
  Root cause: Existing comment explains `keep-id` UID mapping for `~/.claude` only; reader has to re-derive that the same applies to `~/.claude.json`.
  Solutions:
    1. Extend comment to cover both mounts.
  Confidence: 70%

### Summary

| | Count |
|---|---|
| Blockers | 3 |
| Suggestions | 3 |
| Nitpicks | 1 |

**Verdict: Request changes.**

Two of three blockers are real fixes for this PR (missing file guard, RW exposure). The `:z` → `:Z` blocker is worth addressing but is pre-existing on the sibling mount — could defer as a separate sweep.

PR #957, opened 2026-04-27
fix(quality): resolve all actionable code quality exceptions

MickaelV0:
## Summary
- **BLE001 (13)**: resilient catches self-documented with reason tags; narrowed to specific types where possible (e.g. `json.JSONDecodeError` in refiner stages)
- **S608 (3)**: replaced `# noqa: S608` with inline safety comment — table/column names are validated by `_IDENT_RE` before interpolation
- **E501 (10)**: reformatted long lines — split `BuiltinHandler` type alias, SQL trigger strings, and long inline comments
- **ANN001**: annotated `apply_schema_compat(db)` as `aiosqlite.Connection`; removed noqa
- **`[attr-defined]` (19)**: `TraceLogRecord(logging.LogRecord)` subclass for log record fields; typed Protocol + `cast` for cli_pool cross-mixin calls; explicit `Route` import + `_PartialMessageable` cast for discord outbound; `_configured: bool` declared in `Pool.__init__`
- **`[arg-type]` (1)**: cast `CliControlCmd.op` to the correct `Literal` type in cli_nats driver
- **E402 (4)**: self-documented intentional late imports with reason tags

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Implementation | 1 commit on `fix/code-quality-exceptions` | Complete |
| Verification | Lint ✅ Typecheck ✅ All pre-push hooks ✅ | Passed |

## Test Plan
- [ ] `uv run pyright src/` → 0 errors, 0 warnings, 0 informations
- [ ] `uv run ruff check src/` → All checks passed
- [ ] All pre-commit + pre-push hooks green

---

### DIFF
--- a/deploy/nats/acl-matrix.json
+++ b/deploy/nats/acl-matrix.json
@@ -4,6 +4,7 @@
     "hub": {
       "owner": "lyra",
       "description": "Hub core — routes inbound messages, dispatches to voice/LLM/image/clipool workers. Inbox narrowed to _INBOX.hub.> per ADR-051 per-identity prefix invariant.",
+      "notes": "Stream provisioning is a deploy-time admin operation — hub does not create/update streams at runtime. $JS.API.STREAM.CREATE/UPDATE.LYRA_AUDIT are intentionally absent from publish ACLs.",
       "publish": [
         "lyra.outbound.telegram.>",
         "lyra.outbound.discord.>",
@@ -13,8 +14,6 @@
         "lyra.image.generate.request",
         "lyra.clipool.cmd",
         "lyra.clipool.control",
-        "$JS.API.STREAM.CREATE.LYRA_AUDIT",
-        "$JS.API.STREAM.UPDATE.LYRA_AUDIT",
         "lyra.audit.>"
       ],
       "subscribe": [
--- a/deploy/nats/auth.conf
+++ b/deploy/nats/auth.conf
@@ -13,7 +13,7 @@ authorization {
       nkey: "UDUMMYHUB"
       # hub
       permissions: {
-        publish:   { allow: ["lyra.outbound.telegram.>","lyra.outbound.discord.>","lyra.voice.tts.request.>","lyra.voice.stt.request.>","lyra.llm.request","lyra.image.generate.request","lyra.clipool.cmd","lyra.clipool.control","$JS.API.STREAM.CREATE.LYRA_AUDIT","$JS.API.STREAM.UPDATE.LYRA_AUDIT","lyra.audit.>"] }
+        publish:   { allow: ["lyra.outbound.telegram.>","lyra.outbound.discord.>","lyra.voice.tts.request.>","lyra.voice.stt.request.>","lyra.llm.request","lyra.image.generate.request","lyra.clipool.cmd","lyra.clipool.control","lyra.audit.>"] }
         subscribe: { allow: ["lyra.inbound.telegram.>","lyra.inbound.discord.>","lyra.voice.tts.heartbeat","lyra.voice.stt.heartbeat","lyra.llm.health.*","lyra.system.ready","_INBOX.hub.>","lyra.image.heartbeat","lyra.clipool.heartbeat"] }
         allow_responses: true
       }
@@ -41,7 +41,7 @@ authorization {
       # tts-adapter
       permissions: {
         publish:   { allow: ["lyra.voice.tts.heartbeat","lyra.system.ready"] }
-        subscribe: { allow: ["lyra.voice.tts.request","lyra.voice.tts.request.>","_INBOX.tts-adapter.>","_inbox.tts-adapter.>"] }
+        subscribe: { allow: ["lyra.voice.tts.request","lyra.voice.tts.request.>","_INBOX.tts-adapter.>","_inbox.tts-adapter.>"] }  # lowercase variant for nats.py <2.2 compatibility
         allow_responses: true
       }
     }
@@ -50,7 +50,7 @@ authorization {
       # stt-adapter
       permissions: {
         publish:   { allow: ["lyra.voice.stt.heartbeat","lyra.system.ready"] }
-        subscribe: { allow: ["lyra.voice.stt.request","lyra.voice.stt.request.>","_INBOX.stt-adapter.>","_inbox.stt-adapter.>"] }
+        subscribe: { allow: ["lyra.voice.stt.request","lyra.voice.stt.request.>","_INBOX.stt-adapter.>","_inbox.stt-adapter.>"] }  # lowercase variant for nats.py <2.2 compatibility
         allow_responses: true
       }
     }
@@ -59,7 +59,7 @@ authorization {
       # voice-tts
       permissions: {
         publish:   { allow: ["lyra.voice.tts.heartbeat","lyra.system.ready"] }
-        subscribe: { allow: ["lyra.voice.tts.request","lyra.voice.tts.request.>","_INBOX.voice-tts.>","_inbox.voice-tts.>"] }
+        subscribe: { allow: ["lyra.voice.tts.request","lyra.voice.tts.request.>","_INBOX.voice-tts.>","_inbox.voice-tts.>"] }  # lowercase variant for nats.py <2.2 compatibility
         allow_responses: true
       }
     }
@@ -68,7 +68,7 @@ authorization {
       # voice-stt
       permissions: {
         publish:   { allow: ["lyra.voice.stt.heartbeat","lyra.system.ready"] }
-        subscribe: { allow: ["lyra.voice.stt.request","lyra.voice.stt.request.>","_INBOX.voice-stt.>","_inbox.voice-stt.>"] }
+        subscribe: { allow: ["lyra.voice.stt.request","lyra.voice.stt.request.>","_INBOX.voice-stt.>","_inbox.voice-stt.>"] }  # lowercase variant for nats.py <2.2 compatibility
         allow_responses: true
       }
     }
@@ -78,7 +78,7 @@ authorization {
       permissions: {
         publish:   { allow: ["lyra.llm.health.*"] }
         subscribe: { allow: ["lyra.llm.request"] }
-        allow_responses: true
+        allow_responses: true  # intentional: uses allow_responses only; nats_connect() must NOT use identity_name=
       }
     }
     {
@@ -86,7 +86,7 @@ authorization {
       # image-worker
       permissions: {
         publish:   { allow: ["lyra.image.heartbeat"] }
-        subscribe: { allow: ["lyra.image.generate.request","_INBOX.image-worker.>","_inbox.image-worker.>"] }
+        subscribe: { allow: ["lyra.image.generate.request","_INBOX.image-worker.>","_inbox.image-worker.>"] }  # lowercase variant for nats.py <2.2 compatibility
         allow_responses: true
       }
     }
@@ -96,7 +96,7 @@ authorization {
       permissions: {
         publish:   { allow: ["lyra.monitor.>"] }
         subscribe: { allow: ["lyra.monitor.>"] }
-        allow_responses: true
+        allow_responses: true  # intentional: uses allow_responses only; nats_connect() must NOT use identity_name=
       }
     }
     {
@@ -104,7 +104,7 @@ authorization {
       # clipool-worker
       permissions: {
         publish:   { allow: ["lyra.clipool.heartbeat","lyra.system.ready"] }
-        subscribe: { allow: ["lyra.clipool.cmd","lyra.clipool.control","_INBOX.clipool-worker.>","_inbox.clipool-worker.>"] }
+        subscribe: { allow: ["lyra.clipool.cmd","lyra.clipool.control","_INBOX.clipool-worker.>","_inbox.clipool-worker.>"] }  # lowercase variant for nats.py <2.2 compatibility
         allow_responses: true
       }
     }
--- a/docs/architecture/adr/060-cli-protocol-circular-import-resolution.mdx
+++ b/docs/architecture/adr/060-cli-protocol-circular-import-resolution.mdx
@@ -1,0 +1,63 @@
+---
+title: "ADR-060: CLI Protocol Circular Import Resolution via Type Extraction"
+description: Extract shared types from cli_protocol.py into cli_protocol_types.py to eliminate circular import between the facade and its submodules.
+---
+
+## Status
+
+Accepted
+
+## Context
+
+`cli_protocol.py` served dual roles: it defined shared types (`CliResult`,
+`CliProtocolOptions`, `SESSION_ID_RE`, `build_cmd`, `_read_stderr_snippet`)
+and acted as a re-export facade for `cli_non_streaming` and `cli_streaming`.
+
+This created a circular import: `cli_non_streaming` and `cli_streaming` needed
+the shared types at module load time, but those types lived in `cli_protocol`,
+which tried to import from those submodules.  The workaround was bottom-of-file
+re-exports suppressed with `# noqa: E402, I001`, masking the structural problem.
+
+`cli_pool` also imported from `cli_protocol`, consuming `_SESSION_ID_RE`,
+`CliProtocolOptions`, `CliResult`, and `send_and_read` — it was a downstream
+consumer, not a contributor to the cycle.
+
+## Options Considered
+
+### Option A: Extract shared types into `cli_protocol_types.py`
+- **Pros:** Eliminates the circular import structurally; `cli_protocol.py`
+  becomes a clean top-of-file re-export facade with no `noqa` suppressions;
+  submodules import from `cli_protocol_types` (no cycle); all external callers
+  continue importing from `cli_protocol` unchanged.
+- **Cons:** Adds one new file; requires updating imports in two files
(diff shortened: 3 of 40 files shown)


MickaelV0:
## Code Review

**28 files | 6 domain agents | Verdict: Request changes**

---

### 🔴 Blockers

**issue: `json.JSONDecodeError` narrowing misses `TypeError` — introduced by this PR**
`src/lyra/core/agent/agent_refiner_stages.py:43,49`
Root cause: `json.loads()` raises `TypeError` when input is not `str|bytes|bytearray`. The truthiness guard filters `None`/`""` but not a type violation. The original broad `except Exception` covered this; the narrowed version now lets `TypeError` escape to the refiner caller.
Fix: `except (json.JSONDecodeError, TypeError): pass`
Confidence: 92%

---

**issue: `cast(Literal[...], op)` is a static no-op — type constraint must live in the signature**
`src/lyra/llm/drivers/cli_nats.py:181`
Root cause: `cast()` is identity at runtime. An invalid `op` string reaches `CliControlCmd` and raises `ValidationError` unhandled. Change `op: str` → `op: Literal["reset", "resume_and_reset", "switch_cwd"]` in `_build_control_payload` and remove the cast.
Confidence: 88%

---

**issue: `_PartialMessageable` union is missing `discord.StageChannel`**
`src/lyra/adapters/discord/discord_outbound.py:34-36`
Root cause: `fetch_channel()` can return `StageChannel` (a `GuildChannel` + `Messageable` with `get_partial_message()`). The alias is incorrect by its own docstring. Add `discord.StageChannel` to the union.
Confidence: 90%

---

**issue: Protocol Ellipsis (`= ...`) used as runtime default value, not stub marker**
`src/lyra/core/cli/cli_pool_types.py:19-28`
Root cause: `preserve_session: bool = ...` assigns `Ellipsis` as the actual runtime default. Use real defaults: `preserve_session: bool = True`, `system_prompt: str = ""` — matching `CliPoolWorkerMixin._spawn`/`._kill` exactly.
Confidence: 88%

---

**issue: `_CliPoolCore` in `__all__` contradicts the private-name convention**
`src/lyra/core/cli/cli_pool_types.py:7`
Root cause: `__all__` defines the star-import surface. A `_`-prefixed name in `__all__` is self-contradictory. The `__all__` was added to silence a pyright `reportUnusedClass` false positive — but the name IS used via explicit imports in lifecycle/streaming, which already suppresses that warning. Remove `__all__ = ["_CliPoolCore"]` entirely.
Confidence: 82%

---

**issue: Unvalidated `idx_sql` DDL replayed verbatim from `sqlite_master`**
`src/lyra/bootstrap/bootstrap_stores.py:128-138`
Root cause: `_atomic_table_copy` reads raw `sql` from `sqlite_master` for indexes and executes without validation. A tampered `auth.db` (symlink attack, rogue restore) can execute arbitrary DDL. Table/column names are validated via `_IDENT_RE` but index SQL is not. Add allowlist regex before execute: `re.match(r'^CREATE\s+(UNIQUE\s+)?INDEX\s+', idx_sql, re.IGNORECASE)`.
Confidence: 72%

---

**issue: Hub granted `$JS.API.STREAM.CREATE/UPDATE.LYRA_AUDIT` — audit log mutable by hub at runtime**
`deploy/nats/auth.conf:16`
Root cause: A compromised hub process can silently purge or reconfigure the audit JetStream stream (change retention, discard policy, max_msgs). Stream provisioning should be a deploy-time admin operation. Remove these two subjects from hub's publish allow-list; provision the stream in `gen-nkeys.sh` under a one-shot admin key.
Confidence: 82%

---

**issue: `TMPDIR_LOCAL=$(mktemp -d)` shadows the POSIX `$TMPDIR` env var for all subprocesses**
`scripts/check-acl-authconf-drift.sh`
Root cause: `TMPDIR` is read by `mktemp`, `sort`, `tar`, and others. Overwriting it in a non-function scope propagates to all subprocesses in the shell step. Rename to `NATS_TMPDIR`.
Confidence: 92%

---

**issue: `find_sources` uses `grep -z` which is GNU-only and fails on macOS (BSD grep)**
`scripts/check-inbox-prefix.sh:22-23`
Root cause: `grep -z` (null-terminated I/O) is not available on macOS BSD grep. Local pre-push runs on macOS silently miss violations. Replace `find … | grep -zvE '/tests/'` with `find src packages -name "*.py" -not -path "*/tests/*" -print0`.
Confidence: 85%

---

**issue: Docstring detection heuristic `` grep '``inbox_prefix=' `` is brittle — real violations can be suppressed as "docstrings"**
`scripts/check-inbox-prefix.sh:42`
Root cause: Any line containing two backticks + `inbox_prefix=` triggers the exemption regardless of context. Also asymmetric: the f-string pass (line 27) has no equivalent exemption. Replace with an AST-based check or tighten the heuristic to require line starts with `#`.
Confidence: 88%

---

### 🟡 Warnings

**suggestion: `normalize_authconf` strips full-line `#` but not trailing inline comments — spurious diff risk**
`scripts/check-acl-authconf-drift.sh:42-47`
Add `-e 's/[[:space:]]*#[^"]*$//'` before whitespace normalization.
Confidence: 80%

**suggestion: `cli_setup.py` plugin load failure logged at `debug` — too low for operator-facing setup command**
`src/lyra/cli_setup.py:87`
Change `log.debug` → `log.warning`. Confidence: 85%

**suggestion: `discord.http.Route` is an undocumented internal discord.py API**
`src/lyra/adapters/discord/discord_audio_outbound.py:11`
Add version pin comment (`discord.py==X.Y.Z`) documenting the known-internal dependency.
Confidence: 85%

**suggestion: `getattr(pool, "_configured", False)` now redundant — `Pool.__init__` always sets it**
`src/lyra/core/hub/middleware/middleware_pool.py:107`
Replace with `not pool._configured`. Confidence: 80%

**suggestion: `TraceLogRecord(logging.LogRecord)` should be a `Protocol`, not a subclass — never instantiated**
`src/lyra/core/trace.py:31-37`
Use `class TraceLogRecord(Protocol): trace_id: str; pool_id: str; agent_name: str` and cast to it.
Confidence: 78%

**suggestion: `cli_protocol.py` circular import noqa papers over a structural role-conflict**
`src/lyra/core/cli/cli_protocol.py:140-141`
Extract shared types into `cli_protocol_types.py` so `cli_protocol.py` becomes a clean facade. Or document the load-order invariant at module level (not just per-line noqa).
Confidence: 85%

**suggestion: Dual `_INBOX`/`_inbox` subscriptions widen inbox surface; hub subscribes uppercase only**
`deploy/nats/auth.conf:44,52,65,71,89,107`
Audit nats.py version — if ≥2.2, remove `_inbox.*` entries.
Confidence: 68%

**question: `_configured` check-then-set is not atomic in async context**
`src/lyra/core/hub/middleware/middleware_pool.py:104-110`
Two concurrent coroutines can both pass `not pool._configured`. Currently safe only because `configure_pool` is idempotent by coincidence. Document this invariant explicitly.
Confidence: 75%

---

### 🔵 Nitpicks

- `cli_pool_streaming.py:78` — `_pool_id = pool_id` alias unnecessary; `pool_id` is a parameter not a loop variable — remove it (95%)
- `discord_outbound.py:124,205` — `assert reply_msg_id is not None` silently omitted under `python -O`; use explicit `if` guard (80%)
- `discord_normalize.py:33`, `telegram_normalize.py:133` — `# REQUIRED:` comment before `is_admin` visually belongs to prior param; move back to trailing position (75%)
- `pool.py:257,280` — two `except Exception` catches lack `# noqa: BLE001 — resilient:` annotation; inconsistent with pattern this PR establishes (85%)
- `cli_pool_types.py:9-28` — `TYPE_CHECKING`-only imports make `get_type_hints(_CliPoolCore)` fail at runtime; add a comment documenting the limitation (75%)
- `auth.conf:76-83` — `llm-worker` and `monitor` lack `_INBOX.*` subscriptions; document explicitly if intentional (75%)

---

### ✅ Praise

- `bootstrap_stores.py` — `_IDENT_RE` validation is correctly applied before all interpolation; `# safe:` comments are accurate S608 justification (92%)
- BLE001 reason tag format is clean and consistent across all files (90%)

---

**Verdict: Request changes** — 10 blockers (3 correctness bugs, 2 security, 2 type gaps, 3 script issues) must be resolved before merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/code-review` (6 domain agents)

MickaelV0:
## Code Review — Pass 2 (post-fix)

34 files · ~1312 diff lines · Secret scan: clean

---

### 🚫 Blockers (8)

**issue: `acl-matrix.json` not updated — drift-check CI will fail**
`deploy/nats/acl-matrix.json:16-17` / `deploy/nats/auth.conf:16`
`auth.conf` had `$JS.API.STREAM.CREATE/UPDATE.LYRA_AUDIT` removed from hub publish (correct: stream provisioning is deploy-time admin), but `acl-matrix.json` (the SSoT) still declares them. `check-acl-authconf-drift.sh` regenerates `auth.conf` from `acl-matrix.json` and diffs — CI breaks on every run.
- **Recommended:** Remove both `$JS.API` subjects from `acl-matrix.json` hub publish + add comment explaining deploy-time provisioning rationale
- Confidence: 98%

---

**issue: `check-inbox-prefix.sh` exits 1 on this branch — CI failure**
`scripts/check-inbox-prefix.sh:26`
Three bootstrap files use raw `inbox_prefix=` (`adapter_standalone.py`, `hub_standalone.py`, `embedded_nats.py`) — intentional, pre-date the `identity_name=` API. Script has no allowlist mechanism.
- **Recommended:** Add `tools/inbox_prefix_exemptions.txt` allowlist, list `src/lyra/bootstrap/` with a tracking issue (mirrors `file_exemptions.txt` pattern)
- Alt: Migrate the 3 files to `identity_name=` (correct long-term)
- Confidence: 97%

---

**issue: `CliStreamingParser` absent from `cli_protocol.__all__` — runtime `ImportError` risk**
`src/lyra/core/cli/cli_protocol.py:21-32` / `src/lyra/core/cli/cli_streaming.py:21`
`cli_streaming` exports `CliStreamingParser` in its own `__all__`, but the facade doesn't import or re-export it. Any consumer doing `from .cli_protocol import CliStreamingParser` gets `ImportError`.
- **Recommended:** Add `from .cli_streaming import CliStreamingParser` + add to `__all__`
- Confidence: 90%

---

**issue: `assert channel is not None` not replaced — inconsistent with PR's own pattern**
`src/lyra/adapters/discord/discord_outbound.py:70`
PR replaced `assert→raise` in `should_reply` guards but left a bare `assert` in `_discord_typing_worker`. With `-O`, asserts are stripped — `channel.typing()` then raises `AttributeError` instead of a meaningful error.
- **Recommended:** `if channel is None: raise RuntimeError("typing: channel not resolved after 3 attempts")`
- Confidence: 90%

---

**issue: f-string grep in `check-inbox-prefix.sh` misses single-quote and spaced variants**
`scripts/check-inbox-prefix.sh:26`
Pattern `inbox_prefix=f"_INBOX\.` fails on `inbox_prefix=f'_INBOX.{x}'` or `inbox_prefix = f"..."`. AST walker only catches `ast.Constant`, not `ast.JoinedStr` (f-strings), so no fallback.
- **Recommended:** `inbox_prefix\s*=\s*f["']_INBOX\.`
- Confidence: 90%

---

**issue: Python AST walker uses `open()` without `encoding=` — crashes on non-UTF-8 CI**
`scripts/check-inbox-prefix.sh:40`
`open()` without `encoding=` falls back to `locale.getpreferredencoding()`. On Alpine/minimal CI with `LANG=C`, this is ASCII — UTF-8 comments raise `UnicodeDecodeError` not caught by `except SyntaxError`.
- **Recommended:** `open(sys.argv[1], encoding="utf-8", errors="replace")`
- Alt: `except (SyntaxError, UnicodeDecodeError): sys.exit(0)`
- Confidence: 85%

---

**issue: `turn_count` incremented outside `entry._lock` in `send_streaming()`**
`src/lyra/core/cli/cli_pool_streaming.py:113`
Non-streaming `send()` increments `turn_count` inside `entry._lock`. In `send_streaming()`, lock is released before `turn_count += 1` and `last_activity = time.time()`. The event loop can run between lock-release and those writes, making the stale-resume guard (`entry.turn_count == 0`) unreliable.
- **Recommended:** Move both writes inside `async with entry._lock`, before returning the iterator
- Confidence: 80%

---

**issue: `_reset` closure can reset a different entry than it was created for**
`src/lyra/core/cli/cli_pool_streaming.py:78-79`
`_reset` is passed to `StreamingIterator` and may fire long after `send_streaming` returns. The original entry may have been evicted; a new one spawned for the same `pool_id`. `reset(pool_id)` kills the unrelated new entry.
- **Recommended:** `async def _reset() -> None: if self._entries.get(pool_id) is entry: await cast(_CliPoolCore, self).reset(pool_id)`
- Confidence: 75%

---

### ⚠️ Suggestions (non-blocking, 7)

| # | Finding | File | C |
|---|---------|------|---|
| 1 | `IndexError` in `agent_refiner_stages.py` catch is dead code — pre-guard ensures split never raises it; use `except (json.JSONDecodeError, ValueError)` | `agent_refiner_stages.py:96` | 92% |
| 2 | `cli_pool_worker.py` imports from facade instead of `cli_protocol_types` — residual transitive cycle masked by partial module cache | `cli_pool_worker.py:24` | 88% |
| 3 | `_configured` check-then-set comment should explain why safe: "single hub task — only one pipeline invocation at a time; concurrent refactor needs a lock" | `middleware_pool.py:103-111` | 88% |
| 4 | `_read_stderr_snippet` in `cli_protocol.__all__` exposes private name as public API — remove from `__all__` | `cli_protocol.py:26` | 85% |
| 5 | `TraceLogRecord` Protocol purely cosmetic — no callers annotate params with it; `JsonFormatter` uses `getattr` bypassing Protocol entirely | `trace.py:31,94` | 82% |
| 6 | `idx_sql` regex not anchored — tighten to validate full structure through `ON\b` | `bootstrap_stores.py:136` | 78% |
| 7 | `check-inbox-prefix.sh` test-file exclusion too narrow — add `-not -name "test_*.py" -not -name "conftest.py"` | `check-inbox-prefix.sh:21` | 75% |

---

### 💬 Nitpicks

- `_SESSION_ID_RE` in `cli_protocol.__all__` — private alias promoted to documented API; migrate `cli_pool_session.py` to `SESSION_ID_RE`. C: 80%
- `NATS_TMPDIR` naming inconsistent with sibling `check-acl-matrix-spec.sh` (still `TMPDIR_LOCAL`). C: 88%
- `normalize_authconf` sed `[^"]*$` fails on comments containing `"` — use `s/[[:space:]]*#.*$//`. C: 82%
- `_CliPoolCore` get_type_hints comment understates risk — any decorator/Protocol-check triggers NameError, not just explicit `get_type_hints(_CliPoolCore)`. C: 65%
- NATS monitor has symmetric pub+sub on `lyra.monitor.>` — compromised identity can inject fabricated telemetry. C: 65%
- `file_exemptions.txt` `wiring_helpers.py` entry uses ADR ref instead of issue number. C: 70%

---

### ✅ Praise

- `cli_protocol.py` re-export facade: 29 lines, clean `__all__`, zero `noqa` — structurally sound backward-compat solution. C: 92%
- `_CliPoolCore` Protocol: exactly the 4 methods needed across both mixin consumers — no over/under-declaration. C: 90%
- BLE001 noqa on TurnStore catches: `log.exception` called before — not a silent swallow. C: 95%
- `log.warning` for plugin load failure: correct level upgrade — operators need visibility when commands are missing. C: 95%

---

### Verdict: 🔴 Request changes

**8 blockers** — 2 immediate CI failures (acl-matrix.json drift, check-inbox-prefix.sh exit 1), 2 runtime correctness (CliStreamingParser ImportError, assert with -O), 4 async/script correctness (turn_count, _reset closure, f-string grep, encoding).

*Agents: security-auditor · architect · backend-dev ×2 · devops*

MickaelV0:
## Code Review — Pass 3 (final verification)

43 files · All hooks pass (lint, typecheck, file-length, folder-size, import-layers, trufflehog)

### ✅ All pass-2 blockers confirmed resolved

| Area | Status |
|------|--------|
| `acl-matrix.json` / `auth.conf` drift | ✅ Fixed — $JS.API.STREAM entries removed from both |
| `check-inbox-prefix.sh` CI failure | ✅ Fixed — bootstrap files migrated to `identity_name=` |
| `CliStreamingParser` missing from `cli_protocol.__all__` | ✅ Fixed |
| `assert channel is not None` → explicit raise | ✅ Fixed (RuntimeError) |
| f-string grep pattern coverage | ✅ Fixed |
| AST walker `encoding=utf-8` | ✅ Fixed |
| `turn_count` outside lock | ✅ Fixed — inside `entry._lock` with guard updated to `== 1` |
| `_reset()` closure identity check | ✅ Fixed — `is entry` guard added |

### ✅ Pass-3 non-blockers also fixed

- `raise AssertionError` → `RuntimeError` at 2 remaining sites in `discord_outbound.py`
- Final-chunk send failure now re-raises after logging (was silently swallowed)
- Stale-resume `turn_count` asymmetry documented inline on both paths
- `cli_pool_entry.py` pyright suppress annotated with rationale

### Verdict: 🟢 Approve

*Agents: security-auditor · architect · backend-dev ×2 · devops (×3 passes)*

PR #975, opened 2026-04-27
refactor(nats): make PLATFORM_META_ALLOWLIST injectable in sanitize_platform_meta

### LINKED ISSUE #961: refactor(nats): make PLATFORM_META_ALLOWLIST injectable in _sanitize.py (ADR-059 shared V3)
ADR-059 shared V3: packages/roxabi-nats/src/roxabi_nats/_sanitize.py has PLATFORM_META_ALLOWLIST hardcoded with Discord/Telegram field names. Should be injectable config or sourced from roxabi-contracts, not hardcoded in the transport SDK.

MickaelV0:
## Summary
- Add optional keyword-only `allowlist` parameter to `sanitize_platform_meta` so callers can inject a platform-specific set instead of relying on the hardcoded default
- Default behaviour is unchanged — `None` falls back to `PLATFORM_META_ALLOWLIST`, preserving full backward compatibility
- Add `TestInjectableAllowlist` class (4 tests) covering custom allowlist, `None` fallback, empty allowlist, and underscore-strip invariant

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #961: refactor(nats): make PLATFORM_META_ALLOWLIST injectable in _sanitize.py (ADR-059 shared V3) | Open |
| Implementation | 1 commit on `feat/961-platform-meta-allowlist-injectable` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (4 new) | Passed |

## Test Plan
- [ ] `cd packages/roxabi-nats && uv run pytest tests/test_sanitize.py -v` — 18 tests pass
- [ ] Existing callers unaffected (signature is backward compatible via keyword-only optional param)
- [ ] Inject a custom `frozenset` and verify only those keys survive

Closes #961

---

### DIFF
--- a/packages/roxabi-nats/src/roxabi_nats/_sanitize.py
+++ b/packages/roxabi-nats/src/roxabi_nats/_sanitize.py
@@ -40,11 +40,15 @@ def _cap_value(val: str | int | bool) -> str | int | bool:
     return val
 
 
-def sanitize_platform_meta(meta: dict[str, Any]) -> dict[str, Any]:
+def sanitize_platform_meta(
+    meta: dict[str, Any],
+    *,
+    allowlist: frozenset[str] | None = None,
+) -> dict[str, Any]:
     """Filter platform_meta to allowlisted keys and safe scalar values.
 
     Strips:
-    - Keys not in PLATFORM_META_ALLOWLIST
+    - Keys not in the allowlist (defaults to ``PLATFORM_META_ALLOWLIST``)
     - Keys with leading underscore (internal-only, e.g. _session_update_fn)
     - Values that are not scalar (``str | int | bool``) — avoids the
       ``str({deep dict})`` memory-amplification path since the only
@@ -55,15 +59,19 @@ def sanitize_platform_meta(meta: dict[str, Any]) -> dict[str, Any]:
 
     Stripped keys and dropped values are logged at DEBUG (key names only,
     never value content — avoids log-amplification from crafted messages).
+
+    Pass ``allowlist`` to inject a platform-specific set instead of the
+    hardcoded default (e.g. sourced from roxabi-contracts at call time).
     """
+    active_allowlist = allowlist if allowlist is not None else PLATFORM_META_ALLOWLIST
     stripped = [
-        k for k in meta if k not in PLATFORM_META_ALLOWLIST or k.startswith("_")
+        k for k in meta if k not in active_allowlist or k.startswith("_")
     ]
     if stripped:
         log.debug("platform_meta: stripped keys %s", stripped)
     result: dict[str, Any] = {}
     for k, v in meta.items():
-        if k not in PLATFORM_META_ALLOWLIST or k.startswith("_"):
+        if k not in active_allowlist or k.startswith("_"):
             continue
         if not isinstance(v, (str, int, bool)):
             log.debug(
--- a/packages/roxabi-nats/src/roxabi_nats/_sanitize.py
+++ b/packages/roxabi-nats/src/roxabi_nats/_sanitize.py
@@ -63,6 +63,7 @@ def sanitize_platform_meta(
     Pass ``allowlist`` to inject a platform-specific set instead of the
     hardcoded default (e.g. sourced from roxabi-contracts at call time).
     """
+    # empty frozenset is valid (block-all); only None falls back to default
     active_allowlist = allowlist if allowlist is not None else PLATFORM_META_ALLOWLIST
     stripped = [
         k for k in meta if k not in active_allowlist or k.startswith("_")
--- a/packages/roxabi-nats/tests/test_sanitize.py
+++ b/packages/roxabi-nats/tests/test_sanitize.py
@@ -102,10 +102,13 @@ class TestEmptyAndAllowlistShape:
 
 class TestInjectableAllowlist:
     def test_custom_allowlist_used_when_provided(self) -> None:
-        custom = frozenset({"custom_field", "other_field"})
-        meta = {"custom_field": "keep", "guild_id": "strip", "other_field": 1}
+        # "custom_only" is NOT on PLATFORM_META_ALLOWLIST — proves the custom set is
+        # active. "guild_id" IS on PLATFORM_META_ALLOWLIST — proves the default is not.
+        custom = frozenset({"custom_only", "other_field"})
+        meta = {"custom_only": "keep", "guild_id": "default-only", "other_field": 1}
         result = sanitize_platform_meta(meta, allowlist=custom)
-        assert result == {"custom_field": "keep", "other_field": 1}
+        assert result == {"custom_only": "keep", "other_field": 1}
+        assert "guild_id" not in result
 
     def test_default_allowlist_used_when_none(self) -> None:
         meta = {"guild_id": "123", "unknown": "x"}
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Code Review

### Blockers

**issue**: `test_underscore_keys_still_stripped_with_custom_allowlist` doesn't isolate the underscore guard — test passes even if the guard were removed
  `packages/roxabi-nats/tests/test_sanitize.py:118`
  -- tester
  Root cause: `_custom_field` is not in `custom`, so it's rejected by the allowlist check alone; the `startswith("_")` branch is never the operative filter.
  Solutions:
    1. Add `"_custom_field"` to `custom`, then assert it is still absent — that makes the underscore guard the only active filter. (recommended)
    2. Split into two distinct tests isolating each rejection branch.
  Confidence: 95%

---

**issue**: `test_custom_allowlist_used_when_provided` cannot distinguish allowlist filtering from default-allowlist use — test would pass if the default allowlist were used instead
  `packages/roxabi-nats/tests/test_sanitize.py:106`
  -- tester
  Root cause: `guild_id` is on `PLATFORM_META_ALLOWLIST`, so its absence in the result is consistent with both the custom allowlist and the default. A regression where `active_allowlist` fell back to the default would still make this test green.
  Solutions:
    1. Use a key that IS on `PLATFORM_META_ALLOWLIST` but NOT on `custom` and assert it is absent; use a key on `custom` but NOT on `PLATFORM_META_ALLOWLIST` and assert it is present. (recommended)
    2. Add an explicit assertion: `assert "guild_id" not in result` to pin the test to custom-allowlist semantics.
  Confidence: 90%

---

**issue**: underscore guard order is ambiguous — injected allowlist containing `_`-prefixed keys gives misleading semantics
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:65`
  -- backend-dev
  Root cause: `k not in active_allowlist or k.startswith("_")` makes the underscore check look like a peer condition on the allowlist. If a caller passes `allowlist=frozenset({"_foo"})`, the key is admitted past the allowlist gate then stripped by the underscore guard — surprising behavior. The underscore invariant should read as unconditional.
  Solutions:
    1. Reorder: `if k.startswith("_") or k not in active_allowlist` — semantics-preserving, intent explicit. (recommended)
    2. Add a runtime assertion on the injected allowlist: `assert not any(k.startswith("_") for k in (allowlist or frozenset()))`.
  Confidence: 80%

---

### Suggestions

**suggestion(non-blocking)**: Add inline comment clarifying that `frozenset()` is a valid input (block-all) and only `None` falls back to the default
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:62`
  -- backend-dev
  Root cause: `allowlist or PLATFORM_META_ALLOWLIST` would silently replace an empty frozenset with the default (a security regression). The `is not None` guard is correct but a future maintainer may simplify it incorrectly.
  Solution: `# empty frozenset is valid (block-all); only None falls back to default`
  Confidence: 85%

**suggestion(non-blocking)**: Missing test — custom allowlist + non-scalar value
  `packages/roxabi-nats/tests/test_sanitize.py:103`
  -- tester
  Root cause: `TestScalarValidation` covers the scalar-drop path against the default allowlist only; a refactor moving the scalar check before the allowlist check would be invisible.
  Solution: `custom = frozenset({"custom_field"}); meta = {"custom_field": [1, 2, 3]}; assert sanitize_platform_meta(meta, allowlist=custom) == {}`
  Confidence: 85%

**suggestion(non-blocking)**: Docstring omits that `_`-prefix stripping is unconditional regardless of injected allowlist
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:43`
  -- backend-dev
  Solution: Add `Keys with a leading underscore are always stripped, even if present in a caller-supplied allowlist.`
  Confidence: 75%

**suggestion(non-blocking)**: Missing test — custom allowlist + string over `MAX_META_VALUE_LEN`
  `packages/roxabi-nats/tests/test_sanitize.py:103`
  -- tester
  Root cause: `_cap_value` is exercised only via default-allowlist paths in `TestLengthCap`.
  Solution: `custom = frozenset({"custom_field"}); val = "x" * (MAX_META_VALUE_LEN + 100); assert len(sanitize_platform_meta({"custom_field": val}, allowlist=custom)["custom_field"]) == MAX_META_VALUE_LEN`
  Confidence: 75%

**thought**: Open `frozenset[str]` parameter gives no type-level guarantee the injected set originates from a trusted constant — design concern for when a call site wires this to config/runtime data
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:44`
  -- security-auditor
  Note: zero production call sites currently pass the `allowlist` parameter (per `test_sanitize_single_entry.py:EXPECTED_CALL_SITES = {}`), so not exploitable today.
  Confidence: 65%

---

### Summary

| Category | Count |
|----------|-------|
| Blockers (issue) | 3 |
| Suggestions (non-blocking) | 4 |
| Thoughts | 1 |

**Verdict: Request changes** — 2 test correctness issues (tests don't prove what they claim) + 1 code clarity issue (underscore guard ordering). Production logic is correct; fixes are mechanical and low-risk.

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 4 findings
**Applied via 1b1:** 0 findings
**Deferred (issues created):** 1 finding → #983
**Skipped:** 2 findings (verifier-rejected: guard order style, docstring)
**Failed:** 0

### Auto-Applied
- [applied] issue: `test_underscore_keys_still_stripped_with_custom_allowlist` — add `_custom_field` to custom so underscore guard is the operative filter (95%)
- [applied] issue: `test_custom_allowlist_used_when_provided` — use key not on PLATFORM_META_ALLOWLIST; assert guild_id absent (90%)
- [applied] suggestion: add `test_non_scalar_dropped_with_custom_allowlist` (85%)
- [applied] suggestion: comment clarifying `frozenset()` is block-all; only `None` falls back to default (85%)

### Deferred
- suggestion: derive `PLATFORM_META_ALLOWLIST` from roxabi-contracts PlatformMeta models → #983

### Skipped (verifier-rejected)
- issue: underscore guard order rewrite — semantics identical, tester verified at 85% not needed
- suggestion: docstring note on unconditional underscore strip — tester verified at 82% not needed

MickaelV0:
## Code Review (iteration 2)

### Pre-existing Issues (not introduced by this PR — deferred)

**issue**: `_cap_value` logs truncation without key name context — context-free log at the trust boundary
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:34`
  -- backend-dev
  Root cause: `_cap_value` is a pure helper that carries a side-effect (log) with no access to the key name. Every other log line in this module includes the key name (`%r`); truncation is the exception. Pre-existing — `_cap_value` is unchanged by this PR.
  Solution: Move the truncation log into `sanitize_platform_meta`'s loop alongside the non-scalar drop log, strip `_cap_value` back to a pure function.
  Confidence: 88%

---

**suggestion(non-blocking)**: Add runtime `isinstance(allowlist, frozenset)` guard — mutable `set` accepted silently
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:46`
  -- security-auditor
  Confidence: 72%

**suggestion(non-blocking)**: `isinstance(v, (str, int, bool))` — `bool` subclasses `int`; add comment to document intentional inclusion
  `packages/roxabi-nats/src/roxabi_nats/_sanitize.py:77`
  -- backend-dev
  Confidence: 95%

**suggestion(non-blocking)**: `test_non_scalar_dropped_with_custom_allowlist` — add parallel scalar call to prove drop is type-based, not key-based
  `packages/roxabi-nats/tests/test_sanitize.py:130`
  -- tester
  Confidence: 80%

---

### Summary

| Category | Count |
|----------|-------|
| Pre-existing issues (deferred) | 1 |
| Suggestions (non-blocking) | 3 |

**Verdict: Approve** — no new issues introduced by this PR. The one `issue:` finding pre-dates this change (`_cap_value` is untouched). All suggestion-level findings are non-blocking style/clarity items.

PR #1011, opened 2026-04-29
feat(nats): add status/created_at lifecycle fields to acl-matrix identities

### LINKED ISSUE #1001: acl-matrix.json: add status active|retired field to identities
## Context

Identities in `deploy/nats/acl-matrix.json` have no lifecycle field. Retiring an identity currently has no tooling path — the entry is either left in place (credentials still generated on every `--regenerate`) or deleted (no audit trail, dangling seed file).

## Problem

A retired identity with no `status` field keeps accumulating live credentials and ACL grants on every `gen-nkeys.sh --regenerate` run. Security surface grows silently.

## Proposed solution

1. Add `"status": "active"` to all current identity blocks.
2. `gen-nkeys.sh` `render_auth_conf`: skip identities where `status == "retired"` — no `users[]` block emitted.
3. `gen-nkeys.sh` `generate_nkeys`: skip `generate_nkey` call for retired identities — no seed created/regenerated.
4. CI check: reject any identity with `status == "retired"` and no `retired_at` date field.
5. Document retirement process: set `status: retired`, add `retired_at: YYYY-MM-DD`, run `gen-nkeys.sh --regen-authconf`, reload NATS, verify no reconnect from retired identity.

## Result

Retiring an identity is a documented, tooling-enforced step. No dangling credentials, no ACL grants for dead services, full audit trail in the JSON.

MickaelV0:
## Summary

- Add `status` (active|retired) and `created_at` lifecycle fields to all 9 identities in `deploy/nats/acl-matrix.json`, enabling tooling-enforced retirement with full audit trail
- `gen-nkeys.sh` now filters out retired identities from key generation and `auth.conf` rendering; replaces hardcoded key-gen block with `IDENTITIES[]` loop; fixes `--emit-merged-authconf` to derive lists by owner; warns when a retired identity has a seed on disk
- New `scripts/check-acl-matrix-retired.sh` validates `status`, `created_at`, `retired_at` format, and rejects retired identities still referenced in `request_reply_flows`
- `scripts/check-acl-matrix-spec.sh` now derives its `IDENTITIES` column list from active JSON entries (no hardcoded array) and gains a `--update` flag to auto-regenerate the sentinel block in `706-per-role-nkeys-acls-spec.mdx`
- New `docs/ops/nats-identity-retirement.md` runbook covers the full retirement procedure

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1001: acl-matrix.json: add status active\|retired field to identities | Open |
| Frame | [1001-acl-matrix-status-field-frame.mdx](artifacts/frames/1001-acl-matrix-status-field-frame.mdx) | Present |
| Spec | [1001-acl-matrix-status-field-spec.mdx](artifacts/specs/1001-acl-matrix-status-field-spec.mdx) | Present |
| Plan | [1001-acl-matrix-status-field-plan.mdx](artifacts/plans/1001-acl-matrix-status-field-plan.mdx) | Present |
| Implementation | 8 commits on `feat/1001-acl-matrix-status-field` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (3236 pass) | Passed |

## Test Plan

- [ ] `bash scripts/check-acl-matrix-retired.sh` → `ok — acl-matrix lifecycle fields valid`
- [ ] `bash scripts/check-acl-matrix-spec.sh` → exit 0 (no drift)
- [ ] `bash deploy/nats/gen-nkeys.sh --template-only` → 9 active identity `permissions` blocks, no retired identity present
- [ ] Set an identity to `"status": "retired"` without `"retired_at"` → `check-acl-matrix-retired.sh` exits 1
- [ ] Set an identity to `"status": "retired"` with `"retired_at": "28-04-2026"` (wrong format) → exits 1
- [ ] `bash scripts/check-acl-matrix-spec.sh --update` → rewrites sentinel block, subsequent run exits 0

Closes #1001

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,9 @@ jobs:
       - name: Check ACL matrix spec drift
         run: bash scripts/check-acl-matrix-spec.sh
 
+      - name: Check ACL matrix lifecycle fields
+        run: bash scripts/check-acl-matrix-retired.sh
+
       - name: Check request-reply flows
         run: bash scripts/check-request-reply-flows.sh
 
--- a/artifacts/analyses/1001-acl-matrix-review-findings-consensus.mdx
+++ b/artifacts/analyses/1001-acl-matrix-review-findings-consensus.mdx
@@ -1,0 +1,75 @@
+---
+title: "#1001 acl-matrix review findings — Expert Consensus"
+issue: 1001
+status: consensus-reached
+date: 2026-04-29
+panel: architect, devops, product-lead
+confidence: high
+---
+
+## Problem
+
+Four remaining code-review findings for #1001 (acl-matrix.json lifecycle fields) required a long-term decision: whether to apply structural changes or lighter-weight annotations, and whether to reorder CI steps or accept current ordering.
+
+## Panel
+
+| Agent | Focus |
+|---|---|
+| architect | arch soundness, maintainability, future-proofing |
+| devops | ops impact, CI reliability, bash idioms |
+| product-lead | operator UX, docs quality, runbook usability |
+
+## Consensus
+
+### F3 — `retired_at` format-check short-circuit (unanimous B, avg 87%)
+
+**Decision:** Add a comment clarifying the intent of the `-z "$retired"` guard — no structural refactor.
+
+**Rationale:** Both line 21 and line 22 use the same `[[ -z "$x" || "$x" =~ $RE ]]` idiom; rewriting only line 22 to an if/fi block would break visual symmetry with line 21 and add 3 lines for zero behavioral change. A comment is the minimum-cost fix.
+
+**Rejected (A):** Structural refactor introduces diff noise, breaks symmetry, adds indentation for no runtime benefit.
+
+---
+
+### F4 — CI step ordering (unanimous B, avg 86%)
+
+**Decision:** Leave current order — "Check ACL matrix spec drift" before "Check ACL matrix lifecycle fields".
+
+**Rationale:** The current order follows a coherent data-flow narrative: spec table correctness → lifecycle field validity → flow references → authconf drift. Swapping produces no measurable diagnostic improvement (both steps fail independently with clear step names in the CI log). "More fundamental" is not user-observable in a flat CI log.
+
+**Rejected (A):** No runtime benefit; would break an implicit ordering rationale and generate a noise commit.
+
+---
+
+### F5 — jq O(n) forks in `check-acl-matrix-retired.sh` (unanimous B, avg 92%)
+
+**Decision:** Accept the current approach (3–5 jq calls per identity). Revisit only if the matrix grows beyond ~100 identities or the script enters a hot path.
+
+**Rationale:** At ≤20 identities the overhead is 50–200ms, unmeasurable against the pytest bottleneck in the same CI job. The current pattern (one jq call per field per identity) maps directly to error messages and is easy to audit. A single-pass jq refactor would embed complex conditional validation in jq syntax, increasing cognitive surface for ops maintainers.
+
+**Rejected (A):** Premature optimization. Identity counts are bounded by human decisions and grow slowly.
+
+---
+
+### F6 — Runbook annotation (unanimous A, avg 90%)
+
+**Decision:** Add an inline one-liner comment at each `check-acl-matrix-retired.sh` call site in the runbook, explaining it validates all identities (not just the one being retired/added).
+
+**Rationale:** Runbooks are read top-to-bottom under operational pressure. An inline comment at the call site eliminates the most common misread ("this only checks the identity I just touched") without requiring the reader to scroll to the CI checks table. Cross-references are appropriate for architecture docs, not for procedural steps.
+
+**Rejected (B):** Cross-reference requires navigation during a time-sensitive operation; the table is excellent reference material but not a substitute for inline context at the call site.
+
+## Dissent
+
+None — unanimous on all four findings.
+
+## Implementation Notes
+
+- F3: add `# -z guard intentional: empty already caught above` on line 22
+- F4: no change to ci.yml
+- F5: no change to check-acl-matrix-retired.sh
+- F6: add inline annotation at both call sites in nats-identity-retirement.md (retire step 3, add-identity step 2)
+
+## Next
+
+These decisions feed directly into the active `/fix` pipeline for PR #1011.
--- a/artifacts/specs/706-per-role-nkeys-acls-spec.mdx
+++ b/artifacts/specs/706-per-role-nkeys-acls-spec.mdx
@@ -14,13 +14,13 @@ and is verified by `scripts/check-acl-matrix-spec.sh` on every CI run.
 ## ACL Matrix
 
 <!-- acl-matrix:begin -->
-| Subject | hub | telegram-adapter | discord-adapter | voice-tts | voice-stt | llm-worker | image-worker | clipool-worker | monitor |
+| Subject | hub | telegram-adapter | discord-adapter | voice-tts | voice-stt | llm-worker | image-worker | monitor | clipool-worker |
 |---|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
 | `lyra.inbound.telegram.>` | SUB | PUB | — | — | — | — | — | — | — |
 | `lyra.inbound.discord.>` | SUB | — | PUB | — | — | — | — | — | — |
 | `lyra.outbound.telegram.>` | PUB | SUB | — | — | — | — | — | — | — |
 | `lyra.outbound.discord.>` | PUB | — | SUB | — | — | — | — | — | — |
-| `lyra.system.ready` [^ready] | SUB | PUB | PUB | PUB | PUB | — | — | PUB | — |
+| `lyra.system.ready` [^ready] | SUB | PUB | PUB | PUB | PUB | — | — | — | PUB |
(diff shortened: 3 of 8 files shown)


MickaelV0:
## Code Review — #1001 acl-matrix.json: add status active|retired field

### Blockers

**issue: `set -euo pipefail` + `((errors++))` causes silent exit on first error**
  `scripts/check-acl-matrix-retired.sh:15`
  -- devops
  Root cause: `((errors++))` when `errors=0` evaluates the expression to 0 (old value before increment), which is arithmetically false — bash returns exit code 1. With `set -e` active, the script dies on the very first missing field instead of accumulating all errors. The `exit 1` at line 32 is dead code in the common failure case.
  Solutions:
    1. Replace all `((errors++))` with `errors=$((errors + 1))` — assignment always exits 0 (recommended)
    2. Use `(( errors++ )) || true` throughout (safe but verbose)
  Confidence: 97%

---

**issue: retired-seed warn block fires unconditionally, including for `--template-only`**
  `deploy/nats/gen-nkeys.sh:250`
  -- product-lead
  Root cause: The warn block runs immediately after `load_matrix()` before any mode-specific exit. `--template-only` is a no-root, no-filesystem stdout operation; emitting seed-on-disk warnings to stderr during config diffing is unexpected operator noise.
  Solutions:
    1. Move the warn block inside the `--regen-authconf` and default (full regenerate) code paths only (recommended)
    2. Gate with `[[ "${TEMPLATE_ONLY}" == "false" && "${EMIT_MERGED_AUTHCONF}" == "false" ]]`
  Confidence: 75%

---

**suggestion: `load_matrix()` validates field presence but not allowed values for `status`**
  `deploy/nats/gen-nkeys.sh:100`
  -- devops
  Root cause: `jq has($f)` confirms the field exists but not that `status ∈ {active, retired}`. A typo like `"Retired"` (wrong case) passes validation silently and skips the `retired_at` / flow-ref checks in the CI script.
  Solutions:
    1. Add allowlist check after `has()`: `jq -e --arg n "$name" '.identities[$n].status | IN("active","retired")' "$MATRIX_JSON"` (recommended)
    2. Add the same check to `check-acl-matrix-retired.sh`
  Confidence: 88%

**suggestion: retired-seed warn block calls `jq` once per identity for `.status` — redundant re-read**
  `deploy/nats/gen-nkeys.sh:251`
  -- devops
  Root cause: `load_matrix()` already parsed all identities. The warn block then forks a separate `jq` subprocess per key to re-read `.status`. Suggestion: populate an `ALL_STATUS[]` associative array during `load_matrix()` and iterate that directly in the warn block.
  Solutions:
    1. Cache all statuses into `ALL_STATUS[]` during `load_matrix()` loop, iterate that in warn block (recommended)
    2. Accept current approach — at ≤20 identities the overhead is negligible
  Confidence: 90%

**suggestion: header comment in gen-nkeys.sh hardcodes "9" and lists identity names — will silently drift**
  `deploy/nats/gen-nkeys.sh:7`
  -- product-lead
  Root cause: Line reads "Creates 9 user nkey seeds: hub, telegram-adapter, ...". Adding or retiring an identity requires manually updating this comment, which contradicts criterion 10 (no manual edits required).
  Solutions:
    1. Replace with "Creates user nkey seeds for all active identities in acl-matrix.json" (recommended)
    2. Leave as-is with acknowledgment it is documentation-only
  Confidence: 82%

**suggestion: `--update` flag has no CI guard — a future maintainer could silently overwrite the spec artifact**
  `scripts/check-acl-matrix-spec.sh:178`
  -- devops
  Root cause: `cp "${NATS_TMPDIR}/spec_updated.txt" "$SPEC"` overwrites the committed artifact. CI does not pass `--update` today, but there is no guard preventing it. The change would be lost at runner teardown with no error.
  Solutions:
    1. Add `if [ "${CI:-}" = "true" ]; then echo "::error::--update not allowed in CI"; exit 1; fi` at the top of the UPDATE branch (recommended)
    2. Add a comment above the block: `# Local-only flag: never pass --update in CI`
  Confidence: 72%

**suggestion: `retired_at` format check short-circuits silently when field is absent**
  `scripts/check-acl-matrix-retired.sh:22`
  -- devops
  Root cause: `[[ -z "$retired" || "$retired" =~ $DATE_RE ]]` is true when `retired` is empty (so format is never validated in that case). After fixing the `((errors++))` bug, this means a retired identity with a missing `retired_at` triggers only the "missing retired_at" error — the format check on line 22 is a no-op. Intent is correct but could be made explicit.
  Solutions:
    1. Wrap the format check: `if [ -n "$retired" ]; then [[ "$retired" =~ $DATE_RE ]] || { echo "ERROR..."; errors=$((errors+1)); }; fi`
    2. Add a comment: `# -z guard is intentional: empty case already caught above`
  Confidence: 78%

**suggestion: CI step ordering — lifecycle check should precede spec drift check**
  `.github/workflows/ci.yml:66`
  -- product-lead
  Root cause: `check-acl-matrix-spec.sh` runs before `check-acl-matrix-retired.sh`. When an operator forgets `retired_at`, the lifecycle check fires — but the spec drift check passes first (it only considers active identities). Swapping the order gives a clearer first failure message.
  Solutions:
    1. Move "Check ACL matrix lifecycle fields" before "Check ACL matrix spec drift" (recommended)
    2. Leave as-is — both checks run and both must pass regardless of order
  Confidence: 68%

**thought: any future script iterating `keys_unsorted[]` will silently include retired identities**
  `deploy/nats/gen-nkeys.sh:248` / `scripts/check-acl-matrix-retired.sh:27`
  -- architect
  Root cause: The canonical active-only filter `to_entries[] | select(.value.status != "retired") | .key` is not documented as the required pattern. New scripts may copy the simpler `keys_unsorted[]` and silently process retired identities.
  Solutions:
    1. Add a comment near the top of `acl-matrix.json` documenting the canonical jq filter for active-only iteration
    2. Add a comment in `check-acl-matrix-spec.sh` and `gen-nkeys.sh` where `keys_unsorted[]` is intentionally used (all-identities scope) vs the active-only filter
  Confidence: 70%

**nitpick: `check-acl-matrix-retired.sh` makes 3–5 `jq` subprocess calls per identity**
  `scripts/check-acl-matrix-retired.sh:12`
  -- devops / product-lead
  Root cause: ~36–50 jq forks for 9 identities. Not a correctness issue at current scale; becomes perceptible if matrix grows significantly (planned `--retire` automation milestone).
  Solutions:
    1. Single `jq` pass emitting a JSON array of per-identity error objects, iterate in bash
    2. Accept current approach — ≤20 identities is within acceptable CI latency
  Confidence: 85%

**nitpick: runbook does not annotate what `check-acl-matrix-retired.sh` validates inline**
  `docs/ops/nats-identity-retirement.md:30`
  -- product-lead
  Root cause: The add-identity flow calls `bash scripts/check-acl-matrix-retired.sh` without explaining it also validates `status` and `created_at` on the new entry. Operators may skip it thinking it only applies to retired identities.
  Solutions:
    1. Add inline annotation: `# validates status, created_at presence and format for all identities`
    2. Cross-reference to the CI checks table further down the doc
  Confidence: 65%

---

### Summary

| Category | Count |
|---|---|
| Blockers | 2 |
| Suggestions (non-blocking) | 7 |
| Nitpicks | 2 |
| Praise | — |

**Verdict: Request changes** — 2 blocking issues must be fixed before merge.

Critical: the `((errors++))` + `set -e` interaction means `check-acl-matrix-retired.sh` exits on the first validation failure instead of reporting all errors — the spec's accumulate-and-report behavior is not implemented. The warn block placement means `--template-only` callers receive unexpected stderr output.

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 4 findings
**Applied via 1b1:** 3 findings
**Skipped (consensus B):** 3 findings (F4 CI ordering, F5 jq forks, thought: keys_unsorted doc)
**Deferred:** 0

### Auto-Applied
- [applied] issue(blocking): `((errors++))` + `set -e` silent exit → `errors=$((errors+1))` (`check-acl-matrix-retired.sh`) — C=97%
- [applied] suggestion: `load_matrix()` validates field presence but not allowed values for `status` → added `IN("active","retired")` allowlist check (`gen-nkeys.sh`) — C=97%
- [applied] suggestion: header comment hardcodes identity count/names → replaced with dynamic reference (`gen-nkeys.sh`) — C=99%
- [applied] suggestion: redundant `jq` per identity in warn block → rewritten to single `jq` call returning retired keys (`gen-nkeys.sh`) — C=90%

### Applied (1b1)
- [applied] issue(blocking): warn block fires for `--template-only` → moved block after `--fix-perms` exit, runs only for regen/default modes (`gen-nkeys.sh`) — C=75%
- [applied] suggestion: `--update` has no CI guard → added `CI` env check with `::error::` annotation (`check-acl-matrix-spec.sh`) — C=72%
- [applied] suggestion: `retired_at` format check intent unclear → added comment `# -z guard intentional: empty already caught above` (`check-acl-matrix-retired.sh`) — C=78%
- [applied] nitpick: runbook call sites don't explain scope → added inline annotations at both call sites (`nats-identity-retirement.md`) — C=90%

### Skipped (expert consensus B — no change warranted)
- F4: CI step ordering — current order follows coherent data-flow narrative; swapping provides no diagnostic benefit
- F5: jq O(n) forks — premature optimization at ≤20 identities; revisit if matrix grows beyond ~100
- thought: `keys_unsorted[]` latent risk — added to consensus artifact for future reference

### Expert Consensus
Panel (architect + devops + product-lead) reached **unanimous consensus** on all 4 contested findings.
See `artifacts/analyses/1001-acl-matrix-review-findings-consensus.mdx` for full rationale.

Issue #1029, opened 2026-05-02
Remove error_text shim post 48h soak — WorkerError P3


MickaelV0:
Follow-up to #1016 (ADR-066). After P2 has soaked in production for 48h with `tools/check_worker_error_soak_gate.sh` returning PASS, delete the legacy `error_text` shim from:

- `src/lyra/core/messaging/events.py` (`ResultLlmEvent.error_text`)
- `src/lyra/core/processors/stream_processor.py` (P2 transitional fallback path)
- `src/lyra/llm/drivers/nats_driver.py` (population of `error_text` alongside `worker_error`)
- Any test that still asserts the shim shape

## Gate

- Soak: `tools/check_worker_error_soak_gate.sh` returns PASS over a continuous 48h window in production.
- Evidence: paste stderr breakdown + `PASS` exit 0 in the PR description.

## Refs

- ADR: `docs/architecture/adr/066-unified-worker-error-envelope-nats-reply-contracts.mdx`
- Spec: `artifacts/specs/1016-worker-error-envelope-spec.mdx` (C8 — deferred from #1016 closure)
- Origin: #1016

MickaelV0:
Closing as not planned via #1222 — see PR description for the full rationale.

**TL;DR:** the `error_text` "shim" graduated to a permanent presentation channel alongside `worker_error` (structure). Sanitization fixes #1212 and #1215 hardened `error_text` rather than routing through `worker_error.message` — the codebase voted dual-channel. The soak gate (`tools/check_worker_error_soak_gate.sh`) and its tests have been retired since they guarded a removal that no longer makes sense. Archived ADR-066 Status was updated with the dual-channel adoption note.

PR #1031, opened 2026-05-02
feat(contracts): WorkerError envelope across NATS reply contracts (#1016)

### LINKED ISSUE #1016: feat(contracts): WorkerError envelope — structured error propagation across NATS workers
## Problem

Every NATS worker boundary silently drops error context. When a worker (clipool, llm-worker, future harness) fails, the hub receives `is_error=True` with no message, and the user sees a bare `❌` or the generic fallback _"Something went wrong. Please try again."_

The original scope focused on `CliChunkEvent`. After auditing the full NATS surface, the gap is **cross-cutting**: 5 reply envelopes have 4 different error contracts, the SDK silently drops on transport failures, and there is **zero structured error telemetry** to detect regressions.

### Inconsistency surface (verified)

| Envelope | Streaming? | Error fields today |
|---|---|---|
| `LlmChunkEvent` | yes | `is_error: bool`, `error: str \| None` |
| `LlmResponse` | no | `ok: bool`, `error: str \| None` |
| `CliChunkEvent` | yes | `is_error: bool`, `event_type=Literal[…,"error"]` — **no message field** |
| `TtsResponse` / `SttResponse` | no | `ok: bool`, `error: str \| None` |
| `ImageResponse` | no | `ok: bool`, `error: str \| None`, `error_detail: str \| None` (dead — never populated) |

### SDK silent-drop sites (`roxabi-nats`)

- `NatsDriverBase._stream_gen` — timeout returns silently from generator, caller sees "exhausted"
- `NatsAdapterBase._dispatch` — malformed JSON / contract-version mismatch → log + drop, no reply
- `CircuitBreaker` — poll-based (`is_open()`), no event on open/half-open

### Driver inconsistency

`NatsLlmDriver.complete()` returns `LlmResult(error=…, retryable=…)`; `NatsLlmDriver.stream()` yields `ResultLlmEvent(is_error=True, error_text=…)`. Different timeout strings between paths.

## Proposed solution

A canonical `WorkerError` envelope used uniformly on **reply** envelopes (chunks + responses), with structured codes, a registry, and adoption telemetry.

```python
# roxabi_contracts/errors.py
class WorkerError(BaseModel):
    code: str          # machine-readable, hierarchical: "transport.timeout", "cli.auth", …
    message: str       # human-readable, surfaced to user
    retryable: bool = True
    detail: str | None = None  # internal detail for logs, not shown to user
```

### Integration scope (5 envelopes)

| Contract | Change |
|---|---|
| `CliChunkEvent` | add `worker_error: WorkerError \| None = None` |
| `LlmChunkEvent` | add `worker_error` (alongside existing `error: str`) |
| `LlmResponse` | add `worker_error` (alongside existing `ok` + `@model_validator` invariants) |
| `TtsResponse` / `SttResponse` | add `worker_error` (alongside existing `ok` + invariants) |
| `ImageResponse` | add `worker_error`; **delete** `error_detail` (dead field; consumers will use `worker_error.detail`) |
| ~~`CliControlAck`~~ | **excluded** — no error path uses it; adding speculatively forbidden |

### Code namespace (open string + registry)

Hierarchical, dot-delimited. Open string (no Pydantic Literal — would force a version bump on every new code; contradicts ADR-049's additive-only minor-bump rule). Validity is enforced by a registry shipped in P1 (Python module + Markdown table).

```
transport.{timeout, no_responders, parse, contract_mismatch, slow_consumer}
worker.{crash, validation, internal, busy}
cli.{auth, session_lost, parse}
llm.{rate_limit, context_too_long, model_unavailable}
voice.{engine_unavailable, audio_invalid}
image.{engine_unavailable, prompt_rejected}
```

### Hub side

Single `_extract_worker_error(reply) -> WorkerError | None` helper; drives circuit-breaker (`retryable`) and retry-prompt logic. Replaces hardcoded `"Something went wrong. Please try again."` fallback. **`ok` / `is_error` flags remain** as primary predicate (their `@model_validator` invariants are load-bearing); `worker_error` is additive.

### SDK side (driver-only synthesis)

`roxabi-nats` synthesizes `WorkerError(code="transport.*", …)` instead of silent drops. **Driver side only** (`_stream_gen` timeout, JSON parse failure) — `_dispatch` (worker-receive direction) has no reply subject available, so cannot synthesize. Documented as Phase 3 scope constraint.

## Phasing

| Phase | Issue | Scope |
|---|---|---|
| **P1** | #1016 | `WorkerError` model + 5-envelope additive field + canonical code registry + delete `ImageResponse.error_detail` + ship `worker_error_received_total{code,domain}` & `worker_error_populated_total{domain}` adoption counters |
| **P2** | #1016 | `clipool_worker` + `NatsLlmDriver` populate `worker_error`; hub extractor reads it; **delete short-term `error_text` shim** (gated on counter telemetry: populated counter non-zero AND legacy `error_text` counter = 0 for 48h in prod) |
| P3 | _new child issue_ | SDK driver-side synthesis (`_stream_gen` timeout / parse / no-responders → terminal `WorkerError` event); `_dispatch` documented as out-of-scope |
| P4 | _new child issue_ | Voice + image workers populate `worker_error`; remove ad-hoc patterns |
| ~~P5~~ | _struck_ | ~~Deprecate `is_error` / `ok`~~ — invariants on `@model_validator` are load-bearing; churn cost > value |
| P6 | _separate epic_ | Adapter i18n via `code` lookup |

## Prerequisites (must ship before any code)

- **ADR-NNN** — _Unified WorkerError envelope across NATS reply contracts._ Touches 2+ packages (`roxabi-contracts`, `roxabi-nats`), introduces new cross-cutting pattern, supersedes the worker→hub portion of ADR-006. Normative on the code registry.

## Coordination & rollout

- **Additive-only** per ADR-049; `CONTRACT_VERSION` stays at `"1"`.
- **Generation marker** — log `roxabi_contracts.__version__` at process startup so dashboards can identify which deploys have the field. No `contract_version` bump.
- **Field-absence = old-version signal** — `extra="ignore"` (ADR-049) handles staggered restart; no rolling-restart doc required for additive phases.
- **Cross-repo skew** — voiceCLI / imageCLI pin contracts by tag; absence is non-failure during P4. Removal phases (none scheduled in this issue) would require coordinated upgrade per ADR-049 §Release coordination.
- **Quadlet migration** — independent; proceed.
- **Known coupling for P3** — `NatsLlmDriver` predates `NatsDriverBase`; duplicate heartbeat / `_any_worker_alive` logic. Flag for whoever does P3.

## Out of scope

- OTel (#667) — `WorkerError` is the transport; OTel instruments it.
- i18n of user messages (P6 / separate epic) — hub maps `code` → locale string later.
- Retry orchestration at hub (separate concern).
- `_dispatch` synthesis (no reply subject available).
- Deprecating `is_error` / `ok` flags (Phase 5 struck — `@model_validator` invariants depend on `ok`).
- `CliControlAck` (no error path uses it).
- Migrating non-NATS error paths (HTTP, internal exceptions).

## Depends on

- #839 (port llm domain to contracts) — coordinate field additions on `LlmChunkEvent` / `LlmResponse`.
- #734 (roxabi-nats SDK v0.2.0) — contract versioning story.

## Related

- Immediate fix landed in `603c9472` (stream_processor safety net + nats_driver error_text).
- ADR-049 — roxabi-contracts as contract SSoT.
- ADR-006 — hub run-loop error/response gap (this issue extends the principle to worker→hub).

MickaelV0:
## Summary
- Adds canonical `WorkerError` model + open-string code registry to `roxabi-contracts`, with a CI sync gate keeping `KNOWN_CODES` (Python) and `error-codes.md` in lockstep.
- Adds `worker_error: WorkerError | None` to 5 reply envelopes (CliChunk/LlmChunk/LlmResponse/Tts/Stt/Image) and removes dead `ImageResponse.error_detail`.
- Producers (`clipool_worker`, `NatsLlmDriver.complete/stream`, `cli_streaming_parser`) populate `worker_error` with the canonical code namespace; hub-side `_extract_worker_error` renders `WorkerError.message`, replacing the hardcoded `"Something went wrong. Please try again."` fallback.
- Ships `tools/check_worker_error_soak_gate.sh` (POSIX bash, journald-driven) + fixtures + pytest coverage; gate condition decides when the legacy `error_text` shim can be removed (#1029, blocked-by this PR).

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1016: WorkerError envelope | Open |
| Frame | [1016-worker-error-envelope-frame.mdx](artifacts/frames/1016-worker-error-envelope-frame.mdx) | Approved (F-full) |
| Analysis | [1016-worker-error-envelope-analysis.mdx](artifacts/analyses/1016-worker-error-envelope-analysis.mdx) | Approved — Shape B |
| ADR | [066-unified-worker-error-envelope-nats-reply-contracts.mdx](docs/architecture/adr/066-unified-worker-error-envelope-nats-reply-contracts.mdx) | Accepted |
| Spec | [1016-worker-error-envelope-spec.mdx](artifacts/specs/1016-worker-error-envelope-spec.mdx) | Approved (7 criteria, 15 affordances, 3 slices) |
| Plan | [1016-worker-error-envelope-plan.mdx](artifacts/plans/1016-worker-error-envelope-plan.mdx) | 25 micro-tasks, 10 waves |
| Implementation | 3 commits on `feat/1016-worker-error-envelope` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (3398 passed, 65 skipped; 8 new test files) | Passed |

## Acceptance evidence (C1–C7)

- **C1** model + registry + CI sync — `uv run python packages/roxabi-contracts/scripts/check_codes_sync.py` → `OK: 20 codes in sync` (exit 0)
- **C2** 5 envelopes additive + `error_detail` removed — `git grep error_detail packages/roxabi-contracts/src/` → no hits
- **C3** telemetry — `tests/core/test_metrics_format.py` asserts spec-pinned METRIC format verbatim
- **C4** workers + drivers populate — see `clipool_worker.py`, `nats_driver.py`, `cli_streaming_parser.py`
- **C5** hub helper + fallback removal — `git grep "Something went wrong" src/` → no hits
- **C6** integration test — `tests/integration/test_worker_error_e2e.py` passes end-to-end with `caplog` validating both METRIC lines
- **C7** soak gate — `tests/test_soak_gate.py` PASS+FAIL fixtures both assert correctly

C8 (shim deletion) deferred to follow-up #1029, gated on production 48h soak PASS.

## Test Plan
- [ ] CI green (lint + pyright + 3 coverage jobs + new `Verify error codes registry sync` step)
- [ ] Spot-check `uv run lyra hub` startup log includes `roxabi_contracts_version=…`
- [ ] Confirm `tools/check_worker_error_soak_gate.sh --fixture tests/fixtures/soak_gate_pass.log` prints `PASS` exit 0
- [ ] Confirm `--fixture tests/fixtures/soak_gate_fail.log` prints `FAIL` exit 1
- [ ] Post-merge: monitor adoption telemetry; once 48h PASS in prod, work #1029 to drop the `error_text` shim

Closes #1016

---


MickaelV0:
## Code Review

6 fresh review agents (security-auditor, architect, backend-dev, devops, tester, product-lead) reviewed the diff vs `staging`. Findings consolidated below in Conventional Comments format.

**Verdict: Request changes** — 7 blocking findings; 4 of them converge across multiple agents.

---

### 🚫 Blockers

**`issue:` `eval` + unquoted `$SINCE` / `$FIXTURE` enables command injection** *(security-auditor 82% + devops 92% + tester convergence)*
  `tools/check_worker_error_soak_gate.sh:80–86`
  Root cause: `INPUT_CMD="journalctl --since \"$SINCE\" --no-pager"` and `INPUT_CMD="cat -- $FIXTURE"` are passed to `eval`. `--since '"; cmd; echo "'` injects shell. A fixture path with spaces or metacharacters word-splits.
  Solution: drop `eval`. Replace with two arrays: `if [ -n "$FIXTURE" ]; then INPUT_ARGS=(cat -- "$FIXTURE"); else INPUT_ARGS=(journalctl --since "$SINCE" --no-pager); fi` then `"${INPUT_ARGS[@]}" | grep ...`.

**`issue:` `NatsLlmDriver.stream()` discards `worker_error` from worker reply payloads** *(backend-dev 95% + architect 85%)*
  `src/lyra/llm/drivers/nats_driver.py:336–341`
  Root cause: the `elif event_type == "result":` branch reads `is_error` from the raw chunk dict but does NOT read `chunk.get("worker_error")`. clipool now populates it (T13) but the driver drops it; `ResultLlmEvent.worker_error=None`; hub falls through to the empty `error_text` shim. Same pattern in `complete()` for valid replies with worker-reported errors.
  Solution: `chunk.get("worker_error")` → `WorkerError.model_validate(...)` (try/except `ValidationError` → `None`) → pass to `ResultLlmEvent(worker_error=we)`. Mirror in `complete()` for the `data.get("error")` branch.

**`issue:` Generic transport exception branch in `complete()` mislabels code as `transport.timeout`** *(architect 90% + backend-dev 92%)*
  `src/lyra/llm/drivers/nats_driver.py:165–181`
  Root cause: the catch-all `except Exception` (after timeout/no-responders) emits `code="transport.timeout"` regardless of the real exception class. Inflates `worker_error_received_total{code=transport.timeout}` for non-timeouts; misleads operators and future i18n.
  Solution: use `code="worker.internal"` for the catch-all (or add `transport.error` to KNOWN_CODES if a generic-transport slot is preferred).

**`issue:` `cli_streaming_parser` JSONDecodeError silently drops — no `worker_error` emitted** *(architect 85% + product-lead 92%)*
  `src/lyra/core/cli/cli_streaming_parser.py:82–85`
  Root cause: spec C4 / N12 requires path (b) — "parser encountered malformed JSON / unknown event → `worker.parse`". Implementation silently skips. Comment acknowledges `worker.parse` is missing from KNOWN_CODES, but no fallback to `cli.parse` is wired either. C4 not met for path (b).
  Solution: emit a terminal `ResultLlmEvent(is_error=True, worker_error=WorkerError(code="cli.parse", message=str(exc)))` if the failure occurs on (or past) the result line. Add `worker.parse` to KNOWN_CODES if intent is to keep the spec wording; otherwise update the comment to reflect the `cli.parse` fallback decision.

**`suggestion(blocking):` `we.code.split(".")[0]` bypasses authoritative `KNOWN_CODES[code].domain`** *(backend-dev 85%)*
  `src/lyra/core/processors/stream_processor.py:129`
  Root cause: registry already provides `.domain`; the string heuristic diverges for unknown codes (silently emits a synthetic domain).
  Solution: `KNOWN_CODES.get(we.code, CodeMeta(domain=we.code.split(".")[0], default_retryable=True, description="")).domain`.

**`issue:` `soak_gate_fail.log` triggers two FAIL conditions simultaneously — gate logic is not independently tested** *(devops 88% + tester 75%)*
  `tests/fixtures/soak_gate_fail.log`
  Root cause: missing `domain=llm` line AND non-empty `error_text=` line both fire FAIL. Gate test passes for the wrong reason; only 1 of 3 failure modes is exercised in isolation.
  Solution: add `tests/fixtures/soak_gate_fail_no_llm.log` (cli ✓, llm ✗, no legacy) and parametrize `test_soak_gate.py` over all three failure modes.

**`suggestion(blocking):` Contradiction check predicate is correct but reads ambiguously** *(architect 80%)*
  `src/lyra/core/messaging/error_extractor.py:29–32`
  Root cause: `(is_err is False) or (ok_flag is True)` relies on `None is False == False` to be safe across legacy envelopes. A future refactor that "tightens" the OR to `is_err is None` would silently fire false positives on `CliControlAck`.
  Solution: 1-line comments documenting the `is`-identity intent on each leg.

---

### ⚠️ Warnings (non-blocking)

- **`suggestion(non-blocking):` `str(exc)` in `WorkerError.message` can leak credential material** (security-auditor 75%) — NATS exception messages often embed `nats://user:pass@host`. Consider `max_length=512` on `message` + scrubbing connection-string regex.
  `src/lyra/adapters/clipool/clipool_worker.py:55–68`, `src/lyra/llm/drivers/nats_driver.py:134–157`
- **`thought:` log forging via newlines in `we.code`** (security-auditor 68%) — `code` validator allows newlines; soak-gate `grep -c` could be manipulated. Add `pattern=r"^[a-z][a-z0-9._-]*$"` to `WorkerError.code`.
- **`suggestion(non-blocking):` `clipool_worker._handle_cmd` ValidationError branch sends `is_error=True` with no `worker_error` + no metric** (backend-dev 78%) — invisible failure mode. Construct `WorkerError(code="transport.parse", ...)` + `emit_populated_total`.
- **`issue:` `test_metrics_format.py` `caplog` not logger-scoped** (tester 80%) — `caplog.records[-1]` is order-dependent under parallel runs. Use `caplog.at_level(logging.INFO, logger="lyra.core.messaging.metrics")` and `any(r.getMessage() == expected for r in records)`.
- **`issue:` POSIX-portability comment is false** (devops 85%) — script uses `[[`, `=~`, `pipefail`. Drop the claim, change comment to "Requires bash".
- **`thought:` 3 classifier sites (`_classify_exception`, `_classify_cli_error`, inline in `nats_driver`) replicate the taxonomy** (architect 75%) — flag for P3 consolidation when SDK synthesis lands; not action for this PR.
- **`nitpick:` `metrics.py` couples to `roxabi_contracts` only for `log_contracts_version()`** (architect 70%) — consider moving the version log to `bootstrap/`.
- **`thought:` `complete()` legacy `data.get("error")` path doesn't synthesize `WorkerError`** (product-lead 85%) — transitional P2 behavior, document inline.

---

### ✅ Praise

- `WorkerError` model shape is minimal and correct; `extra="ignore"` consistent with ADR-049 (architect 95%).
- `_extract_worker_error` deliberately omits `message` / `detail` from contradiction warning — correct defensive choice (security-auditor).
- METRIC log strings use `%s` lazy interpolation; format pinned in docstring guards against drift (backend-dev).
- `getattr(reply, "worker_error", None)` extractor pattern — clean across legacy + new envelopes (backend-dev).
- AAA structure consistent across all 8 new test files; boundary tests for `detail` (2048/2049) exactly right (tester).
- `check_codes_sync.py` three-way SKIP/FAIL/OK exit semantics — operationally correct (devops).
- C1, C2, C7 fully met per spec (product-lead).

---

### Verdict

**Request changes.** 7 blockers. Most consequential:
1. **`nats_driver.stream()` drops `worker_error`** — the entire P2 instrumentation chain is broken for the streaming path. Without this fix, the user-observable C5 promise is not delivered for the most common code path (LLM streaming).
2. **`eval` injection** in the soak-gate script — security defect.
3. **C4 path (b) not implemented** — spec gap.

Recommend `/fix` to apply the 7 blockers + tester `caplog` scoping issue.

---
Generated by `/code-review` (6 parallel agents)

MickaelV0:
## Review Fixes Applied

All 7 blockers + 4 actionable warnings from the prior review applied with long-term solutions. Verified: `ruff check` clean, `pyright` clean, full pytest 2910 passed / 41 skipped, contracts 496 passed, `check_codes_sync.py` OK.

### Blockers — applied

- [applied] `issue:` `eval` injection in soak-gate script — replaced with bash argv array (`INPUT_ARGS=(...)`); shell metacharacters in `--since` / `--fixture` no longer interpreted.
- [applied] `issue:` `nats_driver.stream()` discards `worker_error` — result branch now decodes `chunk["worker_error"]` via `_decode_worker_error()` (validates with Pydantic, logs at debug on shape mismatch). Synthesises `worker.internal` fallback when worker reports `is_error=True` without an envelope, so the hub instrumentation chain always activates on the streaming path.
- [applied] `issue:` Generic transport exception mislabelled `transport.timeout` — catch-all branch in `complete()` now uses `worker.internal` (the registry slot for "internal failure not covered by a more specific code"). Inline comment documents why.
- [applied] `issue:` `cli_streaming_parser` JSONDecodeError silently dropped — JSON-shaped lines (`startswith("{")`) now emit a terminal `ResultLlmEvent` carrying `cli.parse` + `emit_populated_total(domain="cli")`. Debug lines (`[DEBUG] …`, `INFO:`, blank) still silently skipped.
- [applied] `suggestion(blocking):` `we.code.split(".")[0]` bypasses `KNOWN_CODES.domain` — `meta = KNOWN_CODES.get(we.code); domain = meta.domain if meta else we.code.split(".")[0]`. Authoritative for known codes, safe fallback for unknown.
- [applied] `issue:` `soak_gate_fail.log` triggers two FAIL conditions — fixtures split into `soak_gate_fail_no_cli.log`, `soak_gate_fail_no_llm.log`, `soak_gate_fail_legacy_text.log`. Test parametrized so each gate branch is independently exercised. Original `soak_gate_fail.log` retained (now no-cli content) for backward compat.
- [applied] `suggestion(blocking):` Contradiction predicate ambiguity — added inline comment block on `error_extractor.py` documenting why `is False` / `is True` (identity, not truthiness) is load-bearing across legacy envelopes.

### Warnings — applied

- [applied] `issue:` `test_metrics_format.py` `caplog` not logger-scoped — switched to `caplog.at_level(logging.INFO, logger="lyra.core.messaging.metrics")` + order-independent `any(r.getMessage() == expected for r in records)` assertion.
- [applied] `suggestion(non-blocking):` `str(exc)` may leak credentials — added `_strip_credentials` field validator on `WorkerError.message` + `WorkerError.detail` that scrubs `scheme://user:pass@host` → `scheme://***:***@host` for nats / nats+tls / amqp / redis / http(s) / postgres / mysql. Also `max_length=512` on `message`.
- [applied] `thought:` log forging via newlines in `we.code` — added `pattern=r"^[a-z][a-z0-9._-]*$"` to `WorkerError.code`. All 20 KNOWN_CODES already conform.
- [applied] `suggestion(non-blocking):` clipool `_handle_cmd` ValidationError branch silent — now populates `WorkerError(code="transport.parse", …)` + `emit_populated_total(domain="cli")` so the soak gate sees this failure mode.

### Skipped

- `nitpick:` `metrics.py` couples to `roxabi_contracts` only for `log_contracts_version()` — style preference; current location is intentional (the helper is a metrics emission primitive, not bootstrap glue).
- `thought:` 3 classifier sites replicate the taxonomy — explicitly flagged as P3 SDK-synthesis work (#1029); not in scope for this PR.

### Verification

```
ruff check . → All checks passed
pyright      → 0 errors, 0 warnings, 0 informations
pytest tests/   → 2910 passed, 41 skipped
pytest packages → 496 passed, 24 skipped
check_codes_sync.py → OK: 20 codes in sync
```

Commit: 64074d42

MickaelV0:
## Code Review (iteration 2)

3 fresh agents (backend-dev, tester, security-auditor) re-reviewed the fix commit `64074d42`. The main blockers from iteration 1 are confirmed resolved on the surfaces they touched, but agents converged on **2 new blockers introduced/exposed by the fix** plus **2 high-confidence security issues** the iteration-1 fix didn't fully close.

**Verdict: Request changes** — 4 blocking findings.

---

### 🚫 Blockers

**`issue:` `clipool_worker._handle_cmd_streaming` drops `worker_error` from the streaming `ResultLlmEvent`** *(backend-dev 90%)*
  `src/lyra/adapters/clipool/clipool_worker.py:212-227`
  Root cause: same class as the iteration-1 `nats_driver.stream()` blocker, one layer up. `_handle_cmd_streaming` builds the terminal reply chunk with `is_error=event.is_error, session_id=event.session_id, done=True` but does NOT forward `event.worker_error`. `CliStreamingParser` now populates `worker_error=cli.auth | cli.session_lost | cli.parse` (T16/T17 + iteration-1 fix), and `nats_driver._stream_gen` now decodes `chunk.get("worker_error")` — but the field is `None` on the wire because clipool dropped it. Every CLI streaming error therefore lands on `nats_driver`'s `worker.internal` synthesis fallback instead of the precise `cli.*` code. The user-observable C5 promise is still unmet on the streaming path that actually goes through clipool.
  Solution: add `worker_error=event.worker_error` to the `_make_chunk(...)` call at line 213. `CliChunkEvent` already accepts the field.

**`issue(blocking):` Credential scrubber leaks password suffix when password contains `@`** *(security-auditor 85%, OWASP A3)*
  `packages/roxabi-contracts/src/roxabi_contracts/errors.py:20-22`
  Root cause: regex `[^/@\s:]+:[^/@\s]+@` stops at the first `@`. Input `nats://user:p@ssword@host:4222` becomes `nats://***:***@ssword@host:4222` — the substring `ssword@host:4222` stays in plain text. NATS / Postgres / Redis URLs commonly carry `@` in passwords (RFC 3986 allows it, percent-encoding is optional).
  Solution: parse with `urllib.parse.urlsplit` and replace `netloc` userinfo explicitly, or use `r"[^@\s]+@(?=[^@\s]+[:/])"` to anchor on the LAST `@` before the host. The robust path is the urlsplit one — regex-based parsing of URLs is fragile.

**`issue:` `WorkerError.message` `max_length=512` raises `ValidationError` instead of truncating — every `str(exc)` call site is unguarded** *(security-auditor 85%)*
  `packages/roxabi-contracts/src/roxabi_contracts/errors.py:33`
  Root cause: pydantic `max_length` is a constraint, not a transformer. Long stack traces (NATS framing errors, Pydantic validation reports, deep CLI exceptions) routinely exceed 512 chars. ~10 sites construct `WorkerError(message=str(exc) or ...)` (`nats_driver.py:158/177/201/221/327/351`, `clipool_worker.py:57/63/68/169`). On overflow `WorkerError(...)` raises inside the `except` handler that was meant to *return* a structured error — the error path itself crashes, the worker stops emitting envelopes, and the soak gate sees a silent regression rather than a populated failure.
  Solution: change `max_length=512` to a `@field_validator` that truncates with `…` suffix. Field-level constraints are wrong for "best-effort sanitization" — they belong on `code` (where rejection is correct) but not on free-text `message`.

**`issue:` `error_text=str(exc)` in `cli_streaming_parser` JSONDecodeError branch bypasses the credential scrubber** *(security-auditor)*
  `src/lyra/core/cli/cli_streaming_parser.py:106`
  Root cause: `ResultLlmEvent.error_text` is a raw `str` field — no `WorkerError` validator runs on it. The sibling `WorkerError(message=...)` at line 97 IS scrubbed, but the parallel `error_text` assignment carries the unscrubbed exception text. If a malformed CLI response embeds a URL fragment containing credentials, it leaks via `error_text` directly to the hub renderer.
  Solution: `error_text=worker_error.message` (reuse the already-scrubbed payload) or `error_text=f"CLI emitted malformed JSON: {exc}"` (no `str(exc)` at all).

---

### ⚠️ Warnings (non-blocking, but worth fixing in this round)

- **`issue:` `worker.internal` is semantically wrong for transport-layer catch-all in `nats_driver.complete()`** (backend-dev 85%) — registry definition reads "Worker encountered an internal error". A connection reset never reached the worker. Add `transport.error` to KNOWN_CODES (1-line) and use it here. The iteration-1 fix removed `transport.timeout` mislabel but the replacement inverts the transport/worker domain.
  `src/lyra/llm/drivers/nats_driver.py:199-203`
- **`issue:` `clipool_worker._handle_cmd` ValidationError uses `transport.parse` but should be `worker.validation`** (backend-dev 88%) — `transport.parse` is "couldn't decode bytes/JSON" but at this point JSON already decoded successfully; the failure is field-level schema mismatch, which is the textbook definition of `worker.validation` (already in KNOWN_CODES, retryable=False).
  `src/lyra/adapters/clipool/clipool_worker.py:167-169`
- **`issue:` `caplog` filter still not strict — `at_level(..., logger=...)` only sets the level threshold, doesn't filter records** (tester) — under parallel runs another logger emitting the same exact message would false-pass. Add `r.name == _METRICS_LOGGER` to the `_has` predicate.
  `tests/core/test_metrics_format.py:18-25`
- **`issue:` `test_legacy_dual_failure_still_fails` is misnamed — fixture is byte-identical to `soak_gate_fail_no_cli.log`** (tester) — name promises "dual failure" coverage; reality is single-branch. Rename to `test_original_fail_fixture_still_fails` OR replace `soak_gate_fail.log` with a genuinely dual-branch fixture.
  `tests/test_soak_gate.py:55-65`
- **`thought:` `error_extractor` contradiction warning fires but `stream_processor` doesn't actually honor "WE wins"** (backend-dev 78%) — extractor docstring says WE overrides is_error; `stream_processor` calls extractor only inside `event.is_error` branch, so a contradicted WE is silently dropped. Either enforce or update the docstring.
  `src/lyra/core/messaging/error_extractor.py:36`

### ⚠️ Test coverage gaps (tester)

The fix commit added 4 new code paths with **zero direct test coverage**:
- `nats_driver._decode_worker_error` (4 branches: None / non-dict / invalid / valid)
- `nats_driver._stream_gen` result branch with `worker_error` (synthesis + forward)
- `nats_driver.complete()` data["error"] + structured envelope forward
- `cli_streaming_parser` `{`-shaped malformed JSON → `cli.parse` (the C4 path b fix has no test!)
- `clipool_worker._handle_cmd` ValidationError → `transport.parse` + metric
- `WorkerError.code` regex rejection (uppercase, leading digit, newline)
- `WorkerError._scrub` credential scrubbing on `message` and `detail`

---

### ✅ Praise

- `_decode_worker_error` is correctly safe across all 3 input shapes (backend-dev 98%).
- `KNOWN_CODES.get(...)` fallback in `stream_processor` is correctly registry-first (backend-dev 96%).
- `eval` injection fix is complete — no remaining `eval` in script; argv-array passing prevents shell metachar interpretation in `$SINCE` and `$FIXTURE` (security-auditor).
- `code` regex rejects all relevant injection vectors (newlines, control chars, leading dot, empty, uppercase) (security-auditor).
- Multi-URL scrubbing works correctly via `re.sub` (security-auditor).
- 3 new soak-gate fixtures correctly isolate `no_cli`, `no_llm`, `legacy_text` failure modes (tester).
- Contradiction-warning log strings are safe — no PII, no log-forgery surface (security-auditor).

---

### Verdict

**Request changes.** 4 blockers. The most consequential:
1. **`clipool_worker._handle_cmd_streaming` drops `worker_error`** — same defect class as the iteration-1 `nats_driver.stream()` blocker, one layer up. Iteration-1 fixed the *receiver*; the *sender* still strips the field. Net effect on streaming path: `cli.auth` / `cli.session_lost` / `cli.parse` from the parser become `worker.internal` on the wire. Single-line fix.
2. **Credential scrubber `@`-in-password gap** — real OWASP A3 leak in a security-sensitive validator that was added specifically to close credential-leak risk.
3. **`max_length=512` raises** — the error path can crash the error path on long stack traces.

**Loop cap notice:** This is review iteration **2 of 2** allowed. After one more `/fix`, the next review will be capped — `/code-review` Phase 8 will then refuse to recommend Fix and force Merge as-is or Stop.

---
Generated by `/code-review` (3 parallel agents, iteration 2)

MickaelV0:

All 4 blockers + 5 actionable warnings + critical test-coverage gaps applied. Commit `d89724eb`.

### Blockers — applied

- [applied] `issue:` `clipool_worker._handle_cmd_streaming` drops `worker_error` — added `worker_error=event.worker_error` to the `_make_chunk(...)` call so `cli.auth` / `cli.session_lost` / `cli.parse` from the parser propagate into the published `CliChunkEvent` instead of becoming `worker.internal` at the hub.
- [applied] `issue(blocking):` Credential scrubber `@`-in-password leak — rewrote `_scrub_url` using `urllib.parse.urlsplit` + `rpartition("@")` so the netloc anchor is on the LAST `@` (RFC 3986). New scheme allowlist (`nats / nats+tls / amqp / amqps / redis / rediss / http / https / postgres / postgresql / mysql`). Loose URL match with regex, exact boundary detection with `urlsplit` — eliminates regex-based URL parsing fragility.
- [applied] `issue:` `max_length=512` raises on overflow — replaced with `@field_validator` `_sanitize_message` / `_sanitize_detail` that scrubs and truncates with `…` marker. Error-path code never crashes on long stack traces. Constants `_MESSAGE_MAX=512` / `_DETAIL_MAX=2048`.
- [applied] `issue:` `error_text=str(exc)` in cli_streaming_parser bypasses scrubber — now `error_text=worker_error.message` (already scrubbed via the validator).

### Warnings — applied

- [applied] `nats_driver.complete()` catch-all → `transport.error` (new KNOWN_CODES entry, `retryable=True`) instead of semantically-wrong `worker.internal`.
- [applied] `clipool_worker._handle_cmd` ValidationError → `worker.validation` instead of `transport.parse` (decoded OK, fields wrong = `worker.validation`).
- [applied] `caplog` filter on `r.name == _METRICS_LOGGER` — closes parallel-run flakiness.
- [applied] `test_legacy_dual_failure_still_fails` renamed to `test_dual_failure_fixture_fails_on_either_branch` + replaced fixture with genuine dual-branch (missing llm + non-empty error_text); test asserts both `populated_llm=0` AND `legacy_text=1` in stderr breakdown.
- [applied] `error_extractor` docstring updated to document that contradiction is surfaced but not enforced (and why), defusing future "tightening" refactors.

### Test coverage — added

- `WorkerError.code` pattern rejection: 10 parametrized cases (uppercase, mixed-case, leading-digit, newline, tab, space, leading-dot, slash, colon, at-sign).
- Credential scrubber: 6 URL shapes (basic, `@`-in-password, multi-URL, URL-encoded, empty-password, postgres+redis); `detail` field also exercised; unknown-scheme negative case (`ftp://` left alone).
- `WorkerError.message` 1000-char input truncated to 512 with `…`; same for `detail` 2049 → 2048.
- `nats_driver._decode_worker_error`: 4 branches (None, non-dict, invalid dict, valid dict).
- `complete()` structured-envelope forward path (worker provides `worker_error` dict → preserved, not synthesised).
- `complete()` `transport.error` catch-all on `ConnectionResetError`.
- `clipool` streaming-path `worker_error` forward (cli.session_lost preserved on the wire).
- `clipool` ValidationError → `worker.validation` envelope + retryable=False.
- `cli_streaming_parser` brace-shaped malformed JSON → `cli.parse` (covers spec C4 path b for the first time).

### Verification

```
ruff check .                       → All checks passed
pyright (CI scope)                 → 0 errors, 0 warnings
pytest tests/ + packages/          → 3435 passed, 65 skipped
check_codes_sync.py                → OK: 21 codes in sync
```

Commit: d89724eb

> ⚠️ **Loop cap reached.** This is fix iteration 2 of 2. The next `/code-review` is the final iteration — Phase 8 will refuse to recommend Fix and force Merge as-is or Stop.

MickaelV0:
## Code Review — Iteration 3 (Final, Loop Cap Reached)

Branch rebased on `origin/staging` (`f86ba848`). Three fresh agents reviewed the post-rebase state focused narrowly on iter-2 fix verification.

### Blockers (1)

**issue (blocking): `nats_driver.stream()` WorkerError forwarding has no direct test coverage** — _Confidence: 95% (tester)_

`src/lyra/llm/drivers/nats_driver.py` lines 382–395: `_decode_worker_error(chunk.get("worker_error"))` plus the `worker.internal` synthesis when `is_error=True` without envelope. Neither sub-branch is exercised end-to-end in `tests/llm/drivers/test_nats_driver.py`:

- Missing: `result`-chunk with `worker_error` dict → assert forwarded `ResultLlmEvent.worker_error.code`.
- Missing: `result`-chunk with `is_error=True` + no `worker_error` → assert synthesised `code == "worker.internal"`.

`TestDecodeWorkerError` covers the helper in isolation but doesn't wire it through `_stream_gen`.

**Root cause:** Iter-2 added `_decode_worker_error` and the `stream()` envelope-forwarding logic but only added direct tests for `complete()` (`TestCompleteWorkerError`). The `stream()` path was the original iter-1 regression — direct test still absent.

**Fix:** Add `TestStreamWorkerError` class with two cases (≈30 LOC).

### Warnings / Suggestions

None.

### Praise

- **praise**: `nats_driver` exception ordering — `NoRespondersError` correctly precedes `nats.errors.Error` catch-all; `transport.error` registry slot used semantically (transport not worker domain).
- **praise**: Producer paths intact post-rebase — `clipool_worker._handle_cmd_streaming` forwards `event.worker_error` into the chunk; `cli_streaming_parser` populates on both error branches.
- **praise**: `_extract_worker_error` honours docstring — does NOT enforce "WE wins", only logs contradiction.
- **praise**: Scrubber bypass-resistant — `urlsplit` + `rpartition("@")` handles `@`-in-password, URL-encoded userinfo, IPv6 hosts, empty password, unknown-scheme passthrough.
- **praise**: Soak gate — argv array `INPUT_ARGS=("${INPUT_ARGS[@]}")`, no eval, no command substitution, `cat -- "$FIXTURE"` blocks `-`-prefixed filename injection.
- **praise**: Code regex `^[a-z][a-z0-9._-]*$` correctly anchored — newlines and control chars blocked.
- **praise**: Soak-gate dual-failure fixture asserts both predicates (`populated_llm=0` AND `legacy_text=1`) independently in stderr.
- **praise**: `test_transport_catchall_uses_transport_error` post-rebase — `ConnectionClosedError` MRO confirmed: subclass of `nats.errors.Error`, not `NoRespondersError`/`TimeoutError`. Catch-all branch correctly exercised.

### Verdict

**Approve with comments.** One coverage blocker remains. **Loop cap reached** (this is iteration 3 of 3) — per skill safety, Phase 8 will refuse to recommend Fix. Decision is now Merge as-is (file the gap as a follow-up issue) or Stop.

---

🤖 Generated by `/code-review` — final iteration.

MickaelV0:
## Review Fixes Applied — Iteration 3

Single blocker from final review closed in `b6e200f9`.

### Applied
- **[applied]** `issue (blocking, 95% C)` — `nats_driver.stream()` worker_error path coverage gap
  - Added `TestStreamWorkerError` (2 cases, +126 LOC) in `tests/llm/drivers/test_nats_driver.py`
  - `test_stream_forwards_structured_worker_error` — chunk with `worker_error` dict → forwarded verbatim into `ResultLlmEvent.worker_error`
  - `test_stream_synthesises_worker_internal_when_envelope_absent` — `is_error=True` without envelope → synthesised `worker.internal` with worker-supplied message + retryable preserved

### Verification

| Gate | Result |
|---|---|
| `ruff check .` | All checks passed |
| `ruff format --check .` | 648 files already formatted |
| `pytest tests/ packages/` | 3445 passed, 65 skipped |

Ready to merge.

PR #1032, opened 2026-05-02
refactor(scripts): migrate gen-nkeys.sh → scripts/gen_nkeys.py

### LINKED ISSUE #1017: migrate gen-nkeys.sh → scripts/gen_nkeys.py
## Problem

`deploy/nats/gen-nkeys.sh` has grown from a key-generation script into a full program responsible for:
- Parsing `acl-matrix.json` (via `jq`)
- Filtering retired identities
- Deriving inbox grants from `request_reply_flows`
- Validating identity existence, charset, and lifecycle fields
- Generating `auth.conf` (templating)
- Calling `nk` CLI for key generation
- Emitting merged authconf
- Warning on retired seeds on disk

A code-quality audit (2026-05-01, PRs #975–#1013) identified **11 distinct bash safety bugs** across PRs #993, #1005, #1011, #1013. Of those, ~7 are **invisible to static analysis** (shellcheck) because they are jq-semantic or logic bugs:

- `jq //` coerces `false`→`true` — caused a silent security ACL regression (PR #993, 97% confidence)
- `jq -r` returns literal string `"null"` for absent fields — caused spurious FAIL in validation
- Missing identity-existence guard before subject interpolation — unknown responder hard-crashes jq
- `check_derived_output` reads unmodified JSON, never exercises the mutated regression path
- No character allowlist before NATS subject interpolation
- `return >255` mod-256 wrap — check_flows falsely passes on 256+ failures
- `2>/dev/null` swallows jq errors silently

Shellcheck (band-aid) would catch ~4/11. The root cause is **bash is the wrong language for ACL/identity logic**.

## Solution

Migrate to `scripts/gen_nkeys.py` (Python):

- `false` stays `false` in JSON — no `//` coercion
- `TypedDict` / Pydantic for acl-matrix schema → missing fields are type errors
- `subprocess.run(["nk", ...])` wraps the CLI
- Every derivation rule gets a pytest unit test (fixes RC-1 / test discrimination simultaneously)
- Reuses `roxabi-contracts` patterns already in the codebase

**Interface unchanged**: same CLI flags, same stdin/stdout contract, same output format. Drop-in replacement for all callers (CI scripts, Makefile, deploy runbook).

## Scope

- `scripts/gen_nkeys.py` — main implementation
- `scripts/check_acl_matrix_retired.py` — replaces `check-acl-matrix-retired.sh`
- `scripts/check_request_reply_flows.py` — replaces `check-request-reply-flows.sh`
- `tests/scripts/` — unit tests for every derivation rule and validation guard
- Update `.pre-commit-config.yaml` and `ci.yml` to call Python scripts
- Keep bash originals until Python parity is verified, then delete

## Acceptance criteria

- [ ] All existing CI checks pass via Python scripts
- [ ] `jq //` coercion bug is impossible by construction (Python json.loads)
- [ ] Every validation rule has a negative test (fails when guard removed)
- [ ] `nk` CLI calls wrapped and testable via mock
- [ ] Bash originals deleted

MickaelV0:
## Summary
- Port 702-LOC `deploy/nats/gen-nkeys.sh` + two CI sibling validators to typed Python under `scripts/`, with full test coverage and a nats-server e2e parity gate
- Swap all callers (Makefile, ci.yml, runbooks) to the new `lyra-acl` / `lyra-check-*` console-scripts, then delete the bash originals

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1017: migrate gen-nkeys.sh → scripts/gen_nkeys.py | Open |
| Analysis | [1017-migrate-gen-nkeys-py-analysis.mdx](artifacts/analyses/1017-migrate-gen-nkeys-py-analysis.mdx) | Present |
| Spec | [1017-migrate-gen-nkeys-py-spec.mdx](artifacts/specs/1017-migrate-gen-nkeys-py-spec.mdx) | Present |
| Implementation | 37 commits on `feat/1017-migrate-gen-nkeys-py` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (70 pass, 2 skip + 6 e2e) | Passed |

## What's included

**New Python modules** (`scripts/`):
- `_acl_models.py` — TypedDicts + `ParsedAuthConf` frozen dataclass
- `_loader.py` — `load_matrix()` with full validation (version, owner, status, dates, dup-flows)
- `_renderer.py` — `render_auth_conf()` + `parse_auth_conf()` (parity-verified vs bash)
- `_supervisor.py` — `validate_supervisor()` for NATS_NKEY_SEED_PATH wiring checks
- `_nk.py` — `NkeyProvider` ABC, `FakeNkeyProvider`, `SubprocessNkeyProvider`, `ensure_nk_or_exit()`
- `gen_nkeys.py` — argparse router with all modes: `--template-only`, `--regen-authconf`, `--emit-merged-authconf`, `--regenerate`, `--show`, `--fix-perms`, default (dual-write)
- `check_acl_matrix_retired.py` — port of `check-acl-matrix-retired.sh`
- `check_request_reply_flows.py` — port of `check-request-reply-flows.sh` (NATS-wildcard-aware)

**Console-scripts** (`pyproject.toml`):
- `lyra-acl` → `scripts.gen_nkeys:main`
- `lyra-genkeys` → alias
- `lyra-check-acl-retired` → alias
- `lyra-check-flows` → alias

**Caller swaps**: ci.yml lines 85/88, Makefile lines 197/271/331, runbook docs

**Bash deletions**: `deploy/nats/gen-nkeys.sh`, `scripts/check-acl-matrix-retired.sh`, `scripts/check-request-reply-flows.sh`

**CI**: `nk` install step added (mirrors nats-server pattern, v0.4.4) — SHA256 should be verified before merge

## Test Plan
- [ ] `uv run pytest tests/scripts -k 'not e2e'` → 70 pass, 2 skip
- [ ] `lyra-acl genkeys --template-only --matrix deploy/nats/acl-matrix.json` exits 0
- [ ] `lyra-acl genkeys --validate-supervisor` exits 0 on production
- [ ] `lyra-check-acl-retired` exits 0 on `deploy/nats/acl-matrix.json`
- [ ] `lyra-check-flows` exits 0 on `deploy/nats/acl-matrix.json`
- [ ] CI e2e job (`test_parity_e2e.py`) passes with nk + nats-server installed
- [ ] Verify nk SHA256 for v0.4.4 linux/amd64 before merge

Closes #1017

---


MickaelV0:
## Code Review

### Blockers

**suggestion(blocking): Stale `gen-nkeys.sh` error messages in `gen_nkeys.py`**
  `scripts/gen_nkeys.py:140,179`
  -- devops
  Root cause: Error strings still say "run gen-nkeys.sh without flags first" / "run gen-nkeys.sh first". The bash script was deleted. An operator hitting a missing-seed error will follow a dead reference.
  Solutions:
    1. Replace both with `"run 'uv run lyra-acl genkeys'"` (recommended)
  Confidence: 97%

---

**suggestion(blocking): Stale `gen-nkeys.sh` references in `_renderer.py` template strings (emitted into `auth.conf`)**
  `scripts/_renderer.py:10,15`
  -- devops
  Root cause: The rendered `auth.conf` header still reads `"generated by gen-nkeys.sh"` and the regen instruction says `"sudo ./deploy/nats/gen-nkeys.sh --regenerate"`. Any `lyra-acl genkeys --template-only` run re-emits these stale strings. The committed `deploy/nats/auth.conf` carries them as well.
  Solutions:
    1. Update both strings to reference `lyra-acl genkeys`; re-run `--template-only` to refresh the committed `deploy/nats/auth.conf`. (recommended)
  Confidence: 92%

---

### Issues

**issue: `_fake_pubkeys` is dead code and duplicates the formula in `FakeNkeyProvider.pubkey_from_seed`**
  `scripts/gen_nkeys.py:105`
  -- architect
  Root cause: `_fake_pubkeys` is defined at line 105 but never called anywhere in the module. The `--template-only` branch at line 327–330 inlines the same `UDET{name.upper().replace('-','')}` formula directly. `FakeNkeyProvider` is the third copy of this formula. Three independent copies guarantee drift at the next edit.
  Solutions:
    1. Delete `_fake_pubkeys` entirely; have the `--template-only` branch call `FakeNkeyProvider().pubkey_from_seed(FakeNkeyProvider().gen_seed(name))` as the single source. (recommended)
    2. Delete `_fake_pubkeys`; leave the `--template-only` inline as-is, but document it as a copy of `FakeNkeyProvider`.
  Confidence: 95%

---

**issue: `_cmd_check_retired` and `_cmd_check_flows` inline logic that already lives in the standalone check scripts — three-way drift**
  `scripts/gen_nkeys.py:381,445` / `scripts/check_acl_matrix_retired.py` / `scripts/check_request_reply_flows.py`
  -- architect
  Root cause: The retired-check and flows-check logic were re-implemented inline in `gen_nkeys.py` rather than importing from the standalone scripts. Error strings, wildcard semantics (`_subject_covered`), and exit codes are now in two places each. A logic bug fix in one will silently not be applied to the other.
  Solutions:
    1. Extract `check_retired(matrix) → list[str]` and `check_flows(matrix) → list[str]` into `scripts/_checks.py`; have both the standalone CLI and `gen_nkeys` subcommands call the same function. (recommended)
    2. Have `_cmd_check_retired`/`_cmd_check_flows` delegate to the standalone modules' `main()` directly.
  Confidence: 91%

---

**issue: `_cmd_validate_supervisor` at line 370 duplicates the `--validate-supervisor` branch in `_cmd_genkeys` lines 335–343**
  `scripts/gen_nkeys.py:370`
  -- architect
  Root cause: Two code paths call `validate_supervisor()` and format the output identically. One is unreachable via the primary `genkeys` flag surface once the `validate-supervisor` subcommand exists.
  Solutions:
    1. Have `_cmd_genkeys` delegate to `_cmd_validate_supervisor(args)` when the flag is set, eliminating the copy. (recommended)
    2. Remove the `--validate-supervisor` flag from `genkeys`; direct all callers to `lyra-acl validate-supervisor`.
  Confidence: 87%

---

**issue: `_get_provider` silently falls back to `FakeNkeyProvider` when `nk` is not on PATH**
  `scripts/gen_nkeys.py:47-56`
  -- architect / security
  Root cause: The silent fallback exists to support tests, but there is no runtime signal distinguishing "intentional test run" from "production machine where nk was inadvertently removed". On a production host, this would write `UDETlyra-hub`-style fake pubkeys to `auth.conf` with no stderr output — producing a structurally valid file that nats-server will reject at runtime.
  Solutions:
    1. Remove the `shutil.which` fallback; require callers to set `NKEY_PROVIDER=fake` explicitly for tests. Call `ensure_nk_or_exit()` when `NKEY_PROVIDER` is unset and `nk` is absent. (recommended)
    2. Keep the fallback but emit a `WARNING: nk not found — using FakeNkeyProvider` to stderr so production logs surface the degradation.
  Confidence: 88%

---

**issue: `_run_genkeys` test helper is copy-pasted between `test_genkeys_modes.py` and `test_file_modes.py`**
  `tests/scripts/test_genkeys_modes.py:26-43` / `tests/scripts/test_file_modes.py:36-53`
  -- tester
  Root cause: Identical function defined in two files. If the `NKEY_PROVIDER` env key or the script invocation path changes, both copies must be updated in sync; drift is silent.
  Solutions:
    1. Extract to `tests/scripts/conftest.py` as a shared fixture or module-level utility. (recommended)
  Confidence: 92%

---

**issue: `atomic_write` exception/cleanup path has zero test coverage**
  `tests/scripts/test_file_modes.py:59-185`
  -- tester
  Root cause: All four `TestAtomicWrite` tests exercise the success path only. The `except` branch in `atomic_write` (gen_nkeys.py:81–85) — which calls `os.unlink(tmp)` when an exception occurs — is never triggered by any test. This is the primary atomicity safety property; it must be verified.
  Solutions:
    1. Add a test that monkeypatches `os.replace` to raise `OSError`, then asserts no temp file is left behind and the exception propagates. (recommended)
  Confidence: 95%

---

**issue: `TestRegenerateMode` and `TestShowMode` tests pass when run as root — CI assumption unguarded**
  `tests/scripts/test_genkeys_modes.py:200,229,258`
  -- tester
  Root cause: Tests assert `returncode != 0` on the grounds that the caller is non-root. In a Docker-based CI runner without a `USER` directive the process runs as UID 0, causing the root check to pass and the assertions to fail (or worse: trigger actual provisioning side effects).
  Solutions:
    1. Add `@pytest.mark.skipif(os.getuid() == 0, reason="requires non-root to test guard")` to all three test methods. (recommended)
  Confidence: 90%

---

**issue: `AUTH_DIR` env var serves dual purpose: path override + root-check bypass — orthogonal concerns coupled**
  `scripts/gen_nkeys.py:201-204`
  -- architect / security
  Root cause: `_require_root()` exits early when `AUTH_DIR` is set. Any caller that sets `AUTH_DIR` for path-override purposes also silently disables the privilege check — even if that was not the intent. On GitHub Actions this is reachable without root.
  Solutions:
    1. Introduce `LYRA_ACL_SKIP_ROOT_CHECK=1` as a separate test-only variable; keep `AUTH_DIR` as a pure path override. (recommended)
    2. Document and accept the coupling; add an assertion that `AUTH_DIR` is never set in non-test contexts.
  Confidence: 80%

---

**suggestion(non-blocking): `sys.path.insert` hack duplicated in three entry-point files**
  `scripts/gen_nkeys.py:20` / `scripts/check_acl_matrix_retired.py` / `scripts/check_request_reply_flows.py`
  -- architect
  Root cause: All three standalone scripts need repo root on `sys.path` for `scripts.*` imports. The hack is copy-pasted verbatim.
  Solutions:
    1. Centralise in `scripts/__init__.py` or a `scripts/_bootstrap.py` imported first in each entry point.
  Confidence: 85%

---

**suggestion(non-blocking): `integration` job does not document `NKEY_PROVIDER=fake` intent**
  `.github/workflows/ci.yml:128`
  -- devops
  Root cause: `nk` is not installed in the integration job. The silent `_get_provider` fallback masks this. If any integration test hits a real genkeys path, it silently uses fake keys with no CI signal.
  Solutions:
    1. Set `NKEY_PROVIDER: fake` explicitly in the integration job `env:` block to document the intent.
  Confidence: 80%

---

### Praise

**praise: Atomic write implementation is correct**
  `scripts/gen_nkeys.py:72-86`
  -- security
  `mkstemp(dir=path.parent)` ensures same-filesystem rename; exception handler calls `os.unlink(tmp)`; `os.replace` is POSIX-atomic. All three properties required by the spec are present.

**praise: `SubprocessNkeyProvider.pubkey_from_seed` tmp-file lifecycle is clean**
  `scripts/_nk.py:36-48`
  -- security
  `try/finally: os.unlink(tmp_path)` guarantees cleanup on both success and exception paths. The nk v0.4.x stdin incompatibility is handled correctly.

**praise: `test_parity_e2e.py` is a genuine e2e gate**
  `tests/scripts/test_parity_e2e.py`
  -- tester
  Boots a real nats-server with the rendered auth.conf, verifies structural properties, and runs a live nkey-authenticated connection. This is the only test of its kind in the repo and fills a real coverage gap.

---

## Summary

| Category | Count |
|---|---|
| Blocking | 2 (stale gen-nkeys.sh references) |
| Issues | 8 |
| Suggestions | 2 |
| Praise | 3 |

**Verdict: Request changes** — 2 blockers (stale operator-facing error messages + renderer header referencing the deleted `gen-nkeys.sh`). Issues are tech-debt items; none block correctness or CI.

MickaelV0:
## Code Review — Iteration 2

CI is green. The 2 blockers from iteration 1 (stale `gen-nkeys.sh` refs) are resolved. Fresh domain-specific review found **6 new blocking issues**, primarily spec-criterion gaps.

---

### 🔴 Blockers

**1. Stale `(not yet implemented)` help strings for fully-implemented flags**
`scripts/gen_nkeys.py:499,505,509,515,519,525`
Root cause: 6 argparse help strings say "not yet implemented" despite working `_mode_*` implementations. Operators see this on every `lyra-acl genkeys --help` invocation.
Fix: replace each string with an accurate description; remove unused `_NOT_YET` constant.
Confidence: 100%

---

**2. `_get_provider()` silently uses `FakeNkeyProvider` when `nk` absent — spec criterion unmet**
`scripts/gen_nkeys.py:52-55`
Root cause: Spec requires "nk absent → exit 1 with install hint". If `nk` disappears from production `$PATH`, `_mode_full_provision` silently writes fake pubkeys to `/etc/nats/nkeys/auth.conf` with no error. `ensure_nk_or_exit()` is defined in `_nk.py` but never called from any provisioning mode.
Fix: remove silent fallback; call `ensure_nk_or_exit()` at top of `_mode_full_provision`, `_mode_regen_authconf`, `_mode_emit_merged_authconf`. `NKEY_PROVIDER=fake` remains the explicit test opt-in.
Confidence: 98%

---

**3. e2e test missing retired-identity live-connection rejection assertion**
`tests/scripts/test_parity_e2e.py` (test absent)
Root cause: Spec requires "retired-must-be-absent (connect attempt rejected)". `test_retired_identity_excluded` only checks rendered text. No test attempts a live NATS connect with a retired seed and asserts `AuthorizationError`.
Fix: add `test_retired_identity_connect_rejected` — render `v2-with-retired.json`, connect with `old-worker`'s seed, assert connection is rejected.
Confidence: 97%

---

**4. Only `hub` (operator) tested with live NATS connection — adapter and worker missing**
`tests/scripts/test_parity_e2e.py:198`
Root cause: Spec requires one identity per role-class. `v2-prod.json` has `hub`, `voice-tts`, `clipool-worker`. Only `hub` has a live connect test.
Fix: add `test_voice_tts_can_connect` and `test_clipool_worker_can_connect`.
Confidence: 92%

---

**5. `TestLoadMatrixNegatives` missing `# verified:` annotations — spec criterion unmet**
`tests/scripts/test_loader.py:95-264`
Root cause: Spec explicitly requires each negative test to record guard-removal verification in the docstring. All other negative-test classes have `# verified:` comments; `TestLoadMatrixNegatives`'s 10 tests have `"""Guard: ..."""` but no `# verified: removing <guard> causes ...`.
Fix: add `# verified:` annotation to each of the 10 tests.
Confidence: 97%

---

**6. `gen_nkeys.py` is 598 LOC — exemption added, contradicting spec criterion**
`scripts/gen_nkeys.py:1` / `tools/file_exemptions.txt`
Root cause: Spec criterion: "All scripts/*.py under 300-line cap; no addition to tools/file_exemptions.txt." Slice 2 modes landed in the umbrella router rather than a helper module.
Fix: extract `_mode_*` implementations to `scripts/_modes.py`; remove exemption entry.
Confidence: 100%

---

### 🟡 Suggestions (non-blocking)

- **Dead code**: `_fake_pubkeys` at `gen_nkeys.py:105` is never called — delete it. (100%)
- **Duplication**: `_cmd_validate_supervisor` duplicates `--validate-supervisor` branch in `_cmd_genkeys` — delegate. (100%)
- **Auth bypass**: `_require_root()` is bypassed by any user setting `AUTH_DIR` env var; gate it on `NKEY_PROVIDER=fake` being also set. (72%)
- **e2e semantics**: fold `NATS_PY_AVAILABLE` into module-level `pytestmark` to avoid silent skip on machines with nats-server+nk but no nats-py. (88%)
- **CI**: add `unzip` to `Install system dependencies` step (implicit dep). (85%)

### Nitpicks

- `test_file_modes.py:133`: use `set(siblings) == {dest}` instead of list equality (order-sensitive). (70%)
- `ci.yml`: spec cites lines 70/73 for Python entry points; actual lines 86/89 due to `Install nk` step insertion — update spec, no code change. (95%)

---

### 🟢 Praise

- All `gen-nkeys.sh` references replaced ✓
- `ParsedAuthConf` frozen dataclass ✓
- All 4 `pyproject.toml` entry points ✓
- nk install: pinned, sha256-verified, correctly ordered ✓
- dual-write + SUDO_USER + file modes tested with `# verified:` annotations ✓
- bash originals deleted; no layering violation ✓

---

**Verdict: Request changes** _(Iteration 2 of 2 — next review must be Merge as-is or Stop per loop cap)_

_Generated with [Claude Code](https://claude.com/claude-code) via `/code-review`_

MickaelV0:
## Review Fixes Applied

All 6 iteration-2 blockers resolved in commit `fb743aa7`.

| # | Finding | Fix |
|---|---------|-----|
| 1 | `_get_provider()` silent `FakeNkeyProvider` fallback when `nk` absent | Replaced with `ensure_nk_or_exit()` call — exits 1 with install hint |
| 2 | 6 stale `(not yet implemented)` help strings | Updated to accurate descriptions; removed dead `_NOT_YET` constant |
| 3 | `_cmd_validate_supervisor` duplicated in `_cmd_genkeys` | `--validate-supervisor` branch now calls `_cmd_validate_supervisor(args)` |
| 4 | `gen_nkeys.py` at 598 LOC (spec: ≤300) | Extracted mode functions + helpers to `scripts/_modes.py` → 296 LOC; exemption removed |
| 5 | `TestLoadMatrixNegatives` missing `# verified:` annotations | Added to all 10 negative tests |
| 6 | Only `hub` tested for live NATS connect; no retired-identity rejection test | Added `test_voice_tts_can_connect`, `test_clipool_worker_can_connect`, `test_retired_identity_connect_rejected` |

Lint + pyright + 70 tests pass locally (11 skipped: nats-server/nk/nats-py absent).

---
Generated with [Claude Code](https://claude.com/claude-code)

MickaelV0:
## Code Review — Iteration 3

CI is green. Fresh review of the full diff (41 files, 3755+ / 899-).

### Blockers

**issue(blocking): `_mode_full_provision` and `_mode_fix_perms` drop all `chown` calls — bash parity gap**
  `scripts/_modes.py:235, 219`
  Root cause: Bash used `install -m 0600 -o LYRA_USER -g LYRA_USER` for seeds and `chown root:nats` for auth.conf. Python only calls `atomic_write` (chmod-only) + `seeds_dir.chmod(0o700)`. Running as root, all created files/dirs are owned by `root:root`. Seeds at `~/.lyra/nkeys/*.seed` and the dir itself become root-owned — unreadable by the operator process at runtime. `auth.conf` at `0640 root:root` is unreadable by the `nats` group.
  Solutions:
    1. Add `os.chown(path, uid, gid)` after each `atomic_write` using `pwd.getpwnam(os.environ["SUDO_USER"])` for seeds and `grp.getgrnam("nats").gr_gid` for auth.conf. Extract `_operator_uid_gid()` helper shared across modes. (recommended)
    2. Use `shutil.chown(path, user=SUDO_USER)` — same logic, cleaner.
  Confidence: 97%

---

**issue(blocking): `_mode_regenerate` has no restore-on-failure — bash `trap restore_backups ERR INT TERM` not ported**
  `scripts/_modes.py:165`
  Root cause: Bash installed `trap restore_backups ERR INT TERM` before any destructive op. Python backs up, then calls `shutil.rmtree(seeds_dir)` + `auth_conf.unlink()` + `_mode_full_provision(args)` with no exception handler. If `_mode_full_provision` fails (nk error, disk full), seeds are wiped and backups are the only copy — no automatic recovery. NATS auth is bricked until manual restore.
  Solutions:
    1. Wrap wipe + provision in `try/except BaseException`, restore from epoch-stamped backup in the `except` block, then re-raise. (recommended)
    2. Write new keys to a temp dir first, then atomically swap — eliminates the window where neither old nor new keys exist.
  Confidence: 95%

---

**issue(blocking): E2e test only tests `nats.connect()` — spec requires publish/subscribe ACL enforcement assertions**
  `tests/scripts/test_parity_e2e.py:194-261`
  Root cause: Spec: "asserts publish/subscribe authorization for one identity per role-class." All three connect tests call `nats.connect()` + `drain()` only. A misconfigured allow-list (wrong ACL subjects) would not be caught — NATS accepts the connection but ACL violations only surface on publish/subscribe.
  Solutions:
    1. After connect, publish to an allowed subject and assert success; then publish to a forbidden subject and assert `nats.errors.AuthorizationError`. (recommended)
    2. Add a subscribe-and-publish round-trip between hub and clipool-worker exercising the full ACL path.
  Confidence: 92%

---

**issue(blocking): Duplicate logic — standalone check scripts are dead code; entry points route through `gen_nkeys.py`**
  `scripts/check_acl_matrix_retired.py`, `scripts/check_request_reply_flows.py`
  Root cause: `lyra-check-acl-retired` and `lyra-check-flows` route through `gen_nkeys.py:_check_retired/_check_flows`. The standalone scripts are never called. Both paths implement the same logic independently — guaranteed to diverge.
  Solutions:
    1. Delete `check_acl_matrix_retired.py` and `check_request_reply_flows.py`. (recommended)
    2. Invert: keep standalone scripts as canonical, have `gen_nkeys.py` import and call their functions directly.
  Confidence: 95%

---

**issue(blocking): `sys.argv` mutation as dispatch mechanism — fragile, breaks on parser restructure**
  `scripts/gen_nkeys.py` — `alias_genkeys`, `_check_retired`, `_check_flows`
  Root cause: `sys.argv.insert(1, "check")` physically mutates the process argv to re-route through `main()`. Couples aliases to argument-parsing order; invisible to type checkers; breaks if subcommand schema changes.
  Solutions:
    1. Extract check logic into callable functions, have aliases parse their own args and call those functions directly. (recommended)
    2. Accept `argv=None` in `main()` and pass `["check", "retired"]` from the alias.
  Confidence: 92%

---

**issue(blocking): `scripts/` package in wheel — leaks ops tooling as top-level importable package**
  `pyproject.toml` — `packages = ["src/lyra", "scripts"]`
  Root cause: Bundling operational scripts into the wheel makes `import scripts` available to any downstream library consumer. Not application library code; pollutes package namespace.
  Solutions:
    1. Move ACL generation logic to `src/lyra/acl/`, update entry points, remove `scripts` from wheel packages. (recommended)
    2. Register under a namespaced name (`lyra_scripts`) via hatch path mappings.
  Confidence: 88%

---

**issue(blocking): `test_no_lyra_imports_in_scripts` guard has wrong boolean operator**
  `tests/scripts/test_no_lyra_imports.py:26`
  Root cause: `startswith("_") and == "__init__.py"` is always False for private modules other than `__init__.py`. If the intent is to skip only `__init__.py`, the condition should be just `py_file.name == "__init__.py"`.
  Solutions:
    1. Replace with `if py_file.name == "__init__.py": continue`. (recommended)
  Confidence: 88%

---

### Warnings

**suggestion(non-blocking): `TestRegenerate` missing test for restore-on-failure path**
  `tests/scripts/test_genkeys_modes.py:199`
  — No test covers "provision fails after wipe → backup is only recovery". Add `test_regenerate_full_provision_failure_preserves_backup`: patch `render_auth_conf` to raise, assert seeds_dir is wiped and `.bak.*` backup exists as sole recovery option.
  Confidence: 97%

**suggestion(non-blocking): E2e fixture hardcoded ports 4223/8222 — port collision in parallel CI**
  `tests/scripts/test_parity_e2e.py:77`
  — Use `nats-server -p 0` + read actual port from `/varz`, or add pre-fixture socket check + `pytest.skip`.
  Confidence: 95%

**suggestion(non-blocking): `setup.sh` passes `--yes` to `--regen-authconf` — silently ignored**
  `deploy/nats/setup.sh`
  — `--yes` suppresses `--regenerate`'s TTY prompt only. Remove from the `--regen-authconf` line.
  Confidence: 90%

**suggestion(non-blocking): CI swallows exit code 2 from `check_codes_sync.py` without documented contract**
  `.github/workflows/ci.yml`
  — Document EC=2 contract or remove the bypass.
  Confidence: 88%

**suggestion(non-blocking): `NKEY_PROVIDER=fake` has no production guard**
  `scripts/_modes.py:28`
  — Accidental `NKEY_PROVIDER=fake` in production writes deterministic ASCII as seeds. Add `LYRA_TEST_MODE=1` sentinel check alongside `fake`.
  Confidence: 75%

**suggestion(non-blocking): `conftest.py` `prod_matrix` fixture bypasses `load_matrix` validation**
  `tests/scripts/conftest.py:55`
  — Returns raw `json.loads(...)` dict. Change to `load_matrix(_REAL_MATRIX_JSON)` so validation fires at fixture setup time.
  Confidence: 82%

---

### Summary

| Category | Count |
|---|---|
| Blockers | 6 |
| Warnings | 6 |

**Verdict: Request changes**

Two blockers are production-safety regressions: missing `chown` (seeds will be root-owned, operator process can't read its own keys) and missing rollback in `--regenerate` (key wipe with no recovery on failure). The e2e ACL enforcement gap means the real-NATS gate required by spec is incomplete. The architecture blockers (dead standalone scripts, `sys.argv` mutation, `scripts/` in wheel) are clean-up from the previous iteration where the focus was on functional correctness.

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 6 findings
**Skipped (rejected by verifier):** 3 findings (F4 dead scripts, F7 bool logic, F11 CI EC=2)

### Auto-Applied

- [applied] `_modes.py`: add `_operator_uid_gid()` helper; `os.chown` seeds + seeds_dir to operator user in `_mode_full_provision` and `_mode_fix_perms`; `os.chown` auth.conf to `root:nats`; `auth_dir.chmod(0o750)` — **bash parity restored**
- [applied] `_modes.py`: wrap `_mode_full_provision` call in `_mode_regenerate` with `try/except BaseException`; restore seeds_dir and auth.conf from epoch-stamped backup on failure — **mirrors bash `trap restore_backups ERR INT TERM`**
- [applied] `test_parity_e2e.py`: 3 new pub/sub ACL enforcement tests (hub, voice-tts, clipool-worker) using `error_cb` + `"Permissions Violation"` assertion — **spec gate now actually tests ACL, not just auth**
- [applied] `test_genkeys_modes.py`: `test_regenerate_restores_on_provision_failure` — patches `_mode_full_provision` to raise, asserts seeds_dir is restored from backup
- [applied] `deploy/nats/setup.sh`: removed stray `--yes` from `--regen-authconf` invocation
- [applied] `tests/scripts/conftest.py`: `prod_matrix`, `legacy_matrix`, `with_retired_matrix` fixtures now call `load_matrix()` instead of raw `json.loads` — validation fires at fixture setup time

### Applied (1b1)

- [applied] `scripts/gen_nkeys.py`: `main(argv=None)` param; `alias_genkeys`, `_check_retired`, `_check_flows` pass explicit argv lists instead of mutating `sys.argv`
- [applied] `scripts/_modes.py`: `_get_provider()` exits 1 when `NKEY_PROVIDER=fake` is set without `LYRA_TEST_MODE=1`; updated `_run_genkeys` helpers in `test_genkeys_modes.py` and `test_file_modes.py` to set the sentinel

### Commits

- `33b2e5f4` fix(scripts): apply review fixes — chown, rollback, ACL e2e, conftest
- `4e0bfc82` fix(scripts): sys.argv-free aliases + NKEY_PROVIDER=fake production guard

### Not applied (verifier-rejected)

- F4 dead standalone scripts — `check_acl_matrix_retired.py` and `check_request_reply_flows.py` have dedicated test coverage invoking them via subprocess; they are not dead code
- F7 `test_no_lyra_imports` boolean — behavior is correct as-is; condition is redundant but achieves the right outcome
- F11 CI EC=2 — intentional documented skip path in `check_codes_sync.py`

71 tests pass, 2 skipped (nats-server/nk not on PATH in local env).

PR #1033, opened 2026-05-03
fix(trace): add bare token redaction to TelegramTokenFilter

### LINKED ISSUE #1028: fix(trace): TelegramTokenFilter misses bare token format (no bot prefix)
## Problem

`TelegramTokenFilter` uses `_TOKEN_RE = re.compile(r'bot(\d+):[A-Za-z0-9_-]+')` which only matches tokens embedded in Telegram HTTP URL paths (e.g. `https://api.telegram.org/bot123456:ABC.../method`).

A bot token logged as a bare string — e.g. from a config dump, exception repr, or a debug log of the token variable directly — is **not matched** because the literal `bot` prefix is absent.

Example unmatched: `token=123456:ABCDefghijklmnopqrstuvwxyz_1234567`

## Fix

Add a second pattern targeting the standalone form:
```python
_BARE_TOKEN_RE = re.compile(r'\b(\d{8,12}:[A-Za-z0-9_-]{30,50})\b')
```

Both patterns should be applied in `TelegramTokenFilter.filter()`.

## Notes

- Found in PR #1025 code review (iteration 2), deferred as pre-existing and out of scope.
- Current coverage: URL-embedded tokens (httpx logs). Missing: config dumps, exception reprs, direct token logs.

MickaelV0:
## Summary
- Adds `_BARE_TOKEN_RE` to `TelegramTokenFilter` to catch tokens logged without the `bot` URL prefix (config dumps, exception reprs, direct token variable logs)
- Both URL-embedded (`bot<id>:<secret>`) and bare (`<id>:<secret>`) forms are now redacted; URL pattern runs first to preserve the bot ID, bare pass cleans up any remaining standalone tokens

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1028: fix(trace): TelegramTokenFilter misses bare token format (no bot prefix) | Open |
| Implementation | 2 commits on `fix/1028-telegram-token-filter-bare-format` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (5 new) | Passed |

## Test Plan
- [ ] `uv run pytest tests/core/test_trace.py::TestTelegramTokenFilter -v` — all 9 pass
- [ ] Bare token in config dump string is redacted
- [ ] Bare token in exception repr is redacted
- [ ] Short secrets (`<30 chars`) are not over-matched
- [ ] URL-embedded token still produces exactly one `<REDACTED>` (no double-redact)

Closes #1028

---

### DIFF
--- a/docs/architecture/testing-conventions.md
+++ b/docs/architecture/testing-conventions.md
@@ -64,7 +64,7 @@ The `dev-core:tester` agent flags missing negative tests as `issue:` (≥90% con
 - Import + call real source functions — never mock the module under test
 - `unittest.mock.patch('module.Symbol')` / `mocker.patch(...)` — patch at the import site of the dependency, never patch the module under test itself
 - Integration tests (real modules wired) > unit tests with heavy mocks
-- Verify coverage: `uv run pytest --coverage <file>` (substitute your project's test command) — 0% → wrong mocking
+- Verify coverage: `uv run pytest --cov=src/<module> <test_path>` (substitute your project's test command) — 0% → wrong mocking
 
 ---
 
--- a/src/lyra/core/trace.py
+++ b/src/lyra/core/trace.py
@@ -113,20 +113,24 @@ class TelegramTokenFilter(logging.Filter):
     never appear in persisted logs, crash reports, or log-shipping
     pipelines.
 
-    Matches ``bot<digits>:<alphanumeric-with-hyphens-underscores>`` and
-    replaces the token segment with ``<REDACTED>``. Always returns
-    ``True`` — the record is never suppressed. Defensive: never raises.
+    Matches both the URL-embedded form (``bot<digits>:<secret>``) and the
+    bare form (``<digits>:<secret>`` from config dumps or exception reprs)
+    and replaces them with ``<REDACTED>``. Always returns ``True`` — the
+    record is never suppressed. Defensive: never raises.
     """
 
-    # Bot tokens: 1234567890:AAEhBP0av28...Z (numeric id, colon, ~35 char secret)
+    # URL-embedded: /bot1234567890:AAEhBP0av28...Z/method
     _TOKEN_RE = re.compile(r"bot(\d+):[A-Za-z0-9_-]+")
     _REDACTED_SUB = r"bot\1:<REDACTED>"
+    # Bare token: 1234567890:AAEhBP0av28...Z (config dumps, exception reprs)
+    _BARE_TOKEN_RE = re.compile(r"\b(\d{8,12}:[A-Za-z0-9_-]{30,50})\b")
 
     def filter(self, record: logging.LogRecord) -> bool:
         try:
             # Format the message once (% args interpolation) then redact.
             msg = record.getMessage()
             redacted = self._TOKEN_RE.sub(self._REDACTED_SUB, msg)
+            redacted = self._BARE_TOKEN_RE.sub("<REDACTED>", redacted)
             if redacted != msg:
                 record.msg = redacted
                 record.args = None
--- a/src/lyra/core/trace.py
+++ b/src/lyra/core/trace.py
@@ -119,11 +119,17 @@ class TelegramTokenFilter(logging.Filter):
     record is never suppressed. Defensive: never raises.
     """
 
-    # URL-embedded: /bot1234567890:AAEhBP0av28...Z/method
+    # URL-embedded: /bot<id>:<secret>/method (httpx always emits full-length secrets)
     _TOKEN_RE = re.compile(r"bot(\d+):[A-Za-z0-9_-]+")
     _REDACTED_SUB = r"bot\1:<REDACTED>"
     # Bare token: 1234567890:AAEhBP0av28...Z (config dumps, exception reprs)
-    _BARE_TOKEN_RE = re.compile(r"\b(\d{8,12}:[A-Za-z0-9_-]{30,50})\b")
+    # Lookbehind/lookahead instead of \b: \b breaks after trailing '-' (non-\w).
+    # No upper bound: avoids silently missing tokens with secrets > 50 chars.
+    # _REDACTED_SUB must stay < 30 chars so a partially-redacted URL-form token
+    # (bot<id>:<REDACTED>) is not re-matched by this pattern.
+    _BARE_TOKEN_RE = re.compile(
+        r"(?<!\w)(\d{8,12}:[A-Za-z0-9_-]{30,})(?![A-Za-z0-9_-])"
+    )
 
     def filter(self, record: logging.LogRecord) -> bool:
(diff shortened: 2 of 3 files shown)


MickaelV0:
## Code Review

### Blockers

**issue**: `test_redacts_bare_token_in_exception_repr` missing positive `<REDACTED>` assertion
  `tests/core/test_trace.py:443`
  -- tester
  Root cause: Only asserts the secret is absent — if the filter silently dropped the message or emitted a different string, the test would still pass. Sibling test `test_redacts_bare_token_in_config_dump` correctly adds the positive assertion.
  Solutions:
    1. Add `assert "<REDACTED>" in record.getMessage()` after the existing assert (recommended)
    2. Assert the full expected string: `assert record.getMessage() == "TelegramError: invalid token <REDACTED>"`
  Confidence: 95%

---

**issue**: `test_url_token_already_redacted_before_bare_pass` doesn't assert bot-id is preserved
  `tests/core/test_trace.py:463`
  -- tester
  Root cause: The test checks `count("<REDACTED>") == 1` but doesn't verify the URL-pass form (`bot8500388193:<REDACTED>`) survived. If the bare pass ran first and swallowed the token entirely, the bot-id would be lost and `count == 1` would still pass.
  Solutions:
    1. Add `assert "bot8500388193:<REDACTED>" in msg` to verify the URL-pass form (with bot ID) is the surviving redaction (recommended)
    2. Assert the full redacted URL string for maximum specificity.
  Confidence: 90%

---

**issue**: `_BARE_TOKEN_RE` upper bound `{30,50}` could silently miss tokens > 50 chars
  `src/lyra/core/trace.py:126`
  -- backend-dev
  Root cause: Telegram's current secret is 35 chars, but an arbitrary 50-char ceiling means any token with a longer secret is not redacted. The URL-pattern (`_TOKEN_RE`) uses unbounded `+`, creating an inconsistent contract between the two patterns.
  Solutions:
    1. Change to `{30,}` — "at least the minimum known length" with no ceiling (recommended)
    2. Add the same `{30,}` to `_TOKEN_RE` for symmetry: `r"bot(\d+):[A-Za-z0-9_-]{30,}"`
  Confidence: 88%

---

### Suggestions

**nitpick**: `\b` word-boundary ineffective after trailing `-` in secret
  `src/lyra/core/trace.py:126`
  -- backend-dev
  Root cause: `\b` is `\w`/`\W` boundary; `-` is `\W`, so `\b` does not fire after a trailing hyphen in the secret. Opening `\b` on the digit run is correct; closing boundary is broken for hyphen-terminated secrets.
  Solutions:
    1. Replace with negative lookahead: `r"(?<!\w)(\d{8,12}:[A-Za-z0-9_-]{30,50})(?![A-Za-z0-9_-])"` (recommended)
    2. Drop trailing `\b` — in practice tokens appear at word boundaries anyway.
  Confidence: 85%

---

**thought**: Ordering dependency between `_TOKEN_RE` → `_BARE_TOKEN_RE` is implicit
  `src/lyra/core/trace.py:131-132`
  -- backend-dev
  Root cause: The second regex runs on a partially-redacted string. Today it's safe (`<REDACTED>` is 9 chars, below the 30-char floor) but a future change to `_REDACTED_SUB` could silently break this. Security-auditor confirmed no current risk; document the invariant.
  Solutions:
    1. Add a comment: `# _BARE_TOKEN_RE runs second; _REDACTED_SUB must stay < 30 chars to avoid re-matching`
    2. Apply both regexes to the original `msg` separately and merge, or reverse order (bare first, URL second — no overlap since URL form requires `bot` prefix).
  Confidence: 80%

---

**suggestion(non-blocking)**: Add test for 7-digit ID floor (digit count minimum)
  `tests/core/test_trace.py:454`
  -- tester
  Root cause: `test_bare_token_not_over_matched_by_short_secret` targets the secret-length guard but not the `\d{8,12}` digit-count floor. A 7-digit ID is not covered.
  Solutions:
    1. Add: `record = self._make_record("id=1234567:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA")` and assert no redaction.
    2. Rename existing test to `test_bare_token_not_matched_when_secret_too_short` for precision.
  Confidence: 75%

---

**suggestion(non-blocking)**: `_TOKEN_RE` has no minimum secret length (pre-existing, now visible by contrast)
  `src/lyra/core/trace.py:123`
  -- backend-dev
  Root cause: `[A-Za-z0-9_-]+` matches a 1-char suffix, e.g. `bot42:X` in unrelated log lines. The new `_BARE_TOKEN_RE` enforces `{30,}` — the contrast makes this inconsistency auditable.
  Solutions:
    1. Change to `r"bot(\d+):[A-Za-z0-9_-]{30,}"` — aligns with bare-token minimum.
    2. Leave and document assumption that httpx URL-form logs always have full-length secrets.
  Confidence: 75%

---

**nitpick**: Per-method `TelegramTokenFilter` imports could be module-level
  `tests/core/test_trace.py:433,444,456,465`
  -- tester
  Root cause: Consistent within the class but the file-level import at line 17 is the preferred style. Per-method imports add noise without isolation benefit.
  Solutions:
    1. Add `TelegramTokenFilter` to module-level import at line 17.
    2. Leave as-is for consistency within the class.
  Confidence: 70%

---

### Summary

| Domain | Findings | Verdict |
|--------|----------|---------|
| Security | 0 | ✅ Clean |
| Tests | 4 (2 blocking, 2 suggestions) | ❌ Changes requested |
| Backend | 4 (1 blocking, 3 suggestions) | ❌ Changes requested |

**Verdict: Request changes** — 3 blockers (missing test assertions ×2, bare-token upper bound).

---
*Generated with [Claude Code](https://claude.com/claude-code) via `/code-review`*

MickaelV0:
## Review Fixes Applied

**Skipped:** 2 finding(s)
**Failed:** 0

### Auto-Applied
- [applied] `_BARE_TOKEN_RE` upper bound `{50}` removed → `{30,}` (C:88%)
- [applied] `\b` boundary replaced with lookahead/lookbehind to fix trailing-hyphen edge case (C:85%)
- [applied] test: positive `<REDACTED>` assertion added to `test_redacts_bare_token_in_exception_repr` (C:95%)
- [applied] test: `assert "bot8500388193:<REDACTED>" in msg` added to double-redact guard test (C:90%)

### Applied (1b1)
- [applied] test: `test_bare_token_not_matched_when_id_too_short` added for 7-digit ID floor (C:75%)

### Skipped
- `_TOKEN_RE` min secret length (C:75%) — adding `{30,}` would break 2 existing tests using short fake secrets; httpx always emits full-length secrets (documented in comment)
- Per-method `TelegramTokenFilter` imports (C:70%) — consistent within class; left as-is

---
*Generated with [Claude Code](https://claude.com/claude-code) via `/fix`*

PR #1041, opened 2026-05-03
fix(scripts): auto-derive ROWS in check-acl-matrix-spec.sh from acl-matrix.json

### LINKED ISSUE #1039: fix(scripts): auto-derive ROWS in check-acl-matrix-spec.sh from acl-matrix.json
## Problem

\`scripts/check-acl-matrix-spec.sh\` validates the ACL matrix spec doc against \`acl-matrix.json\`, but the ROWS array (lines 54–77) is hardcoded bash. When a new subject is added to \`acl-matrix.json\`, it does not automatically appear in the validation — someone must manually add it to ROWS.

This is a silent gap: new subjects pass CI without being checked against the spec table.

## Expected behaviour

ROWS should be derived directly from \`acl-matrix.json\` (union of all \`publish[]\` and \`subscribe[]\` subjects across active identities), so new subjects are picked up automatically.

## Effort: M

MickaelV0:
## Summary
- Replace hardcoded 24-entry `ROWS` array with a `jq` mapfile that extracts the union of `publish[]`+`subscribe[]` across all active identities from the pre-computed `EFFECTIVE_JSON` (post `request_reply_flows` expansion), filters to `lyra.*` and `_inbox.*`, and sorts alphabetically
- Regenerate the spec sentinel block via `--update` to match the new table structure — 4 previously-missing subjects now covered: `lyra.clipool.control`, `_inbox.voice-client.>`, `_inbox.telegram-adapter.*.*`, `_inbox.discord-adapter.*.*`

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1039: fix(scripts): auto-derive ROWS in check-acl-matrix-spec.sh from acl-matrix.json | Open |
| Implementation | 1 commit on `feat/1039-auto-derive-rows` | Complete |
| Verification | Lint ✅ Tests ✅ (`bash check-acl-matrix-spec.sh` exit 0) | Passed |

## Test Plan
- [ ] `bash scripts/check-acl-matrix-spec.sh` exits 0 (no drift)
- [ ] Add a new subject to any active identity in `acl-matrix.json` → script renders it automatically (no ROWS edit needed)
- [ ] `$JS.API.>` and `$KV.*` subjects are excluded from the rendered table

Closes #1039

---


MickaelV0:
## Code Review

### Blockers

**issue: `pub_subject` always equals `sub_subject` in auto-derived rows — intentional pub/sub split silently dropped**
  `scripts/check-acl-matrix-spec.sh` (new `mapfile -t ROWS` block)
  -- devops
  Root cause: The jq query unions `publish[]` and `subscribe[]` into one set, then formats each subject as `` `s`|s|s ``. The old hardcoded rows deliberately split pub/sub subjects for request/reply pairs (e.g. display=`lyra.voice.tts.request`, pub_subject=`lyra.voice.tts.request.>`, sub_subject=`lyra.voice.tts.request`). Auto-derive loses this: hub's publish of `lyra.voice.tts.request.>` and voice-tts's subscribe of `lyra.voice.tts.request` now appear on two separate rows. `cell_value` for hub on the `lyra.voice.tts.request` row returns `—`, so the table no longer captures "hub publishes the wildcard that routes to voice-tts's exact subscription."
  Solutions:
    1. Keep auto-derive for discovery but retain an override map (`SUBJECT_OVERRIDES` associative array) for rows where pub_subject ≠ sub_subject; merge at render time. (recommended)
    2. Add a `subject_map` block to `acl-matrix.json` with `{ display, pub_subject, sub_subject }` triples; script reads it. Keeps intent in JSON rather than bash.
  Confidence: 92%

---

### Suggestions

**suggestion(non-blocking): newly-visible rows (`lyra.clipool.control`, `_inbox.*.*.` wildcards, `_inbox.voice-client.>`) need intent confirmation**
  `scripts/check-acl-matrix-spec.sh` (old lines 58–83 vs new `mapfile` block)
  -- devops
  Root cause: These subjects were absent from the old hardcoded ROWS. Auto-derive correctly surfaces them. The spec was regenerated with `--update`. Reviewers should confirm these rows are intentional inclusions in the published spec, not implementation noise.
  Solutions:
    1. Confirm in PR description these are intentional. No code change needed.
    2. If `lyra.clipool.control` or `*.*` inbox variants are internal only, add a `startswith` exclusion in the jq `select()` filter.
  Confidence: 85%

**issue(non-blocking): comment says "auto-derived from acl-matrix.json" but source is `$EFFECTIVE_JSON` (post-expansion)**
  `scripts/check-acl-matrix-spec.sh:47`
  -- devops
  Root cause: The comment is misleading — the mapfile reads `EFFECTIVE_JSON`, not the raw file. Correct today (all _inbox subjects already exist in static arrays), but a future flow whose _inbox subject is not pre-declared would appear in effective ACL without a row — the contract is fragile.
  Solutions:
    1. Update comment: "auto-derived from `$EFFECTIVE_JSON` (post request_reply_flows expansion)" (recommended)
    2. Add an assertion that effective subjects ⊇ raw subjects, failing with a diagnostic if expansion adds something not already declared.
  Confidence: 78%

**suggestion(non-blocking): JSON-controlled strings flow into unquoted bash array parse in `render_table`**
  `scripts/check-acl-matrix-spec.sh:154`
  -- security-auditor
  Root cause: Subjects from `acl-matrix.json` containing `|` or shell-special chars would corrupt the `${row%%|*}` / `${rest##*|}` parse, producing wrong cell assignments and silently incorrect diff output — a poisoned JSON entry could make a non-compliant matrix appear compliant.
  Solutions:
    1. Validate every derived subject matches `^[a-zA-Z0-9._>*-]+$` immediately after the `mapfile`, `exit 1` on mismatch. (recommended)
    2. Use NUL-delimited format (`jq -r0`) and `mapfile -d ''` so the delimiter cannot appear in data.
  Confidence: 72%

**suggestion(non-blocking): `IDENTITIES` values unquoted in markdown table header**
  `scripts/check-acl-matrix-spec.sh:137`
  -- security-auditor
  Root cause: Identity keys from `acl-matrix.json` appended to header/separator strings without validation. A key containing `|` would break the Markdown column count, making drift detection unreliable.
  Solutions:
    1. After `mapfile -t IDENTITIES`, assert each entry matches `^[a-zA-Z0-9_-]+$`, `exit 1` on violation. (recommended)
    2. Pass identities as jq `--argjson` into the render step to eliminate shell interpolation.
  Confidence: 68%

---

### Nitpicks

**nitpick: `IDENTITIES` reads raw `$JSON`, `ROWS` reads `$EFFECTIVE_JSON` — inconsistent sources**
  `scripts/check-acl-matrix-spec.sh:44`
  -- devops
  Root cause: No bug today (expansion only touches pub/sub arrays, not identity keys). Maintenance trap if `reduce` ever adds/removes an identity entry.
  Solutions:
    1. Change line 44 to `<<< "$EFFECTIVE_JSON"` for consistency.
    2. Add inline comment explaining why raw file is intentionally used here.
  Confidence: 70%

**nitpick: `EFFECTIVE_JSON` in shell variable — O(identities × rows) herestring jq forks**
  `scripts/check-acl-matrix-spec.sh:32`
  -- security-auditor
  Root cause: `json_has` spawns a jq subprocess per (identity × row) cell, each re-reading the full variable via herestring. With auto-derived ROWS, row count scales with JSON. In adversarial JSON (very large arrays), could cause resource exhaustion in CI.
  Solutions:
    1. Write `EFFECTIVE_JSON` to `$NATS_TMPDIR/effective.json` once; pass the path to all `jq` invocations. Reduces per-call overhead. (recommended)
    2. Pre-compute a lookup table keyed on `identity:key:subject` in one jq pass; eliminate per-cell forks entirely.
  Confidence: 62%

---

### Summary

| Finding | Category | File | Confidence |
|---------|----------|------|-----------|
| pub=sub split dropped | issue (blocking) | check-acl-matrix-spec.sh | 92% |
| New rows need intent confirmation | suggestion | check-acl-matrix-spec.sh | 85% |
| Misleading comment (EFFECTIVE_JSON) | issue (non-blocking) | check-acl-matrix-spec.sh | 78% |
| JSON subjects in unquoted bash parse | suggestion | check-acl-matrix-spec.sh | 72% |
| IDENTITIES/ROWS inconsistent sources | nitpick | check-acl-matrix-spec.sh | 70% |
| IDENTITIES unquoted in table header | suggestion | check-acl-matrix-spec.sh | 68% |
| EFFECTIVE_JSON herestring forks | nitpick | check-acl-matrix-spec.sh | 62% |

**Verdict: Request changes** — 1 blocker (pub/sub split regression at 92% confidence).

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 0 findings
(Blocker finding rejected by architect verifier at 90% — two-row representation is strictly more accurate than the old combined row)

**Applied via consensus + 1b1:** 5 findings (single-pass jq rewrite)
**Deferred:** 0
**Skipped:** 1 (new-rows intent confirmation — no code change, confirmed intentional)
**Failed:** 0

### Applied

All 5 remaining findings resolved by consensus-driven rewrite (`artifacts/analyses/1039-fix-scripts-auto-derive-rows-consensus.mdx`):

- [applied] `suggestion` — JSON subjects in unquoted bash parse (72%) → eliminated: JSON never enters bash string handling
- [applied] `issue(non-blocking)` — misleading comment "acl-matrix.json" vs `$EFFECTIVE_JSON` (78%) → comment updated
- [applied] `nitpick` — IDENTITIES reads raw `$JSON`, ROWS reads `$EFFECTIVE_JSON` (70%) → both from `$EFFECTIVE_JSON` now
- [applied] `suggestion` — IDENTITIES keys unquoted in table header (68%) → eliminated: header built in jq
- [applied] `nitpick` — O(I×R) herestring jq forks (62%) → 1 fork per `render_table()` call

### What changed

`mapfile -t IDENTITIES`, `mapfile -t ROWS`, `json_has()`, `cell_value()`, and the bash `render_table()` loop replaced with a single `jq -r '...'` invocation that derives identities, subjects, and computes all cells from `$EFFECTIVE_JSON`, outputting markdown directly.

Script: 179 → 90 lines. QG: `bash check-acl-matrix-spec.sh` exit 0 ✓

MickaelV0:
## Code Review (iteration 2)

**Iter-1 fixes verified:** all 5 previously-raised findings confirmed resolved — JSON data no longer enters bash string handling, IDENTITIES + ROWS from same source, O(1) jq forks. ✓

---

### Blockers

**issue: `reduce` on `request_reply_flows` appends `_inbox` grants once per matching flow — produces duplicate entries in `EFFECTIVE_JSON` arrays**
  `scripts/check-acl-matrix-spec.sh:29–35`
  -- devops
  Root cause: `hub` appears as requester in 5 flows, so `_inbox.hub.>` is appended 5× to its `subscribe` array. Current output is unaffected (`any()` + `unique` in subjects derivation are idempotent over duplicates), but `EFFECTIVE_JSON` is semantically incorrect — any future consumer sees bloated arrays without warning.
  Solutions:
    1. Apply `unique` inside the reduce after `+=`: `.identities[$flow.requester].subscribe |= (. + ["_inbox.\($flow.requester).>"]) | unique` (recommended)
    2. Add a post-reduce cleanup: `| .identities |= with_entries(.value.publish |= unique | .value.subscribe |= unique)` before render_table
  Confidence: 98%

**issue: `--update` silently truncates `$SPEC` when `acl-matrix:end` sentinel is absent**
  `scripts/check-acl-matrix-spec.sh:97–104`
  -- devops + security-auditor
  Root cause: awk sets `found=1` on begin but never resets if end is missing. All lines after begin are suppressed, `getline` never fires, and `cp` overwrites `$SPEC` with a truncated file — no error, no diff output.
  Solutions:
    1. Validate both markers exist before running update awk: `grep -q '<!-- acl-matrix:begin -->' "$SPEC" && grep -q '<!-- acl-matrix:end -->' "$SPEC"` — exit 1 with error if either missing. (recommended)
    2. Post-write validation: after awk, verify `spec_updated.txt` still contains both markers before `cp`; abort if not.
  Confidence: 97%

---

### Suggestions

**nitpick: `reduce` auto-vivifies a ghost identity object in `EFFECTIVE_JSON` if a flow references a responder not in `.identities`**
  `scripts/check-acl-matrix-spec.sh:30–34`
  -- devops
  Root cause: jq's `.identities[$key].publish += [...]` auto-vivifies `$key` with a partial object when absent. A typo in `request_reply_flows[].responder` silently injects a malformed entry into `EFFECTIVE_JSON`.
  Solutions:
    1. Guard inside reduce: add `select(.identities[$flow.requester] != null and .identities[$flow.responder] != null)` (recommended)
    2. Add a pre-flight cross-check of all flow requester/responder names against identity keys.
  Confidence: 88%

**suggestion(non-blocking): no pre-flight existence check for `$JSON` or `$SPEC`**
  `scripts/check-acl-matrix-spec.sh:25–26`
  -- devops
  Root cause: `set -euo pipefail` aborts on missing file, but raw jq/awk error messages give no context about which variable is misconfigured.
  Solutions:
    1. Add guards: `[[ -f "$JSON" ]] || { echo "::error::acl-matrix.json not found: $JSON"; exit 1; }` + same for `$SPEC`. (recommended)
    2. Add `set -x` behind a `DEBUG` flag for self-documenting CI output on failure.
  Confidence: 85%

**suggestion(non-blocking): non-atomic spec file update risks partial write on interruption**
  `scripts/check-acl-matrix-spec.sh:102`
  -- security-auditor
  Root cause: `cp` is not atomic — interruption during copy leaves `$SPEC` partially overwritten. `mv` on the same filesystem is a single `rename(2)` syscall.
  Solutions:
    1. Replace `cp` with `mv "${NATS_TMPDIR}/spec_updated.txt" "$SPEC"` (recommended)
    2. Keep `cp` + add `trap` restoring a pre-copy backup on ERR/INT.
  Confidence: 72%

**nitpick: `unique` already sorts — `| sort` after `unique` is redundant**
  `scripts/check-acl-matrix-spec.sh:53`
  -- devops
  Root cause: jq's `unique` is sort-then-dedup; output is already ascending order. `| sort` is a no-op that adds reader confusion.
  Solutions:
    1. Remove `| sort`: `flatten | unique | map(select(...))` (recommended)
    2. Replace with `sort | unique` to make intent explicit.
  Confidence: 100%

---

### Summary

| Finding | Category | Confidence | Agents |
|---------|----------|-----------|--------|
| Duplicate entries in EFFECTIVE_JSON | issue (blocking) | 98% | devops |
| --update silent truncation | issue (blocking) | 97% | devops + security-auditor |
| Ghost identity auto-vivification | nitpick | 88% | devops |
| No pre-flight file checks | suggestion | 85% | devops |
| Non-atomic spec write | suggestion | 72% | security-auditor |
| `sort` redundant after `unique` | nitpick | 100% | devops |

**Verdict: Request changes** — 2 blockers.

MickaelV0:

**Auto-applied:** 6 findings (all C≥72, user accepted all)
**1b1:** 0

### Applied

- [applied] `issue` — `reduce` duplicate inbox grants: `|= (. + [...] | unique)` inside reduce (98%)
- [applied] `issue` — `--update` silent truncation: sentinel marker validation before awk (97%)
- [applied] `nitpick` — ghost identity auto-vivification: `select()` guard on flow identities (88%)
- [applied] `suggestion` — no pre-flight file checks: `[[ -f ... ]]` guards for `$JSON` + `$SPEC` (85%)
- [applied] `suggestion` — non-atomic spec write: `cp` → `mv` (atomic rename) (72%)
- [applied] `nitpick` — redundant `| sort` after `unique` removed (100%)

Script: 115 lines (was 90 before this pass, 179 originally).
QG: `bash check-acl-matrix-spec.sh` exit 0 ✓

MickaelV0:
## Code Review — Iteration 3

**Agents:** devops, security-auditor
**Focus:** post iter-2 fix validation

---

### Findings

#### blocking — `select()` in `reduce` body terminates entire reduction (devops, 97%)

**File:** `scripts/check-acl-matrix-spec.sh:36–42`

`select(cond) |` inside a `reduce` body does not skip the current iteration — it returns empty, which causes the **entire reduction** to terminate immediately. If any `$flow` references an unknown identity, all subsequent flows are silently dropped.

Current code:
```bash
reduce (.request_reply_flows[]?) as $flow (
  .;
  select(.identities[$flow.requester] != null and .identities[$flow.responder] != null) |
  .identities[$flow.requester].subscribe |= ...
  .identities[$flow.responder].publish   |= ...
)
```

Required fix:
```bash
reduce (.request_reply_flows[]?) as $flow (
  .;
  if (.identities[$flow.requester] != null and .identities[$flow.responder] != null) then
    .identities[$flow.requester].subscribe |= (. + ["_inbox.\($flow.requester).>"] | unique) |
    .identities[$flow.responder].publish   |= (. + ["_inbox.\($flow.requester).>"] | unique)
  else
    .
  end
)
```

---

### Confirmed Clean (iter-2 fixes)

- Pre-flight `-f` guards: ✓
- `|= (. + [...] | unique)` grouping: ✓
- Sentinel validation before `--update`: ✓
- `mv` for atomic replacement: ✓
- Redundant `| sort` removed: ✓
- Single-pass jq `render_table()`: ✓

---

### Non-blocking

| # | Finding | Agent | C% |
|---|---------|-------|-----|
| 1 | `jq` exit code not checked in `EFFECTIVE_JSON=` assignment | security-auditor | 72% |
| 2 | CI guard uses `${CI:-}` — can be bypassed by setting `CI=false` in env | security-auditor | 68% |
| 3 | `awk` getline in `--update` path lacks error check | security-auditor | 65% |

---

### Status

**Iteration cap reached (3/3).** The `select()` bug is real and was introduced in iter-2. The non-blocking findings are cosmetic. Decision: **fix `select()` before merge** or **merge as-is** (risk: silent ACL grant loss if any flow references a future unknown identity).

PR #1072, opened 2026-05-05
chore(provision): warn on re-run if existing subuid range overlaps other users

### LINKED ISSUE #876: chore(provision): warn on re-run if existing subuid range overlaps other users
## Context

Deferred from PR #869 (issue #825) review — 3-expert panel consensus ([artifact](artifacts/analyses/825-subuid-review-followups-consensus.mdx)).

`has_sufficient_subids` in `deploy/provision.sh` gates on count (≥65536), not placement. If a prior provisioning run created a correctly-sized but badly-placed range (e.g. two users with `100000:65536` each due to pre-825 hardcoded band), the gate accepts it and overlap goes undetected on re-run.

## Why deferred

`has_sufficient_subids` is intentionally permissive to preserve idempotency. Panel: existing `podman info` smoke test downstream in `provision.sh` surfaces rootless misconfigurations at runtime on the single-admin `roxabituwer` machine. No need to fail-fast here today.

After the gate passes (user already has a range), emit a warn-only scan:

```bash
# inside the 'else' branch of the has_sufficient_subids check
admin_range=\$(awk -F: -v u=\"\$ADMIN_USER\" '\$1 == u {print \$2 \" \" \$2+\$3}' /etc/subuid)
overlapping=\$(awk -F: -v u=\"\$ADMIN_USER\" -v range=\"\$admin_range\" '...' /etc/subuid)
[[ -n \"\$overlapping\" ]] && warn \"subuid range for \$ADMIN_USER overlaps: \$overlapping\"
```

## Revisit when

- `AGENT_USER` subuid provisioning is added to `provision.sh`.
- More than one human user on the host requires rootless containers.

## Links

- PR #869
- Issue #825

MickaelV0:
## Summary
- Add `warn_subid_overlap()` helper that scans `/etc/subuid` and `/etc/subgid` for ranges belonging to other users that overlap the admin user's existing allocation
- Call it in the idempotent re-run path (the `else` branch of `has_sufficient_subids`) so operators are warned on repeat runs without breaking idempotency

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #876: chore(provision): warn on re-run if existing subuid range overlaps other users | Open |
| Analysis | Absent | — |
| Spec | Absent | — |
| Implementation | 1 commit on `feat/876-provision-warn-subuid-overlap` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (0 new) | Passed |

## Test Plan
- [ ] Re-run `provision.sh` on a host where two users share `100000:65536` — confirm warning is emitted and script continues (no error)
- [ ] Re-run on a host with no overlaps — confirm no warning, info line only
- [ ] Run on a fresh host (no subuid entries) — no warning, allocation proceeds normally via the `if` branch

Closes #876

---

### DIFF
--- a/artifacts/analyses/876-provision-warn-subuid-overlap-consensus.mdx
+++ b/artifacts/analyses/876-provision-warn-subuid-overlap-consensus.mdx
@@ -1,0 +1,76 @@
+---
+title: "provision.sh warn_subid_overlap — Long-Term Defensive Posture Consensus"
+issue: 876
+status: consensus-reached
+date: 2026-05-05
+panel: architect, devops, security-auditor
+confidence: high
+---
+
+## Problem
+
+Three review findings (F2, F5, F6) on PR #1072 remained after auto-apply. Panel was asked for long-term recommendations on all three pending findings in the context of `warn_subid_overlap` and `assert_no_subid_overlap` in `deploy/provision.sh`.
+
+## Panel
+
+| Agent | Focus |
+|-------|-------|
+| architect | Arch soundness, contract parity, maintainability |
+| devops | Ops observability, curl\|bash run visibility, idempotency |
+| security-auditor | Defense-in-depth, path traversal, privilege boundary |
+
+## Consensus ρ
+
+Apply all three fixes:
+
+**F2** — Split the combined missing/unreadable guard in `warn_subid_overlap`:
+```bash
+[[ ! -e "$file" ]] && return 0
+[[ ! -r "$file" ]] && { warn "Cannot read $file (check permissions) — skipping overlap check"; return 0; }
+```
+
+**F5** — Add allowlist guard at function entry in `warn_subid_overlap`:
+```bash
+[[ "$file" == /etc/subuid || "$file" == /etc/subgid ]] \
+  || { warn "warn_subid_overlap: unexpected file path '$file' — skipping"; return 0; }
+```
+
+**F6** — Add blank line between `warn_subid_overlap` and `assert_no_subid_overlap`.
+
+**Contract rule to document:** Advisory-path functions (`warn_*`) must distinguish missing (silent return 0) from unreadable (warn + return 0). This mirrors the hard-path (`assert_*`) pattern: missing = skip, unreadable = abort.
+
+### Rationale
+
+All three experts converged on the same solutions without debate:
+- F2: Missing vs unreadable distinction is an observability fix, not a security fix — the devops and arch perspectives both identified this as the right long-term posture for a `curl | bash` provisioner where operator signal matters
+- F5: Allowlist (not regex) because the permitted set is exactly two paths; exact match detects future drift loudly
+- F6: Blank line is unconditionally correct; security-auditor additionally recommends `shfmt` in CI as a systemic fix
+
+### Trade-offs
+
+- F2 warn on unreadable adds noise on permission-hardened hosts — accepted cost; advisory, not hard failure
+- F5 whitelist must be updated if distros relocate subid files — acceptable visible maintenance cost
+- F6 blank line: consider adding `shfmt` to the CI lint gate to prevent recurrence of shell style issues
+
+## Alternatives
+
+| Option | Proposed by | Rejected because |
+|--------|-------------|------------------|
+| F2: hard error on unreadable | — | Contradicts `warn_subid_overlap`'s advisory contract; `assert_no_subid_overlap` owns the abort path |
+| F2: silent skip with comment (status quo) | security-auditor (initial review) | Hides permission regressions in curl\|bash runs permanently |
+| F5: regex path guard (`^/etc/`) | — | Too broad; allows `/etc/passwd`, `/etc/shadow` to reach awk |
+| F5: no guard (status quo) | — | Implicit contract; breaks as script accumulates callers over time |
+| Shared file-access helper | architect | Over-engineering for single-file bash provisioner; bash lacks structured return types |
+
+## Dissent
+
+None — unanimous.
+
+## Implementation Notes
+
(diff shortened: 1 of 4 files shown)


MickaelV0:
## Code Review

### Blockers

**issue**: `local end=$(( start + count ))` — `local` masks arithmetic exit code
  `deploy/provision.sh:134`
  -- devops
  Root cause: In bash, `local var=$(...)` always exits 0 regardless of what the subshell returns. Under `set -euo pipefail` this means an arithmetic failure (e.g. non-integer `count`) would be silently swallowed rather than aborting the script.
  Solutions:
    1. Separate declaration from assignment: `local end; end=$(( start + count ))` (recommended)
    2. Leave as-is and add explicit `count` validation above so the arithmetic can't fail.
  Confidence: 92%

---

**issue**: Unreadable file silently returns 0 with no log output
  `deploy/provision.sh:130`
  -- devops
  Root cause: `[[ ! -e "$file" || ! -r "$file" ]] && return 0` treats a missing file and an unreadable file identically. On a permission regression (`chmod 600 /etc/subuid`), the overlap check is skipped with zero diagnostic output — invisible in a `curl | bash` run.
  Solutions:
    1. Split the two cases: `[[ ! -e "$file" ]] && return 0; [[ ! -r "$file" ]] && { warn "Cannot read $file — skipping overlap check"; return 0; }` (recommended)
    2. Add a comment documenting the silent-skip intent so reviewers don't accidentally "fix" it.
  Confidence: 90%

---

**issue**: `count` unvalidated before arithmetic — empty count silently produces zero-width range
  `deploy/provision.sh:132-134`
  -- devops
  Root cause: `read -r start count` takes both fields from awk without validating `count`. A truncated or manually edited entry like `user:100000:` sets `count=""`, making `end=$(( start + 0 )) = start`. The overlap predicate `$2 + $3 > s && $2 < e` then never triggers, silently skipping the check with no diagnostic. `start` is guarded by `[[ -z "$start" ]] && return 0` but `count` has no equivalent guard.
  Solutions:
    1. Add `[[ -z "$count" || ! "$count" =~ ^[0-9]+$ ]] && { warn "malformed subid entry for $ADMIN_USER in $file — skipping overlap check"; return 0; }` after line 133 (recommended)
    2. Merge into a single awk expression that validates field count before printing, falling back to an empty string on malformed input.
  Confidence: 88%

---

**suggestion(non-blocking)**: `assert_no_subid_overlap` awk lacks `$1 != u` self-filter — pre-existing gap
  `deploy/provision.sh:148-149`
  -- devops
  Root cause: `warn_subid_overlap` correctly filters out the admin user's own entry (`$1 != u`), but the sibling `assert_no_subid_overlap` at line 148 uses no such filter. When called during first-run allocation, `$ADMIN_USER`'s entry doesn't exist yet so this is a non-issue today — but if a re-provisioning scenario adds a second range before calling `assert_no_subid_overlap`, the function would flag the user's own existing range as an overlap.
  Solutions:
    1. Add `-v u="$ADMIN_USER"` and `$1 != u &&` to the `assert_no_subid_overlap` awk (recommended — consistent pattern, closes latent self-overlap false positive)
    2. Accept current behavior with a comment explaining the call-site ordering guarantee.
  Confidence: 85%

**suggestion(non-blocking)**: `$file` arg unvalidated in `warn_subid_overlap`
  `deploy/provision.sh:128-130`
  -- security-auditor
  Root cause: The function accepts `$1` as `file` and passes it directly to `awk`. Current call sites pass hardcoded literals so there is no injection path today, but no guard prevents a future refactor from passing a user-controlled value.
  Solutions:
    1. Document as internal-only with a comment noting the hardcoded-caller assumption (recommended — low overhead)
    2. Add a caller-side assertion: `[[ "$file" == /etc/subuid || "$file" == /etc/subgid ]]`
  Confidence: 60%

**nitpick**: Missing blank line between `warn_subid_overlap` and `assert_no_subid_overlap`
  `deploy/provision.sh:142`
  -- devops
  Root cause: The two functions share no blank line separator, making the function boundary visually ambiguous.
  Solutions:
    1. Add a blank line after the closing `}` of `warn_subid_overlap` (recommended)
  Confidence: 75%

---

### Summary

3 blocking issues (C≥88%) around defensive guards in `warn_subid_overlap`: `local`/exit-code masking, unreadable-file silent skip, and unvalidated `count`. All are 2-line fixes. 1 pre-existing gap in `assert_no_subid_overlap` surfaced as a non-blocking suggestion.

**Verdict: Request changes**

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 3 finding(s)
**Applied via 1b1 + consensus:** 3 finding(s)
**Deferred:** 0
**Skipped:** 0
**Failed:** 0

### Auto-Applied (commit 68b4df42)
- [applied] issue: separate `local end; end=$((...))` — propagate exit under set -e (C=92%)
- [applied] issue: validate `count` before arithmetic — guard against zero-width range (C=88%)
- [applied] suggestion: add `$1 != u` self-filter to `assert_no_subid_overlap` (C=85%)

### Applied via 1b1 + consensus panel (commit 13fd0561)
- [applied] F2: split missing vs unreadable guard — emit `warn` on unreadable file, silent on missing
- [applied] F5: add allowlist guard (`/etc/subuid`|`/etc/subgid`) at `warn_subid_overlap` entry
- [applied] F6: blank line between `warn_subid_overlap` and `assert_no_subid_overlap`

### Consensus
3-expert panel (architect + devops + security-auditor) reached unanimous consensus on all 3 pending findings. Key outcome: established contract rule — `warn_*` functions distinguish missing (silent) from unreadable (warn + skip), mirroring the `assert_*` hard path. Documented in [ADR-069](docs/architecture/adr/069-provision-warn-subid-overlap-defensive-posture.mdx).

### Follow-up noted (not in this PR)
- Add `shfmt` to CI lint gate for shell files (security-auditor recommendation)

PR #1127, opened 2026-05-07
ci(quadlet-lint): pin plucky podman/conmon/crun versions for reproducibility

### LINKED ISSUE #1120: ci(quadlet-lint): pin plucky podman/conmon/crun versions for reproducibility
Discovered during code review of PR #1116 (Podman 5.x install for quadlet-lint).

## Issue

`.github/workflows/quadlet-lint.yml:44-45` installs:

```yaml
sudo apt-get install -y --no-install-recommends \
  -t plucky podman conmon crun
```

No version pin → tracks plucky's current version. Risks:
- Ubuntu pushes a broken podman update to plucky → all CI runs break with no diff.
- Supply-chain hygiene gap (no exact version reproducibility).
- Latent: a future plucky podman bump that adds new hard deps with higher-soname than noble's may silently broaden the plucky footprint via apt's transitive resolution (pin=100 is bypassed when noble doesn't satisfy a versioned Depends).

## Source

- security-auditor finding (C=78%) — supply-chain hygiene
- devops finding (C=78%) — non-blocking, same concern
- security-auditor follow-up: `thought` C=72% on transitive deps

## Fix options

**Option 1 (recommended):** pin exact versions
```yaml
sudo apt-get install -y --no-install-recommends \
  -t plucky 'podman=5.4.1+ds1-1' 'conmon=...' 'crun=...'
```
Hard-fail if apt resolves a different version. Capture current versions via `apt-cache policy` once, commit as constants, update intentionally.

**Option 2:** add `apt-get install --dry-run` pre-check that logs resolved versions to CI output → audit trail without hard pin.

**Option 3 (defensive):** after install, assert `dpkg -l | grep plucky` only contains {podman, conmon, crun} — fail step if any other plucky package leaked in.

Recommend Option 1 + Option 3 combined.

## Mitigations already in place

- `set -euo pipefail` + `MAJOR>=5` version assertion (catches major downgrade)
- Pin-Priority=100 (prevents unrelated package upgrades — but bypassed for transitive deps)
- Ephemeral runner (no persistent state to compromise)

So this is **hygiene/reproducibility**, not active vulnerability.

MickaelV0:
Closes #1120.

## Summary

- Pins `podman=5.4.1+ds1-1`, `conmon=2.1.12-4`, `crun=1.20-1syncable1` in the `quadlet-lint` CI step (versions fetched from plucky universe archive 2026-05-07)
- Adds post-install assertion (Option 3 from issue) that fails CI if any unexpected package from the plucky source was pulled in transitively
- Source of truth: `archive.ubuntu.com/ubuntu/dists/plucky/universe/binary-amd64/Packages.gz`

## Test plan

- [ ] CI `quadlet-lint` workflow runs green on this PR
- [ ] Confirm installed versions match pins in CI log (`Installed Podman 5.4.1`)
- [ ] Plucky-leak assertion produces no output (no extra packages)

### DIFF
--- a/.github/workflows/quadlet-lint.yml
+++ b/.github/workflows/quadlet-lint.yml
@@ -49,11 +49,32 @@ jobs:
             'conmon=2.1.12-4' \
             'crun=1.20-1syncable1'
           # Assert no other plucky packages leaked in via transitive deps.
-          PLUCKY_EXTRA=$(dpkg-query -W -f='${Package} ${Status}\n' \
-            | awk '{print $1}' \
-            | xargs -I{} apt-cache policy {} 2>/dev/null \
-            | grep -B20 'Installed:' \
-            | awk '/^[^ ]/{pkg=$1} /plucky/ && /\*\*\*/{print pkg}' \
+          # Gather full apt-cache policy output in one batch call (no -I{} to
+          # avoid one subprocess per package; xargs batches all names at once).
+          RAW=$(dpkg-query -W -f='${Package}\n' \
+            | xargs apt-cache policy 2>/dev/null)
+          # Stateful awk over the full stream — no grep -B context window needed.
+          # apt-cache policy format (exact indentation matters):
+          #   podman:                          ← package header (no leading whitespace, ends in ':')
+          #     Installed: 5.4.1+ds1-1
+          #     Version table:
+          #    *** 5.4.1+ds1-1 100            ← active candidate (1 leading space + ***)
+          #           100 https://.../plucky  ← archive source (8 leading spaces)
+          #        4.9.0 500                  ← non-active version (5 leading spaces + non-space)
+          #           500 https://.../noble
+          # Rules:
+          #   - New package: line with no leading whitespace ending in ':'  → reset state
+          #   - Active candidate: " *** ..."                                → set in_cand=1
+          #   - Non-active version: "     <non-space>..."                   → set in_cand=0
+          #   - Archive line while in_cand && contains "plucky"             → print pkg (once)
+          PLUCKY_PKGS=$(printf '%s\n' "${RAW}" \
+            | awk '/^[^ \t].*:$/ { pkg = $0; sub(/:$/, "", pkg); in_cand = 0; next }
+                   /^ \*\*\*/     { in_cand = 1; next }
+                   /^     [^ ]/   { in_cand = 0; next }
+                   in_cand && /plucky/ { print pkg; in_cand = 0 }')
+          # grep -Fxv exits 1 when all lines are excluded (list is clean) — that
+          # is the expected happy path, so || true is scoped only to this step.
+          PLUCKY_EXTRA=$(printf '%s\n' "${PLUCKY_PKGS}" \
             | grep -Fxv -e podman -e conmon -e crun || true)
           if [[ -n "${PLUCKY_EXTRA}" ]]; then
             echo "::error::Unexpected plucky packages installed: ${PLUCKY_EXTRA}"
--- a/.github/workflows/quadlet-lint.yml
+++ b/.github/workflows/quadlet-lint.yml
@@ -40,9 +40,25 @@ jobs:
           printf 'Package: *\nPin: release n=plucky\nPin-Priority: 100\n' \
             | sudo tee /etc/apt/preferences.d/plucky-pin
           sudo apt-get update -qq
-          # Install Podman and its direct runtime deps from plucky explicitly.
+          # Install Podman and its direct runtime deps from plucky with pinned
+          # versions for reproducibility (captured from plucky archive 2026-05-07).
+          # Update intentionally when upgrading: apt-cache policy podman conmon crun
           sudo apt-get install -y --no-install-recommends \
-            -t plucky podman conmon crun
+            -t plucky \
+            'podman=5.4.1+ds1-1' \
+            'conmon=2.1.12-4' \
+            'crun=1.20-1syncable1'
+          # Assert no other plucky packages leaked in via transitive deps.
+          PLUCKY_EXTRA=$(dpkg-query -W -f='${Package} ${Status}\n' \
+            | awk '{print $1}' \
+            | xargs -I{} apt-cache policy {} 2>/dev/null \
+            | grep -B20 'Installed:' \
+            | awk '/^[^ ]/{pkg=$1} /plucky/ && /\*\*\*/{print pkg}' \
+            | grep -Fxv -e podman -e conmon -e crun || true)
+          if [[ -n "${PLUCKY_EXTRA}" ]]; then
+            echo "::error::Unexpected plucky packages installed: ${PLUCKY_EXTRA}"
+            exit 1
+          fi
           # Verify we have Podman ≥ 5.x before proceeding.
           PODMAN_VERSION=$(podman --version | awk '{print $3}')
           echo "Installed Podman ${PODMAN_VERSION}"


MickaelV0:
## Code Review

**File:** `.github/workflows/quadlet-lint.yml`
**PR:** #1127 — `chore(ci): pin plucky podman/conmon/crun versions in quadlet-lint workflow`

---

### Blockers

**blocker: `|| true` neuters the footprint guard — Option 3 silently passes on `grep` failure**

`.github/workflows/quadlet-lint.yml` lines 25–29 (diff +25)

-- devops
Root cause: The final `grep -Fxv … || true` suppresses the exit code of the entire pipeline. If `grep` exits 1 because _no_ extra packages exist (the happy path), `|| true` is correct. But if any upstream command in the pipe (`dpkg-query`, `awk`, `xargs apt-cache policy`, the first `grep -B20`) fails or emits nothing, the variable is empty and the guard passes silently. Specifically: `xargs -I{} apt-cache policy {} 2>/dev/null` — if `apt-cache` binary is missing or all packages return empty policy output, the `grep -B20 'Installed:'` match returns 1, which makes the whole pipe exit 1, and `|| true` swallows it. The intent is only to suppress `grep -Fxv` exit 1, but `|| true` applies to the entire subshell expression.
Class: [bash-error-suppression, missing-error-handling]
Solutions:
  1. (recommended) Split the pipeline: run the dpkg enumeration in a separate variable, then pipe into the filter with `|| true` applied only to the final `grep -Fxv` step:
     ```bash
     ALL_PKGS=$(dpkg-query -W -f='${Package}\n')
     PLUCKY_EXTRA=$(printf '%s\n' "${ALL_PKGS}" \
       | xargs -I{} apt-cache policy {} 2>/dev/null \
       | awk '/^[^ ]/{pkg=$1} /plucky/ && /\*\*\*/{print pkg}' \
       | grep -Fxv -e podman -e conmon -e crun || true)
     ```
     This keeps `set -euo pipefail` effective on the data-gathering steps while still tolerating `grep` returning 1 when the list is clean.
  2. Add `pipefail` protection by checking `PLUCKY_EXTRA` as an empty-string test but also asserting the enumeration step itself succeeded via a separate `||` trap.
Confidence: 92%

---

**blocker: `grep -B20` context window is fragile — misattributes package names under policy output churn**

`.github/workflows/quadlet-lint.yml` lines 22–24 (diff +22–24)

-- devops
Root cause: `grep -B20 'Installed:'` emits up to 20 preceding lines per match. The `awk '/^[^ ]/{pkg=$1} /plucky/ && /\*\*\*/{print pkg}'` pattern relies on the package-name header appearing within those 20 lines. `apt-cache policy` output for packages with many candidate versions (e.g. `libglib2.0-0` pulled as a transitive dep) can produce > 20 lines before `Installed:`, causing the `awk` package-name tracker to lag and attribute the wrong package name to the plucky `***` marker. This produces both false positives (block on non-plucky-installed packages) and false negatives (miss a real leak).
Class: [bash-arithmetic-trap, missing-input-validation]
Solutions:
  1. (recommended) Enumerate installed packages directly from dpkg's policy output in a single pass using `apt-cache policy $(dpkg-query -W -f='${Package} ')` and parse with a stateful awk that resets on each package header:
     ```bash
     awk '/^[^ \t]/{pkg=substr($0,1,index($0,":")-1)} \
          /plucky/ && /\*\*\*/{print pkg}'
     ```
     And remove the `grep -B20` step entirely — let awk track state across the full stream without a line-count window.
  2. Use `apt-mark showmanual` restricted to the plucky source instead of post-processing policy output.
Confidence: 85%

---

### Warnings

**warning: no apt cache key — runner cache could serve stale package index masking version unavailability**

`.github/workflows/quadlet-lint.yml` — install step (no caching step present)

-- devops
Root cause: `sudo apt-get update -qq` fetches the package index on every run, but GitHub Actions runners are ephemeral and do not cache apt state between runs. This is actually correct behaviour for reproducibility. However, if a `cache: apt` action were added in future (common pattern), it could serve a cached index from before the plucky source was configured, making the pinned versions appear unavailable and failing loudly. There is no caching directive here today, so this is a forward-looking warning rather than a current defect.
Class: []
Solutions:
  1. Add a comment explicitly stating that apt-get update must remain uncached for this step to prevent future contributors from adding an apt cache action that would break version resolution.
  2. No change required if team policy is against apt caching in CI.
Confidence: 75%

**warning: `crun=1.20-1syncable1` version string uses a non-standard Debian epoch suffix (`syncable1`) — worth documenting**

`.github/workflows/quadlet-lint.yml` line 18 (diff +18)

-- devops
Root cause: The `syncable1` suffix is a Ubuntu sync/merge artifact indicating this package was synced from Debian with a local revision. This is valid Debian version syntax and `apt-get install` will match it exactly. However, if Ubuntu re-uploads with a `syncable2` suffix (common on point updates), the pin will fail with `E: Version '1.20-1syncable1' for 'crun' was not found`. The failure will be loud (apt exits non-zero), but the error message may not immediately convey that a simple version bump is needed.
Class: []
Solutions:
  1. (recommended) Add an inline comment on the `crun` pin line noting the `syncable1` suffix semantics and that `apt-cache policy crun` should be checked when this fails.
  2. Pin to `crun=1.20-1` without the `syncable` suffix if apt allows partial version matching (it does not — `=` requires an exact version string match, so this alternative does not work; stay with option 1).
Confidence: 80%

**warning: plucky archive removal of exact pinned versions will fail loudly but with an unhelpful message**

`.github/workflows/quadlet-lint.yml` — install step

-- devops
Root cause: When Ubuntu removes a version from the plucky archive (e.g. superseded by a security update), `apt-get install podman=5.4.1+ds1-1` will produce `E: Version '5.4.1+ds1-1' for 'podman' was not found` and exit non-zero. CI will fail loudly, which is the correct behaviour (no silent regression). The failure message is unhelpful without context — a maintainer unfamiliar with the pin strategy will not immediately know to update the versions.
Class: []
Solutions:
  1. Add a comment block above the install step pointing to the update procedure: `apt-cache policy podman conmon crun` on a plucky host (or container) to get the current available version, then update the three pins.
  2. Add a `||` handler that emits a `::error::` annotation with guidance text when the install fails.
Confidence: 88%

---

### Suggestions

**suggestion: `xargs -I{}` over full package list is O(n) apt-cache calls — slow on runners with many installed packages**

`.github/workflows/quadlet-lint.yml` lines 21–24 (diff +21–24)

-- devops
Root cause: `dpkg-query -W` on a GitHub Actions ubuntu-latest runner returns ~500–700 packages. `xargs -I{}` spawns one `apt-cache policy` process per package (because `-I{}` disables batching). This is 500–700 subprocesses. Using `xargs` without `-I{}` batches packages into fewer calls, or `apt-cache policy` accepts multiple package names in a single invocation.
Solutions:
  1. (recommended) Remove `-I{}` and pass all packages in one call: `dpkg-query -W -f='${Package}\n' | xargs apt-cache policy 2>/dev/null`
  2. Scope the check: instead of enumerating all installed packages, only check the known transitive deps of podman/conmon/crun by listing them with `apt-cache depends --recurse`.
Confidence: 80%

**suggestion: the version-pinning comment says "Update intentionally when upgrading" but gives no update procedure**

`.github/workflows/quadlet-lint.yml` lines 11–12 (diff +11–12)

-- devops
Solutions:
  1. Expand the comment to specify the update command: `apt-cache policy podman conmon crun` (run on a plucky host or `docker run ubuntu:25.04`) and note where in the workflow the three version strings appear.
Confidence: 70%

---

### Praise

- Option 1 (version pinning) is correctly implemented: all three packages use exact Debian version strings with `=` (not `>=`), and `-t plucky` ensures apt resolves from the correct source.
- The post-install Podman version gate (`MAJOR -lt 5` check) provides a second loud failure mode independent of the install step — good defence in depth.
- `set -euo pipefail` is present and the `|| true` is targeted (not a blanket `set +e`), showing awareness of the issue; the blocker above is about scope, not absence of care.
- The `::error::` annotation on the footprint guard will surface in the GitHub UI if triggered — correct annotation format.
- No apt caching action was added, keeping the install step reproducible.

---

### Verdict

**Request changes** — 2 blockers, 3 warnings, 2 suggestions.

The `|| true` on the full pipeline subshell (blocker 1) means the footprint guard (Option 3) can silently pass when the enumeration pipeline itself errors. The `grep -B20` context window (blocker 2) can produce wrong package names and both false positives and false negatives. Both must be fixed before merge.

MickaelV0:
## Re-review (after fix — commit aab0ad46)

**Blocker 1 — `|| true` scoped only to `grep -Fxv`** ✅ resolved
`grep -n '|| true'` returns 1 result (line 78): the `grep -Fxv -e podman -e conmon -e crun || true` subshell. The outer pipeline (`$PLUCKY_EXTRA` assignment, `if [[ -n ... ]]` guard, `exit 1`) is fully live. Option 3 is not neutered.

**Blocker 2 — `grep -B20` removed; replaced with stateful awk** ✅ resolved
`grep -B` no longer appears in the workflow (only a comment noting it was removed). A 4-rule stateful awk over the full `apt-cache policy` stream handles package attribution:
- `/^[^ \t].*:$/` → new package header, reset state
- `/^ \*\*\*/` → active candidate, `in_cand=1`
- `/^     [^ ]/` → non-active version, `in_cand=0`
- `in_cand && /plucky/` → print pkg, reset

**Awk fixture sanity-test passed** (30-line heredoc):
- `podman` with `***` plucky → printed ✅
- `containerd` with `***` noble (plucky non-active) → NOT printed ✅
- `runc` with `***` noble (plucky non-active) → NOT printed ✅

Both blockers addressed. Approving for merge.

Issue #1125, opened 2026-05-07
harden(deploy): nats/setup.sh — validate LYRA_USER from SUDO_USER + use printf for .env append


MickaelV0:
Pre-existing finding from PR #1123 code review (security-auditor, C=72%).

## Issue

`deploy/nats/setup.sh:179-196`:

```bash
LYRA_USER="${SUDO_USER:-$(id -un)}"
LYRA_HOME=$(getent passwd "$LYRA_USER" | cut -d: -f6)
HUB_SEED="${LYRA_HOME}/.lyra/nkeys/hub.creds"
...
echo "NATS_NKEY_SEED_PATH=${HUB_SEED}" >> "${LYRA_DIR}/.env"
```

`SUDO_USER` is attacker-controllable in scenarios where the script is invoked through a sudo wrapper that preserves environment (e.g. `sudo -E`, sudoers `env_keep`). Malicious payload like:

```bash
SUDO_USER=$'foo\nNATS_NKEY_SEED_PATH=/attacker/path'
```

Injects an extra line into `.env` at append time.

## Class

- `shell-injection`
- `missing-input-validation`

## Fix options

**Option 1 (recommended):**
```bash
LYRA_USER="${SUDO_USER:-$(id -un)}"
[[ "$LYRA_USER" =~ ^[a-z_][a-z0-9_-]*$ ]] || { echo "[!] Invalid LYRA_USER: $LYRA_USER"; exit 1; }
...
printf 'NATS_NKEY_SEED_PATH=%s\n' "${HUB_SEED}" >> "${LYRA_DIR}/.env"
```

**Option 2:** drop `SUDO_USER` heuristic entirely — derive from repo owner: `LYRA_USER=$(stat -c '%U' "${LYRA_DIR}")` and validate.

## Mitigations already in place

- Script refuses root invocation
- `set -euo pipefail` will catch some downstream failures

But the `.env` injection vector bypasses these — `echo` succeeds, the malicious path lands.

Source: PR #1123 code review (https://github.com/Roxabi/lyra/pull/1123)

PR #1128, opened 2026-05-07
refactor(deploy): extract XDG_RUNTIME_DIR guard to deploy/lib/env.sh

### LINKED ISSUE #1126: refactor(deploy): extract XDG_RUNTIME_DIR guard to deploy/lib/env.sh + propagate via Makefile SSH
Bundled finding from PR #1123 code review (devops, C=85% drift + 72% Makefile gap).

## Issue

The `export XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-/run/user/$(id -u)}"` pattern + 4-line preceding comment is now copy-pasted across **4 deploy scripts**:

| File | Line |
|---|---|
| `deploy/nats/setup.sh` | 25 |
| `deploy/quadlet-install-verify.sh` | 14 |
| `deploy/scripts/rotate-gh-key.sh` | 26 |
| `deploy/scripts/rotate-claude-oauth.sh` | 25 |

Plus a **latent gap** in the Makefile SSH heredocs:

| File | Line |
|---|---|
| `Makefile` | 207-224 (`deploy` target) |
| `Makefile` | 226-243 (`full-deploy` target) |

Current code path is covered (verify scripts self-export at top), but any future recipe added to `quadlet-install` that calls `systemctl --user` before forking to verify script will fail in non-interactive SSH sessions.

## Class

- `parallel-path-drift` (4 copies of identical pattern)
- `candidate/env-propagation-gap` (Makefile SSH heredoc)

## Fix options

**Option 1 (recommended):** consolidate to `deploy/lib/env.sh`
```bash
# deploy/lib/env.sh
export XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-/run/user/$(id -u)}"
```

Then in each script:
```bash
source "$(dirname "$0")/../lib/env.sh"  # path-adjusted per script
```

And in Makefile SSH heredocs:
```makefile
ssh ... <<'EOF'
  export XDG_RUNTIME_DIR="/run/user/$(id -u)"
  cd "$LYRA_DIR" && make quadlet-install
EOF
```

Collapses 4 callsites to 1 + closes the Makefile gap.

**Option 2:** leave as-is, document pattern in `deploy/README`.

## Source

- PR #1123 code review (devops): https://github.com/Roxabi/lyra/pull/1123
- Refs: #1111, #1115 (rotate-gh-key XDG fix), #1106 (rotate-claude-oauth original)

MickaelV0:
Closes #1126

## Summary

- Creates `deploy/lib/env.sh` with a single idempotent `XDG_RUNTIME_DIR` guard (`:=` no-op pattern, safe to source multiple times)
- Replaces the 4-copy inline export in `setup.sh`, `quadlet-install-verify.sh`, `rotate-gh-key.sh`, `rotate-claude-oauth.sh` with `source "$(dirname "$0")/[../]lib/env.sh"`
- Adds `export XDG_RUNTIME_DIR="/run/user/$(id -u)"` to the Makefile `deploy` and `full-deploy` SSH heredocs, closing the latent env-propagation gap flagged in the issue

Pure refactor — no behaviour change.

## Test plan

- [ ] `bash -n` clean on all 5 modified files
- [ ] `bash -c '. deploy/lib/env.sh && env | grep XDG_RUNTIME_DIR'` → `XDG_RUNTIME_DIR=/run/user/1000`
- [ ] Idempotent double-source → exactly 1 `XDG_RUNTIME_DIR` line
- [ ] `make -n deploy` shows `export XDG_RUNTIME_DIR="/run/user/$(id -u)"` in expanded output
- [ ] All pre-commit hooks pass (lint, typecheck, file-length, folder-size, duplicate-test-basenames, import-layers, trufflehog, license)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,6 +206,7 @@ deploy:
 	@echo "Deploying quadlet units to $(DEPLOY_HOST)..."
 	@ssh $(DEPLOY_HOST) '\
 	set -eu; \
+	export XDG_RUNTIME_DIR="/run/user/$$(id -u)"; \
 	LYRA_DIR=$(DEPLOY_DIR); \
 	VOICE_DIR=$$(grep "^VOICE_DEPLOY_DIR=" "$$LYRA_DIR/.env" 2>/dev/null | cut -d= -f2); \
 	VOICE_DIR=$${VOICE_DIR:-$$HOME/projects/voiceCLI}; \
@@ -228,6 +229,7 @@ full-deploy:  ## atomic deploy: git pull → quadlet-install → regen auth.conf
 	@echo "Full deploy to $(DEPLOY_HOST)..."
 	@ssh $(DEPLOY_HOST) '\
 	set -eu; \
+	export XDG_RUNTIME_DIR="/run/user/$$(id -u)"; \
 	LYRA_DIR=$(DEPLOY_DIR); \
 	VOICE_DIR=$$(grep "^VOICE_DEPLOY_DIR=" "$$LYRA_DIR/.env" 2>/dev/null | cut -d= -f2); \
 	VOICE_DIR=$${VOICE_DIR:-$$HOME/projects/voiceCLI}; \
--- a/deploy/lib/env.sh
+++ b/deploy/lib/env.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+# deploy/lib/env.sh — shared environment guards for deploy scripts
+#
+# Source this file at the top of any deploy script that calls `systemctl --user`
+# or `podman` via a non-interactive SSH session (e.g., `make remote`).
+#
+# Safe to source multiple times — all assignments use the :- no-op pattern.
+#
+# Usage (from a script in deploy/ or deploy/nats/):
+#   source "$(dirname "$0")/../lib/env.sh"
+#
+# Usage (from a script in deploy/scripts/):
+#   source "$(dirname "$0")/../lib/env.sh"
+
+# Required when invoked via `make remote` (SSH non-interactive shell): without
+# it, `systemctl --user` fails to locate the dbus session ("Failed to connect
+# to bus: No such file or directory"). Provision.sh sets this consistently;
+# all deploy scripts must too.
+: "${XDG_RUNTIME_DIR:=/run/user/$(id -u)}"
+export XDG_RUNTIME_DIR
--- a/deploy/nats/setup.sh
+++ b/deploy/nats/setup.sh
@@ -21,11 +21,8 @@
 # To rotate keys: sudo rm -f /etc/nats/nkeys/auth.conf && rm -rf ~/.lyra/nkeys && make nats-setup
 
 set -euo pipefail
-# Required when invoked via `make remote` (SSH non-interactive shell): without
-# it, `systemctl --user` fails to locate the dbus session ("Failed to connect
-# to bus: No such file or directory"). Provision.sh sets this consistently;
-# the deploy scripts must too.
-export XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-/run/user/$(id -u)}"
+# shellcheck source=../lib/env.sh
+source "$(dirname "$0")/../lib/env.sh"
 
 [[ $EUID -eq 0 ]] && { echo "[!] Do not run as root — use: make nats-setup"; exit 1; }
 
--- a/deploy/quadlet-install-verify.sh
+++ b/deploy/quadlet-install-verify.sh
@@ -10,11 +10,8 @@
 #
 # Exits non-zero if any unit fails to reach `active`.
 set -euo pipefail
-# Required when invoked via `make remote` (SSH non-interactive shell): without
-# it, `systemctl --user` fails to locate the dbus session ("Failed to connect
-# to bus: No such file or directory"). Provision.sh sets this consistently;
-# the deploy scripts must too.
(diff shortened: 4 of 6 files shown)


MickaelV0:
## Code Review

**PR #1128** — `refactor(deploy): extract XDG_RUNTIME_DIR guard to deploy/lib/env.sh`

Audited dimensions: behavior preservation, idempotency, path resolution, Makefile expansion, strict-mode safety, missed callers, PR #1130 conflict, architecture note.

---

### suggestion(non-blocking): `$(dirname "$0")` is fragile when script is sourced

**File:** `deploy/lib/env.sh` (header comment) / all 4 callers

When a script is **sourced** (`. script.sh` or `source script.sh`), `$0` remains the invoking shell's `$0`, not the sourced file's path. `$(dirname "$0")` therefore resolves to the wrong directory. The correct idiom for a sourced library is `$(dirname "${BASH_SOURCE[0]}")`.

In this PR the lib is sourced from *caller scripts* (not from an interactive shell), so `$0` in the caller at the time the `source` line executes still equals the caller script itself — which is correct. However the usage example in `env.sh`'s own header (`source "$(dirname "$0")/../lib/env.sh"`) would mislead anyone who tries to source the callers themselves from another script.

More critically: if `quadlet-install-verify.sh` is ever sourced rather than invoked directly (e.g. from a future wrapper), the path breaks silently. Hardening all four callers to `$(dirname "${BASH_SOURCE[0]}")` costs nothing and is the established bash idiom for library-relative paths.

```bash
# Before (all 4 callers)
source "$(dirname "$0")/lib/env.sh"
source "$(dirname "$0")/../lib/env.sh"

# After
source "$(dirname "${BASH_SOURCE[0]}")/lib/env.sh"
source "$(dirname "${BASH_SOURCE[0]}")/../lib/env.sh"
```

---

### issue(non-blocking): `: "${XDG_RUNTIME_DIR:=…}"` vs `export XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-…}"` — semantic difference

**File:** `deploy/lib/env.sh` line 45–46

The old per-script pattern was:
```bash
export XDG_RUNTIME_DIR="${XDG_RUNTIME_DIR:-/run/user/$(id -u)}"
```

The new lib does:
```bash
: "${XDG_RUNTIME_DIR:=/run/user/$(id -u)}"
export XDG_RUNTIME_DIR
```

The `:=` form **sets the shell variable and exports nothing**; the subsequent `export` line then exports the now-set variable. This is functionally equivalent for child processes. However there is one edge case: if `XDG_RUNTIME_DIR` was already set but **not exported** in the calling environment, the old form would re-export it; the new form's `export XDG_RUNTIME_DIR` would export whatever value the variable holds (which may be the pre-existing un-exported value, not the default). This is a theoretical difference only — in practice `XDG_RUNTIME_DIR` is either fully absent or already exported in SSH sessions. Flag as awareness, not a blocker.

---

### praise: idempotency is correct

**File:** `deploy/lib/env.sh`

The `:=` (assign-if-unset) operator correctly makes double-sourcing a no-op. `id -u` is only called once (on first source), and subsequent sources skip the assignment entirely. This is the right pattern.

---

### praise: Makefile `$$(id -u)` expansion is correct

**File:** `Makefile` lines 9, 17

Inside a Make recipe single-quoted SSH heredoc, `$$` correctly escapes to a single `$` at the shell level, so `$$(id -u)` expands to `/run/user/<UID>` on the **remote** host (not on the local make process). This is the intended behaviour.

---

### nitpick: CI failure is pre-existing, unrelated to this PR

**Context:** The `ci` check is failing with:
```
FAILED tests/test_bootstrap_credential_resolution.py::TestCredentialResolution::test_discord_token_passed_to_adapter_start
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.lyra/discord.db'
```
This is unrelated to shell/deploy changes. Coverage passes at 85.17% (threshold 50%). The failure should be tracked separately and does not block this refactor.

---

### observation(architecture): `deploy/lib/env.sh` is the natural home for `validate_unix_user()` from #1125

**Non-blocking.** PR #1125 adds a `LYRA_USER` validation pattern to `deploy/nats/setup.sh`. When that pattern needs to be shared across more scripts, `deploy/lib/env.sh` (or a sibling `deploy/lib/validate.sh`) is the natural SSoT. Worth noting in the lib's header comment at that time.

---

### observation: PR #1130 conflict check — clear

PR #1130 (`fix/1125-nats-setup-lyra-user-injection`) touches `deploy/nats/setup.sh` lines ~28–35 (validation + printf fix). This PR replaces lines 55–59 of the same file (XDG export → source). The sections are non-overlapping; no merge conflict is expected. Whichever merges second will need a trivial rebase at most.

---

### missed-caller audit — clear (with proviso)

`git grep XDG_RUNTIME_DIR deploy/ Makefile` on staging shows 4 deploy scripts still carrying the inline export (all updated in this PR) plus `deploy/provision.sh`. The `provision.sh` references are intentionally different: they pass XDG inline per `sudo -u $ADMIN_USER XDG_RUNTIME_DIR=… cmd` for a different user context — not a missed update.

---

**Verdict: LGTM with suggestion**

The refactor is correct and the idempotency claim holds. One suggestion to harden: replace `$(dirname "$0")` with `$(dirname "${BASH_SOURCE[0]}")` in all 4 caller `source` lines before merge. Not a blocker given current invocation patterns, but it is the defensive bash standard for sourced-library resolution.

PR #1130, opened 2026-05-07
fix(deploy): validate LYRA_USER + use printf in nats/setup.sh (#1125)

### LINKED ISSUE #1125: harden(deploy): nats/setup.sh — validate LYRA_USER from SUDO_USER + use printf for .env append

MickaelV0:
## Summary

Security fix for shell variable injection via `SUDO_USER` in `deploy/nats/setup.sh`.

- **Validate at entry:** `LYRA_USER` is now checked against `^[a-z_][a-z0-9_-]*$` immediately after assignment; script exits 1 on mismatch (fail-closed).
- **Eliminate injection sink:** all `.env` appends that carry user-derived paths switch from `echo` to `printf 'KEY=%s\n'`, closing the newline-injection vector even if validation were bypassed.

## Security note

`SUDO_USER` is attacker-controllable when the script is invoked via `sudo -E` or a sudoers `env_keep` rule. A payload like `$'foo\nNATS_NKEY_SEED_PATH=/attacker/path'` would have injected an extra line into `.env`. The validation check is placed before `LYRA_HOME`, `HUB_SEED`, and all other derivations — nothing downstream runs with an untrusted username.

## Test plan

- [ ] `bash -n deploy/nats/setup.sh` passes (syntax OK)
- [ ] `SUDO_USER=$'foo\nNATS_NKEY_SEED_PATH=/attacker/path' <validation snippet>` → exits 1, prints `[!] Invalid LYRA_USER`
- [ ] `SUDO_USER='; touch /tmp/pwned' <validation snippet>` → exits 1
- [ ] `SUDO_USER=mickael <validation snippet>` → passes through

Closes #1125

### DIFF
--- a/deploy/nats/setup.sh
+++ b/deploy/nats/setup.sh
@@ -184,8 +184,9 @@ section "Done"
 LYRA_USER="${SUDO_USER:-$(id -un)}"
 # Security: validate LYRA_USER before any use — SUDO_USER is attacker-controllable
 # (sudo -E / env_keep). Reject anything that doesn't look like a valid Unix username.
-[[ "$LYRA_USER" =~ ^[a-z_][a-z0-9_-]*$ ]] \
-  || { echo "[!] Invalid LYRA_USER: $LYRA_USER"; exit 1; }
+# Regex matches deploy/provision.sh USER_RE — 32-char POSIX cap.
+[[ "$LYRA_USER" =~ ^[a-z_][a-z0-9_-]{0,31}$ ]] \
+  || error "Invalid LYRA_USER: $LYRA_USER"
 LYRA_HOME=$(getent passwd "$LYRA_USER" | cut -d: -f6)
 ENV_FILE="${LYRA_DIR}/.env"
 HUB_SEED="${LYRA_HOME}/.lyra/nkeys/hub.seed"
--- a/deploy/nats/setup.sh
+++ b/deploy/nats/setup.sh
@@ -182,6 +182,10 @@ section "Done"
 
 # ── 10. Wire NATS env vars into .env ─────────────────────────────────────
 LYRA_USER="${SUDO_USER:-$(id -un)}"
+# Security: validate LYRA_USER before any use — SUDO_USER is attacker-controllable
+# (sudo -E / env_keep). Reject anything that doesn't look like a valid Unix username.
+[[ "$LYRA_USER" =~ ^[a-z_][a-z0-9_-]*$ ]] \
+  || { echo "[!] Invalid LYRA_USER: $LYRA_USER"; exit 1; }
 LYRA_HOME=$(getent passwd "$LYRA_USER" | cut -d: -f6)
 ENV_FILE="${LYRA_DIR}/.env"
 HUB_SEED="${LYRA_HOME}/.lyra/nkeys/hub.seed"
@@ -191,21 +195,21 @@ if [ -f "${ENV_FILE}" ]; then
   if grep -q "^NATS_URL=" "${ENV_FILE}"; then
     info ".env already has NATS_URL — not overwriting."
   else
-    echo "NATS_URL=tls://127.0.0.1:4222" >> "${ENV_FILE}"
+    printf 'NATS_URL=tls://127.0.0.1:4222\n' >> "${ENV_FILE}"
     info "NATS_URL=tls://127.0.0.1:4222 added to .env."
   fi
   # NATS_NKEY_SEED_PATH — nkey authentication
   if grep -q "^NATS_NKEY_SEED_PATH=" "${ENV_FILE}"; then
     info ".env already has NATS_NKEY_SEED_PATH — not overwriting."
   else
-    echo "NATS_NKEY_SEED_PATH=${HUB_SEED}" >> "${ENV_FILE}"
+    printf 'NATS_NKEY_SEED_PATH=%s\n' "${HUB_SEED}" >> "${ENV_FILE}"
     info "NATS_NKEY_SEED_PATH=${HUB_SEED} added to .env."
   fi
   # NATS_CA_CERT — CA certificate for TLS verification
   if grep -q "^NATS_CA_CERT=" "${ENV_FILE}"; then
     info ".env already has NATS_CA_CERT — not overwriting."
   else
-    echo "NATS_CA_CERT=${NATS_CA}" >> "${ENV_FILE}"
+    printf 'NATS_CA_CERT=%s\n' "${NATS_CA}" >> "${ENV_FILE}"
     info "NATS_CA_CERT=${NATS_CA} added to .env."
   fi
 else


MickaelV0:
## Code Review

**PR:** #1130 — fix(deploy): validate LYRA_USER + use printf in nats/setup.sh
**Scope:** 1 file, 40 lines (`deploy/nats/setup.sh`)
**Agents:** security-auditor, devops · single chunk · no recall triggered
**Verdict:** **Request changes** — 1 blocking, 1 high-confidence drift issue, 4 advisory

The core fix is correct: validation gate + `printf` migration meaningfully reduces the SUDO_USER injection surface. Findings below land on consistency, defense-in-depth, and parity with `deploy/provision.sh` which already has tighter validation for the same class of value.

---

### 🔴 Blocking

**issue(blocking): Validation error path uses bare `echo` — terminal-injection on the rejection branch + breaks the printf-everywhere policy this PR establishes**
  `deploy/nats/setup.sh:187`
  -- security-auditor + devops (merged)
  Root cause: When the regex rejects `$LYRA_USER`, the failing value reaches `echo "[!] Invalid LYRA_USER: $LYRA_USER"`. By definition, any value reaching this branch failed the `[a-z0-9_-]` character class — meaning it *can* contain ANSI escapes, control chars, or newlines. `echo` passes those bytes verbatim to the terminal/log. Independently, the bare `echo … ; exit 1` compound diverges from the `error()` helper used everywhere else in the script (defined at line 37) and from the `printf 'KEY=%s\n'` pattern this PR introduces three lines below. Two reviewers flagged the same site from security and consistency angles.
  Class: [shell-injection, missing-input-validation, missing-error-handling]
  Raw callsites: [{file: "deploy/nats/setup.sh", line: 37}, {file: "deploy/nats/setup.sh", line: 187}]
  Solutions:
    1. Replace with `error "Invalid LYRA_USER: $LYRA_USER"` — uses the existing helper, exits 1, goes to stderr, and the helper itself uses `printf` internally so the value is safely formatted. (recommended)
    2. `{ printf '[!] Invalid LYRA_USER: %s\n' "$LYRA_USER" >&2; exit 1; }` — minimal change, fixes both terminal-injection and stderr routing without touching helpers.
  Confidence: 97%

---

### 🟡 Issues (high-confidence, fix recommended)

**issue: Regex omits the 32-char POSIX cap and drifts from `provision.sh`**
  `deploy/nats/setup.sh:186`
  -- security-auditor + devops (merged)
  Root cause: `deploy/provision.sh:29` already defines `USER_RE='^[a-z_][a-z0-9_-]{0,31}$'` — same intent, with the POSIX `LOGIN_NAME_MAX`/`useradd` 32-char cap. The new regex is `^[a-z_][a-z0-9_-]*$` (unbounded). An overlong username passes the regex, fails downstream at `getent passwd` returning empty, and ends up writing a malformed `NATS_NKEY_SEED_PATH=/.lyra/nkeys/hub.seed` to `.env` with no clear rejection. The two scripts validating the same class of value should agree.
  Class: [missing-input-validation, parallel-path-drift]
  Raw callsites: [{file: "deploy/provision.sh", line: 29}, {file: "deploy/nats/setup.sh", line: 186}]
  Solutions:
    1. Change to `^[a-z_][a-z0-9_-]{0,31}$` to match `provision.sh`. Follow-up: extract `validate_unix_user()` into the upcoming `deploy/lib/` (PR #1128) so both scripts source the same regex. (recommended)
    2. Keep current regex and add `[[ ${#LYRA_USER} -le 32 ]]` as a separate check — functional but still drifts from `provision.sh`.
  Confidence: 93%

---

### 🟢 Suggestions

**suggestion: Validate `LYRA_HOME` non-empty after `getent` — silent failure on unknown user**
  `deploy/nats/setup.sh:185`
  -- devops + security-auditor (merged)
  Root cause: `getent passwd "$LYRA_USER"` returns exit 0 with empty stdout when the user doesn't exist. `LYRA_HOME` becomes `""`, `HUB_SEED` becomes `"/.lyra/nkeys/hub.seed"`, a bogus path is written to `.env`, and the script reports success. `provision.sh:34` already guards this exact case: `[[ -n "$ADMIN_HOME" && -d "$ADMIN_HOME" ]] || error "..."`. Same parity gap as the regex finding above. Also: `LYRA_HOME` (the home-dir field of the passwd record) is itself unvalidated and flows into `info`/`warn` calls that use `echo -e`, so a passwd entry with weird chars would also escape into terminal output — defense-in-depth concern.
  Class: [missing-input-validation, shell-injection]
  Raw callsites: [{file: "deploy/provision.sh", line: 34}, {file: "deploy/nats/setup.sh", line: 185}, {file: "deploy/nats/setup.sh", line: 207}, {file: "deploy/nats/setup.sh", line: 214}]
  Solutions:
    1. Add after line 185: `[[ -n "$LYRA_HOME" && -d "$LYRA_HOME" ]] || error "Could not resolve home for $LYRA_USER — does the account exist?"` — mirrors `provision.sh` exactly. (recommended)
    2. Switch the `info`/`warn` helpers from `echo -e` to `printf '%s\n'` to remove the escape-interpretation surface globally — bigger change, defers consolidation.
  Confidence: 90%

**suggestion: User-resolution mechanism diverges from `provision.sh` (SUDO_USER vs ADMIN_USER)**
  `deploy/nats/setup.sh:184`
  -- devops
  Root cause: `provision.sh:24` uses `ADMIN_USER="${ADMIN_USER:-$(whoami)}"` — explicit, documented env-override. `setup.sh` uses the implicit `SUDO_USER` ambient variable. Same purpose, different mechanism. Natural consolidation point when `deploy/lib/` lands.
  Class: [parallel-path-drift]
  Raw callsites: [{file: "deploy/provision.sh", line: 24}, {file: "deploy/nats/setup.sh", line: 184}]
  Solutions:
    1. Track as follow-up — extract `resolve_lyra_user()` into `deploy/lib/validate.sh` covering both resolution and validation. The current PR is net-positive as-is. (recommended)
    2. Add a `LYRA_USER="${LYRA_USER:-${SUDO_USER:-$(id -un)}}"` override knob now — one-liner, gives operators an explicit path.
  Confidence: 85%

**suggestion(non-blocking): Pin locale before regex match to remove `LC_CTYPE`/`LC_COLLATE` ambiguity**
  `deploy/nats/setup.sh:186`
  -- security-auditor
  Root cause: Bash `[[ =~ ]]` honors the inherited locale. With `sudo -E`, `LC_*` is preserved from the caller. For a security-critical regex on attacker-influenced input, locale-pinning is good practice. Low-impact (the `[a-z0-9_-]` class is robust across reasonable locales) but cheap to fix.
  Raw callsites: [{file: "deploy/nats/setup.sh", line: 186}]
  Solutions:
    1. `LC_ALL=C [[ "$LYRA_USER" =~ ^[a-z_][a-z0-9_-]{0,31}$ ]]` — pins ASCII semantics. (recommended)
    2. Wrap the whole validation block in a function with `local LC_ALL=C` set at the top.
  Confidence: 65%

---

### 💭 Nitpick

**nitpick: Comment "before any use" — `LYRA_USER` is interpolated at line 184 (assignment) and 185 (`getent`) before the validation runs**
  `deploy/nats/setup.sh:183-186`
  -- devops
  Root cause: Strictly, line 185 (`getent passwd "$LYRA_USER"`) executes before the validation on line 186. In practice this is safe (`getent` will return empty for malicious values), but the comment is mildly misleading. Reordering to validate before `getent` is also a minor robustness improvement.
  Solutions:
    1. Move the validation block to between lines 184 and 185 (validate immediately after assignment, before any expansion). (recommended)
    2. Update comment to "Security: validate LYRA_USER before downstream use".
  Confidence: 80%

---

### 🌟 Praise

**praise: `printf 'KEY=%s\n'` migration on the three `.env` writes**
  `deploy/nats/setup.sh:198, 205, 212`
  -- security-auditor
  Closes the newline-injection sink on `HUB_SEED` and `NATS_CA` independently of the validation gate. Defense-in-depth pattern done right — the validation could be removed/relaxed in the future and the `.env` writes stay safe.

---

### Summary

| Category | Count | Top concern |
|---|---|---|
| 🔴 Blocking | 1 | terminal-injection on rejection branch (line 187) |
| 🟡 Issues | 1 | 32-char regex cap + drift from `provision.sh` (line 186) |
| 🟢 Suggestions | 3 | LYRA_HOME guard parity, SUDO_USER drift, locale pinning |
| 💭 Nitpick | 1 | comment accuracy |
| 🌟 Praise | 1 | `printf 'KEY=%s\n'` migration |

**Recurring theme:** `deploy/provision.sh` already has tighter validation (32-char cap, home-dir guard, explicit env override) for the same class of value. The cleanest path is to apply the blocker fix here, plus the regex/length cap to match `provision.sh`, and track the consolidation into `deploy/lib/validate.sh` as a follow-up that lands alongside or after PR #1128.

**Verdict: Request changes** — primarily for the line 187 terminal-injection blocker. Once that's addressed, the regex cap is a small, mechanical follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/code-review`

MickaelV0:
## Re-review (after fix commit 451ec875)

Both blockers and the tracked issue are resolved:

- ✅ **Blocker (line 187):** bare `echo … ; exit 1` replaced with `error "Invalid LYRA_USER: $LYRA_USER"` — uses the established helper, consistent with provision.sh style.
- ✅ **Issue (line 186):** regex updated to `^[a-z_][a-z0-9_-]{0,31}$` — cap matches `provision.sh:29` USER_RE exactly.
- ✅ **Comment (line 187):** updated to reference provision.sh USER_RE parity.

Sanity tests on the regex:
- `mickael` → match ✅
- 32-char string → match ✅ (boundary)
- 33-char string → no match ✅ (cap enforced)
- `; touch /tmp/x` → no match ✅ (injection blocked)

Remaining suggestions from the original review are **intentionally deferred** as follow-up scope (not blocking merge):
- LYRA_HOME non-empty guard after `getent` lookup
- SUDO_USER vs ADMIN_USER naming drift with provision.sh
- Locale pinning for `getent` / `cut`
- Comment accuracy on the `|| error` line

Merging.

Issue #1118, opened 2026-05-07
harden(deploy): rotate-gh-key.sh PEM_RE allows path traversal via '/'


MickaelV0:
Discovered during code review of PR #1115 (XDG export fix). Pre-existing — not introduced by that PR.

## Issue

`deploy/scripts/rotate-gh-key.sh:24` defines:

```bash
PEM_RE='^[A-Za-z0-9._/-]+$'
```

The regex allows `/` without blocking `..` sequences. A caller passing `../../../../etc/some.pem` passes validation; if the file exists, it gets loaded as the GitHub App private key via:

```bash
podman secret create lyra-gh-pem "$NEW_PEM"
```

Substituting attacker-controlled credentials into the Podman secret store → impersonation of the Lyra App installation on GitHub API.

## Exploit precondition

Argument control:
- CI/CD write access (Makefile target args)
- `make remote` SSH foothold

Not zero-precondition, but realistic in lyra's threat model where argument control = anyone who can trigger the rotation workflow.

## Fix options

**Option 1 (recommended):** Resolve to canonical path + assert trusted prefix
```bash
RESOLVED=$(realpath -e "$NEW_PEM")
[[ "$RESOLVED" == /home/lyra/secrets/* || "$RESOLVED" == /etc/lyra/* ]] || die "path outside trusted dirs"
```

**Option 2:** Drop `/` from PEM_RE, require absolute path, `realpath --canonicalize-existing` matches expected dir.

## Sister scripts to audit at the same time

- `deploy/scripts/rotate-claude-oauth.sh` — does it have similar argument-pattern validation?

Source: security-auditor finding from PR #1115 review (confidence 72%, classed `path-traversal` + `missing-input-validation`).

PR #1132, opened 2026-05-07
fix(test): isolate LYRA_VAULT_DIR per test to prevent discord.db worker race

MickaelV0:
## Summary

Fixes a parallel-worker race on `~/.lyra/discord.db` that surfaces in CI as:

```
FileNotFoundError: /home/runner/.lyra/discord.db
test_discord_token_passed_to_adapter_start
```

This failure has been blocking auto-merge on every deploy/security PR (it's a required CI check) — surfaced during the #1118/#1120/#1125/#1126 cleanup pass.

## Root cause

`pytest -n auto` runs tests in parallel workers. All tests using the shared `_patch_nats_stubs` helper (called from `patch_bootstrap_common` / `patch_all` / `patch_auth_config_test`) write to the same default vault dir `~/.lyra`. `_ensure_discord_db` has a TOCTOU window:

1. Worker A creates `discord.db`
2. Worker B sees `exists() == True`, `_has_sentinel() == False`
3. Worker A deletes the file
4. Worker B calls `unlink()` → `FileNotFoundError`

## Fix

`tests/conftest.py` (+4 lines): set `LYRA_VAULT_DIR=tempfile.mkdtemp()` inside `_patch_nats_stubs`. Each worker now writes to its own isolated tmp dir.

## Test plan

- [x] `uv run pytest tests/ -n auto` — full suite passes
- [x] `uv run pytest tests/adapters/test_discord_*.py -n auto` — passes
- [x] All pre-push hooks (lint, typecheck, trufflehog, import-layers) pass
- [ ] CI green on this PR (this is the test that's currently failing — verifies the fix)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 from typing import Any, cast
 from unittest.mock import AsyncMock, MagicMock
@@ -180,6 +181,9 @@ def _patch_nats_stubs(monkeypatch: pytest.MonkeyPatch) -> None:
     )
     monkeypatch.setenv("NATS_URL", "nats://localhost:4222")
     monkeypatch.setenv("LYRA_HEALTH_PORT", "0")
+    # Isolate vault dir per test to prevent parallel-worker races on
+    # ~/.lyra/discord.db (_ensure_discord_db TOCTOU with -n auto).
+    monkeypatch.setenv("LYRA_VAULT_DIR", tempfile.mkdtemp())
 
 
 def make_fake_stores(


MickaelV0 (commented):
## Code Review

**PR:** fix(test): isolate LYRA_VAULT_DIR per test to prevent discord.db worker race
**Verdict:** Approve with comments

---

### Blockers

_None._

---

### Warnings

**suggestion(cleanup): `tempfile.mkdtemp()` leaks tmp dirs — use `tmp_path_factory` instead**
  tests/conftest.py:186
  -- tester
  Root cause: `mkdtemp()` creates a dir that is never cleaned up. Each parallel worker (and each test within a worker) accumulates a tmp dir for the lifetime of the process. The pattern is inconsistent with the rest of the test suite: every other `LYRA_VAULT_DIR` isolation in `tests/cli/`, `tests/bootstrap/`, `tests/nats/`, etc. uses `tmp_path` or `tmp_path_factory` (pytest-managed, auto-cleaned). `_patch_nats_stubs` is a plain function, not a fixture, so it cannot declare `tmp_path` as a parameter — but the callers (`patch_bootstrap_common`, `patch_all`, `patch_auth_config_test`) already receive `monkeypatch` and could accept an optional `tmp_path` or `tmp_path_factory` arg forwarded from their test fixtures.
  Solutions:
    1. (recommended) Accept an optional `tmp_vault: Path | None = None` parameter in all three callers, threading a `tmp_path`-provided path down to `_patch_nats_stubs`; fall back to `mkdtemp()` when not provided. This preserves backward compat while enabling auto-cleanup for callers that can supply a fixture path.
    2. Add a module-scoped `autouse` fixture that registers a `tempfile.mkdtemp()` result with `request.addfinalizer(shutil.rmtree)` — eliminates leaks without changing call sites.
  Confidence: 85%

---

### Suggestions

**nit: comment could mention the tmp dir is not auto-cleaned**
  tests/conftest.py:184–186
  -- tester
  The comment says "Isolate vault dir per test" but does not warn that the dir is not cleaned up. Future readers may not realise this differs from `tmp_path` semantics.
  Solutions:
    1. Append `# NOTE: mkdtemp is not auto-cleaned; see suggestion to use tmp_path_factory` to the comment.
    2. No change if the warning finding is acted on (becomes moot).
  Confidence: 90%

---

### Praise

- **Race analysis is correct.** `_ensure_discord_db(vault_dir)` is called inside `open_stores()`, which is invoked by `_bootstrap_unified()`. `vault_dir` is resolved from `LYRA_VAULT_DIR` at call time (`unified.py:47`). Setting the env var before any bootstrap call fully redirects the TOCTOU-prone path to a per-worker tmp dir — the fix works as described.
- **Scope is correct.** `_patch_nats_stubs` is the sole entry point for all three public helpers (`patch_bootstrap_common`, `patch_all`, `patch_auth_config_test`), so a single injection point covers all affected tests. Tests that do not call these helpers (adapters, CLI, core) either do not touch `_ensure_discord_db` at all or already use `tmp_path` directly.
- **Single-worker safety.** `monkeypatch.setenv` is test-scoped and auto-reverts; `mkdtemp()` returns a fresh path each time. Neither breaks single-worker (`pytest` without `-n auto`) runs.
- **Minimal diff.** 4 lines, no logic change to source, no new fixtures required. Appropriate scope for a CI-blocking flake fix.

Issue #1078, opened 2026-05-05
feat: GitHub App identity for Lyra clipool — v1


MickaelV0:
## Problem

Lyra has no GitHub identity. The `lyra-clipool` container running `claude` cannot commit, open PRs, or manage issues on behalf of Lyra. Today, any `git`/`gh` operation from inside Claude fails — no token, no gitconfig, no SSH.

## Approach

Decided via 3-expert consensus panel (architect + security-auditor + devops). Source of truth: `artifacts/analyses/lyra-github-app-identity-consensus.mdx` (commit e5e931a7).

**Headline**: GitHub App-only identity (no machine user), token isolated to a helper-uid process (never in claude subprocess env), tmpfs cache with 45-min refresh, 2 Apps day 1 (Lyra-prod on M₁, Lyra-dev on M₂), allowlisted Roxabi repos, minimum permissions only.

## Hard non-negotiables (from panel verdict)

- **Token NEVER in claude subprocess env** — prompt-injection threat: `Read /proc/self/environ` or `Bash curl …?t=$GH_TOKEN` exfiltrates org-wide write token. Token lives only in helper process owned by a distinct uid.
- Tmpfs cache `/run/lyra-gh-token` mode 0700, owned by helper uid (≠ lyra app uid 1500). Filesystem ACL keeps Claude out.
- `ReadOnly=true` container preserved — only `Tmpfs=` additions, no writable bind mounts for credentials.
- **Hub container excluded** from GitHub auth — preserves issue #941 boundary (hub does not mount `~/.claude` or PEM).
- `gh` CLI dropped from Dockerfile — replaced by Python token-mint helper + `git-credential-lyra-gh` + `lyra-gh` shim scripts.
- Branch protection on Roxabi repos: `lyra[bot]` PRs cannot auto-merge, human review mandatory, no org-owner bypass on `lyra[bot]` PRs, no force-push from App.

## Out of scope (v1)

- `BotIdentity` registry generalization (deferred to 2nd platform / 2nd bot)
- Sandbox `lyra-sandbox` org (Phase 2 trigger)
- OIDC federation (24-month north star)

## Slices (refined by `/spec`)

1. Token-mint helper (Python, JWT + tmpfs cache + Unix socket dispenser)
2. `git-credential-lyra-gh` + `lyra-gh` wrapper scripts
3. Dockerfile + Quadlet wiring — drop `gh`, add `Tmpfs=`, `Secret=`, `Environment=`
[...]

PR #1153, opened 2026-05-09
feat(clipool): symlink gh to lyra-gh shim for transparent token isolation

MickaelV0:
## Summary
- `/usr/local/bin/gh` → symlinked to `lyra-gh` shim so callers that hardcode `gh` (e.g. Claude Code subprocesses) transparently mint via the dispenser instead of failing with `command not found`.
- Updated shim header comment to reflect the new equivalence; added a Dockerfile comment explaining no-recursion via `LYRA_GH_BIN` precedence.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | Follow-up to #1078 (AC-2 isolation gap for `gh`-hardcoded callers) | — |
| Implementation | 1 commit on `worktree-agent-ae3751c21ba7ec10f` | Complete |
| Verification | Lint ✅ Typecheck ✅ (pre-commit hooks) | Passed |

## Test Plan
- [ ] Rebuild clipool image, redeploy Quadlet
- [ ] `podman exec lyra-clipool gh --version` → succeeds, no token leak in env
- [ ] `podman exec lyra-clipool sh -c 'env | grep -i token'` after a `gh` call → token absent from parent env
- [ ] Verify recursion-safe: `podman exec lyra-clipool readlink /usr/local/bin/gh` → `/opt/lyra-gh/lyra-gh`; shim still execs `\$LYRA_GH_BIN`
- [ ] Claude Code inside clipool: `gh pr list` works without explicit `lyra-gh` prefix

---


MickaelV0:
## Code Review — PR #1153

**Verdict: Request changes** — 2 blocking findings.

Scope: 2 files, 9 lines. Single chunk. Reviewers: `security-auditor` (token isolation + recursion), `devops` (Dockerfile + deploy). Skipped: architect (|Δ| ≤ 5), product-lead (no spec), tester (comment-only on shim), frontend/backend (out of scope).

### Blockers

**1. `issue(blocking):` `|| true` masks symlink failure — `gh` shim may silently be absent**

`Dockerfile:60`
— flagged independently by `security-auditor` and `devops` (max C=82%)

Root cause: the entire compound `{ chmod … && ln -s lyra-gh && ln -s gh || true; }` shares one `|| true`. If `/usr/local/bin/gh` already exists (e.g. future base-image upgrade ships it), the second `ln -s` exits non-zero, `|| true` swallows the error, the build succeeds, but the `gh` shim alias is absent. Combined with the `mv /usr/bin/gh` on line 69, `gh` silently vanishes from the container — callers get `command not found`, not the shim.

Class: `bash-error-suppression`
Solutions:
1. Use `ln -sf` for both symlinks (idempotent, no `|| true` needed for these ops). **(recommended)**
2. Add `RUN test -L /usr/local/bin/gh` assertion after line 60 to fail loud.

Confidence: 82%

---

**2. `issue(blocking):` `LYRA_GH_BIN` unset → `command -v gh` resolves to shim → infinite exec loop**

`src/lyra/tools/gh_token/lyra-gh:14-18`
— `security-auditor` (C=88%)

Root cause: with `/usr/local/bin/gh` now pointing at `lyra-gh`, the fallback path (`GH_BIN=$(command -v gh)`) resolves to the shim itself. `exec env GH_TOKEN=… "$GH_BIN" "$@"` re-enters the shim. Pre-PR, `command -v gh` returned empty (gh not on PATH) and the `[ -z "$GH_BIN" ]` check exited 127 cleanly. This PR breaks that safety net.

Production happy path is safe — `LYRA_GH_BIN` is set via `ENV` in the image. The risk surfaces in: (a) a debug shell where someone `unset LYRA_GH_BIN`, (b) `env -i` strip-env contexts, (c) future test/tooling that doesn't inherit the image env. With `exec`, no stack growth; just an infinite PID-recycling loop pegging CPU until external kill.

Class: `unbounded-loop`
Solutions:
1. Add a self-reference guard right after the `command -v gh` fallback — abort if resolved path is the shim itself. **(recommended)**
   ```sh
   [ "$GH_BIN" = "$(readlink -f "$0" 2>/dev/null || echo "$0")" ] && {
     echo "lyra-gh: LYRA_GH_BIN unset and \`gh\` resolves to this shim — set LYRA_GH_BIN." >&2
     exit 127
   }
   ```
2. Remove the `command -v gh` fallback entirely — `LYRA_GH_BIN` is always set in the production image; any other context should fail loudly.

Confidence: 88%

### Nitpick

**3. `nitpick:` New Dockerfile comment partially duplicates the existing AC#5 block**

`Dockerfile:68-70`
— `devops` (C=70%)

The existing comment (lines 63-67) already explains why `gh` is off PATH. The 3 added lines re-state the same recursion guarantee. Could be folded into the AC#5 block as a single appended sentence. Non-blocking — minor redundancy.

### Praise

**`praise:`** Recursion safety reasoning is explicit in both code and comments — `LYRA_GH_BIN` precedence on lines 15-16 of the shim genuinely guarantees no recursion in the production image. Comment isn't aspirational; it matches the code. ✓

### Cleared

- AC-2 token isolation: preserved (`exec env GH_TOKEN=…` unchanged).
- Build ordering: clean (line 60 symlink target is direct, doesn't depend on `command -v gh` at build time).
- `$0` introspection / dual-name behavior: shim has none — `lyra-gh` and `gh` are identical entrypoints.
- Quadlet env: doesn't override `LYRA_GH_BIN` (inherits image `ENV`).
- Direct `/opt/lyra-gh/gh` access: not exploitable — without a token in env, the binary can't authenticate.

---
Generated via `/code-review`

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 2 findings
**Applied via 1b1:** 1 finding
**Deferred:** 0
**Skipped:** 1 (praise)
**Failed:** 0

### Auto-Applied
- [applied] **issue(blocking):** `Dockerfile:60` — `ln -s` → `ln -sf` for both shim symlinks; `|| true` no longer masks "file exists" failures. *(bash-error-suppression, C=82%)*
- [applied] **issue(blocking):** `src/lyra/tools/gh_token/lyra-gh:14-22` — added self-reference guard around the `command -v gh` PATH fallback; exits 127 with a clear message when `LYRA_GH_BIN` is unset and the resolved path equals the shim itself. Smoke-tested locally: guard fires correctly. *(unbounded-loop, C=88% / verified C=97%)*

### Applied (1b1)
- [applied] **nitpick:** `Dockerfile:63-69` — folded the 3-line gh-shim alias addendum into a single appended sentence inside the existing AC#5 block. Less redundant.

### Skipped
- praise (no action needed): recursion safety reasoning is explicit and matches the code.

### Commits
- `7468a889` — `fix(clipool): harden gh→lyra-gh symlink against silent install + self-loop`
- `66f8740e` — `docs(clipool): fold gh-shim alias note into AC#5 comment block`

---
Generated via `/fix`

MickaelV0:
## Code Review — PR #1153 (iteration 2)

**Verdict: Request changes** — 1 new blocking finding (partial-fix gap from iteration 1).

Scope unchanged: 2 files, ~25 lines. Re-reviewed by `security-auditor` (F2 fix verification) and `devops` (F1 + F3 fix verification).

### Fixes confirmed clean
- **F1 (`bash-error-suppression`)** ✅ resolved — `ln -sf` makes the install idempotent; `|| true` no longer masks symlink failures.
- **F3 (nitpick — comment dup)** ✅ resolved — folded sentence reads well, no redundancy.

### New blocker

**F4. `issue(blocking):` Self-loop guard bypassed when `LYRA_GH_BIN` itself points at the shim**

`src/lyra/tools/gh_token/lyra-gh:15-26`
— `security-auditor` (C=88%)

Root cause: the F2 fix placed the `SELF`/`RESOLVED` comparison **inside** the `if [ -z "$GH_BIN" ]` branch. When `LYRA_GH_BIN` is set to any non-empty value — including a path that resolves to the shim itself (e.g. `LYRA_GH_BIN=/usr/local/bin/gh` from a misconfigured Quadlet override or a debugging session) — line 15 populates `$GH_BIN`, the `if [ -z … ]` block is skipped entirely, and the guard never runs. The subsequent `[ ! -x "$GH_BIN" ]` check on line 27 passes (the shim symlink is executable), then `exec env GH_TOKEN=… "$GH_BIN" "$@"` re-enters the shim → infinite exec loop. The original F2 hazard is reintroduced through a different door.

Class: `unbounded-loop` (re-tagged from agent's free-text labels per canonical taxonomy)
Solutions:
1. **Move the self-reference check out of the `if [ -z "$GH_BIN" ]` block** — run it unconditionally after `$GH_BIN` is resolved, regardless of source. Any resolution that lands on the shim is an error. **(recommended)**
2. Add a parallel guard inside the `LYRA_GH_BIN` branch with a distinct error message. Keeps the two branches symmetric.

Confidence: 88%

### Non-blocking

**F5. `suggestion(non-blocking):` `readlink -f` failure fallback produces non-canonical paths — guard silently passes**

`src/lyra/tools/gh_token/lyra-gh:20-21`
— `security-auditor` (C=65%)

Root cause: when `readlink -f` fails for both `$0` and `$GH_BIN`, the fallbacks (`echo "$0"`, `echo "$GH_BIN"`) substitute raw string values. If those strings differ but resolve to the same inode (e.g. `./lyra-gh` vs `/usr/local/bin/gh`), the string comparison returns false and the guard silently passes. Real-world impact low — shim is always invoked via absolute path in the container — but the fallback is logically "don't error" rather than "verified safe."

Solutions:
1. Treat `readlink -f` failure as a hard error: `SELF=$(readlink -f "$0") || { echo "lyra-gh: cannot resolve own path" >&2; exit 1; }`. Container always has GNU coreutils, so the failure path should never fire.
2. Keep fallback but log a warning when `readlink` fails.

Confidence: 65%

### Note (recurring bug class)

This is a textbook `fix-pass-creates-defects` instance from the project's recurring-bug-classes RCA: the original F2 fix didn't apply the falsification step ("can the bug still happen via another path?"). The narrow scoping to the fallback branch missed the explicit-LYRA_GH_BIN-points-at-shim path. Iteration 2's F4 catches it — but ideally `/fix` Phase 6.5 would have surfaced it during fix application.

---
Generated via `/code-review` (iteration 2/3)

PR #1188, opened 2026-05-13
refactor(scripts): port check-inbox-prefix.sh to Python (#1040)

### LINKED ISSUE #1040: refactor(scripts): port check-inbox-prefix.sh to Python
## Problem

\`scripts/check-inbox-prefix.sh\` uses an inline Python heredoc for AST parsing inside bash (lines 46–62). It works but is untestable as a unit and inconsistent with the other ACL check scripts (\`check_acl_matrix_retired.py\`, \`check_request_reply_flows.py\`) which are proper Python with argparse and tests under \`tests/scripts/\`.

## Expected behaviour

Port to a standalone Python script \`scripts/check_inbox_prefix.py\` with:
- argparse CLI (consistent with other check scripts)
- Tests under \`tests/scripts/test_check_inbox_prefix.py\`
- CI step updated to call the new entrypoint
- Old \`.sh\` deleted

## Effort: M

MickaelV0:
## Summary

- Replaces `scripts/check-inbox-prefix.sh` (bash + embedded Python heredoc) with `scripts/check_inbox_prefix.py` — a proper standalone Python script mirroring sibling conventions (`check_acl_matrix_retired.py`, `check_request_reply_flows.py`)
- Exit-code contract follows the audit-exit-codes rule: exit 0 = scan ran OK (violations reported on stdout), exit 1 = scanner failure (bad path / I/O error)
- CI step in `.github/workflows/ci.yml` updated from `bash scripts/check-inbox-prefix.sh` to `uv run python scripts/check_inbox_prefix.py`

## Test plan

- [x] `uv run pytest tests/scripts/test_check_inbox_prefix.py` — 18 passed
  - Happy path: clean file exits 0, prints `OK:`, prod `src/` + `packages/` clean
  - f-string violation detected, names file, single-quote variant
  - Literal (AST) violation detected, names file + line number
  - Comments and docstrings not flagged (AST pass)
  - `test_*.py`, `conftest.py`, `tests/` subdirs excluded
  - Missing `--src` exits 1 with error on stderr
- [x] `uv run ruff check scripts/check_inbox_prefix.py tests/scripts/test_check_inbox_prefix.py` — all checks passed
- [x] All pre-commit hooks passed (lint, typecheck, file-length, folder-size, import-layers, trufflehog, license)

Closes #1040


MickaelV0:
## Code Review

**Verdict: Request changes** — 1 blocking finding (CI gate regression), plus 3 non-blocking issues and 5 nitpicks. 3 raw findings invalidated.

Reviewed by: security-auditor, tester, devops (single-chunk dispatch; |Δ|=4 files).

---

### 🔴 Blockers

**issue(blocking): CI step silently demoted from gate to observability — no verdict-to-CI bridge**
  `.github/workflows/ci.yml`:87
  -- devops
  Root cause: New script exits 0 unconditionally when scan runs (violations → stdout only). The old `.sh` exited 1 on violations, causing the CI step to fail. The new step `run: uv run python scripts/check_inbox_prefix.py` has no `|| exit 1`, no exit-code bridge. Violations now appear in logs but the step passes — a future PR introducing `inbox_prefix=f"_INBOX.{name}"` will silently merge. The repo memory rule (`feedback_audit_exit_codes_in_gates`) correctly says exit code = "scanner ran OK / scanner broke", but the verdict still needs a downstream consumer for the step to actually gate.
  Class: [missing-error-handling] candidate/silent-gate-demotion
  Solutions:
    1. (recommended) Add a verdict bridge matching the sibling `check_codes_sync` pattern: `uv run python scripts/check_inbox_prefix.py | tee /tmp/out; grep -q '^FAIL' /tmp/out && exit 1; exit 0` — or expose `--exit-non-zero` flag on the script and use it in CI.
    2. If observability-only is the intent, add a comment on the CI step: `# observability-only: violations logged, not gating` and consider relocating the check to a non-gating workflow.
  Confidence: 90%

---

### 🟡 Non-blocking issues

**issue: OSError silently suppressed in `_check_fstring` — unreadable files skipped without signal**
  `scripts/check_inbox_prefix.py`:75
  -- devops
  Root cause: `except OSError: pass` returns an empty violations list. Caller has no way to distinguish "file clean" from "file unreadable" (permission denied, broken symlink). `_check_literal` bundles OSError with SyntaxError/UnicodeDecodeError which are legitimate scan-time conditions; the fstring check has no such justification — files came from `_iter_source_files` so they exist.
  Class: [bare-except]
  Raw callsites: [{file: "scripts/check_inbox_prefix.py", line: 75}, {file: "scripts/check_inbox_prefix.py", line: 88}]
  Solutions:
    1. (recommended) Remove the try/except in `_check_fstring` entirely — let OSError propagate. In `_check_literal`, keep `(SyntaxError, UnicodeDecodeError)` but drop OSError.
    2. Log to stderr and re-raise: `except OSError as e: print(f"ERROR: cannot read {path}: {e}", file=sys.stderr); raise`.
  Confidence: 85%

**issue: subprocess spawned without coverage env — script coverage attribution will be 0%**
  `tests/scripts/test_check_inbox_prefix.py`:13
  -- tester
  Root cause: `subprocess.run([sys.executable, str(CLI)], ...)` spawns a separate process. Under `pytest --cov` or `coverage run`, the subprocess does not contribute coverage unless `COVERAGE_PROCESS_START` env is set and `[coverage:run] concurrency=subprocess` is configured. The entire 302-line test suite exercises the script only via subprocess — reported coverage for `check_inbox_prefix.py` will be 0%.
  Solutions:
    1. (recommended) Extract `_check_fstring`, `_check_literal`, `_iter_source_files` into unit tests that import them directly. Keep one or two subprocess tests for the CLI surface (argparse, exit codes).
    2. Configure subprocess coverage via `coverage.process_startup()` in a `sitecustomize.py` or add `COVERAGE_PROCESS_START` to the test runner.
  Confidence: 90%

**issue: `test_prod_source_exits_0` — tautological repo-state-dependent test**
  `tests/scripts/test_check_inbox_prefix.py`:27
  -- tester
  Root cause: Invokes the scanner against the live `src/` and `packages/` tree. Passes as long as prod source is clean today, but cannot fail when the scanner's detection logic is broken. Mental delete of `_check_fstring` and `_check_literal`: zero violations found, "OK:", exit 0 — test still passes. Also brittle if a contributor temporarily adds a raw `_INBOX.` during development.
  Class: [test-tautology]
  Solutions:
    1. (recommended) Replace with a fixture-based pair: write one clean file → assert exit 0 + "OK"; write one known-bad file → assert exit 0 + "FAIL" in stdout. The prod-source check, if kept, should be commented as a regression sentinel, not a logic test.
    2. Drop this test — the other 23 tests already cover the scanner logic via tmp_path fixtures.
  Confidence: 93%

---

### 🟢 Nitpicks

**nitpick: `uv run python` used for a stdlib-only script — unnecessary venv activation cost**
  `.github/workflows/ci.yml`:87
  -- devops
  Root cause: Script imports only `ast`, `re`, `sys`, `pathlib`, `argparse` — all stdlib. `uv run python` resolves and activates the project venv. Other stdlib-only checks in the same workflow use `bash scripts/...`; this creates an inconsistent pattern. Cold-start cost in CI is small but non-zero.
  Solutions: change to `python3 scripts/check_inbox_prefix.py` (matches shebang, consistent with siblings).
  Confidence: 80%

**nitpick: `_REPO_ROOT = Path(__file__).resolve().parents[1]` fragile under symlink/copy**
  `scripts/check_inbox_prefix.py`:37
  -- devops
  Root cause: `.parents[1]` hard-assumes script lives at `<root>/scripts/`. Old `.sh` used `cd "$(git rev-parse --show-toplevel)"` which is CWD-independent and symlink-safe. Low blast radius (CI always checks out at repo root, no pre-commit reference) but worth noting if the script is ever invoked as a pre-commit `entry:`.
  Class: candidate/fragile-path-anchor
  Solutions:
    1. (recommended) `subprocess.check_output(["git", "rev-parse", "--show-toplevel"], text=True).strip()` wrapped in try/except.
    2. Accept current approach with a module docstring note constraining the install location.
  Confidence: 75%

**nitpick: line-number assertion in `test_literal_violation_names_file_and_line` over-permissive**
  `tests/scripts/test_check_inbox_prefix.py`:71
  -- tester
  Root cause: `":2:" in combined or ":2 " in combined or "lineno=2" in combined or "2:" in combined` — the last alternative `"2:" in combined` matches any digit-colon pair, e.g. timestamps `12:34` or paths like `/tmp/pytest-2:.../`. Test passes vacuously when line number is missing or wrong.
  Class: [test-tautology]
  Solutions:
    1. (recommended) `assert re.search(r':2:', combined)` matching the exact `<path>:<line>:` format the script emits, OR `assert f"{src}:2:" in combined`.
  Confidence: 88%

**nitpick: regex `_FSTRING_PATTERN` misses concat / raw-string variants — no tests cover these bypass patterns**
  `scripts/check_inbox_prefix.py`:43
  -- tester
  Root cause: Pattern only matches `f"_INBOX."` literal start. Misses `inbox_prefix=f"_INBOX." + name` (concat, BinOp not Constant), `inbox_prefix=r"_INBOX.foo"` (raw string — AST should catch via `_check_literal`, but no test exists), `inbox_prefix=("_INBOX." f"{name}")` (parenthesized mixed). Same gap as the bash original, but worth covering since the port advertises stricter detection.
  Solutions:
    1. (recommended) Add a test for `inbox_prefix=r"_INBOX.foo"` (AST should catch — verifies `_check_literal` regression resistance). Document concat as out-of-scope or add detection.
  Confidence: 82%

**nitpick: missing negative guard for fstring detection path**
  `tests/scripts/test_check_inbox_prefix.py`:34
  -- tester
  Root cause: All fstring tests are positive (`FAIL in stdout`). No test exercises `_check_fstring` in isolation — both detection paths run on every input, so deleting `_check_fstring` would still cause some tests to fail via `_check_literal`'s output (or vice versa). Hard to distinguish which path is broken from the test names alone.
  Solutions:
    1. Add a test using a file with an f-string in a comment (regex matches text, AST doesn't) — distinguishes the two paths.
  Confidence: 85%

**nitpick: `"tests" in parts` exclusion — exact-component semantics not tested**
  `tests/scripts/test_check_inbox_prefix.py`:89
  -- tester
  Root cause: Exclusion logic in `_iter_source_files` checks `"tests" in p.parts` — exact path-component match. No test confirms that `tests_integration/` or `mytests/` is NOT excluded, nor that `tests_helper.py` in a non-test dir IS scanned. Semantics implicit.
  Solutions:
    1. Add a test with `tests_integration/foo.py` containing a violation → assert scanned. Document the exact-component semantics in `_iter_source_files`.
  Confidence: 80%

---

### Praise

`praise:` Exit-code contract docstring is crisp and matches the repo's audit-exit-codes rule (`feedback_audit_exit_codes_in_gates`) — "Exit 0: scan ran OK; Exit 1: scanner failed" is the right model for a static checker. The CI gate concern is about the *consumer* of the report, not the script's contract.

`praise:` Proper argparse + repeatable `--src` flag + sensible defaults (`src/` and `packages/`) match sibling check-script conventions cleanly. The AST-based literal check correctly avoids comment/docstring false positives.

`praise:` Test suite is well-organized into 5 classes by scenario (HappyPath, FstringViolation, LiteralViolation, Exclusion, MissingSrcTarget) with descriptive method names.

---

### Summary

| Category | Count |
|---|---|
| Blockers | 1 |
| Non-blocking issues | 3 |
| Nitpicks | 5 |
| Praise | 3 |
| Invalidated | 3 |

**Next step:** address the CI gate bridge before merge. Suggested fix is a 1-line shell bridge in `.github/workflows/ci.yml`.

MickaelV0:
## Review Fixes Applied

**Applied:** 3 finding(s) (F1, F4, F7) — commit `c6db1166`
**Skipped (verifier rejected):** 2 (F2, F5)
**Skipped (low-value nitpick):** 5 (F6, F8, F9, F10) — see notes below
**Enforcement diagnostics:** 1 (taxonomy path drift — non-blocking)

### Applied

- [applied] **F1 (blocker → revised)**: script now `sys.exit(1)` when violations found, matching sibling scanners `check_acl_matrix_retired.py` and `check_request_reply_flows.py`. Restores the CI gate the bash original provided. Original fix proposed a shell-bridge wrapper; deeper analysis of the workflow showed every other scanner here exits non-zero on violation — the script should match its siblings, no bridge needed. The `feedback_audit_exit_codes_in_gates` memory rule applies to stacked `set -e` quality-debt gates, not standalone CI workflow steps.
- [applied] **F4**: `test_prod_source_exits_0` replaced with `test_fixture_pair_clean_then_bad` — exercises clean fixture (exits 0, "OK") AND known-bad fixture (exits 1, "FAIL"). Falsification-resistant: silently removing either detection function fails the bad-fixture half.
- [applied] **F7**: line-number assertion tightened from `":2:" in combined or ":2 " in combined or "lineno=2" in combined or "2:" in combined` (last alt false-passes on `/tmp/pytest-2:.../`) to exact `f"{src}:2:"` substring match.

Cascading: 2 violation-exit tests renamed `_exits_0 → _exits_1` and assertions flipped to `returncode == 1` to match the new script contract.

**Test run:** 18/18 pass. Shell-level verification confirmed: clean → exit 0, known-bad → exit 1.

### Skipped (verifier rejected at C_v<T)

- **F2 (OSError suppress)** — C_v=30. Architect verifier: silent skip is intentional and matches `_check_literal`'s exception handling; consistent with the bash original. Not a real defect.
- **F5 (uv run for stdlib)** — C_v=45. Architect verifier: `uv run python` matches the sibling `check_codes_sync.py` step pattern; consistency outweighs minor cold-start cost.

### Skipped (low-value, defer to operational follow-up if needed)

- **F3 (subprocess coverage attribution)** — C_v=72. Real concern but the proper fix is a test refactor (extract `_check_fstring`/`_check_literal`/`_iter_source_files` into importable unit tests). Out-of-scope for this PR; file a follow-up if you want to chase the coverage metric.
- **F6 (_REPO_ROOT path anchor)** — C_v=60. Verifier modified: `git rev-parse` introduces subprocess dep + can fail in non-git invocations; current approach is acceptable given the script is only invoked from CI checkout root.
- **F8 (regex misses concat/raw-string)** — C_v=55. Verifier clarified: raw-string `r"_INBOX.foo"` IS caught by AST `_check_literal` (it's still an `ast.Constant`). Only the concat form (`f"_INBOX." + name`) is genuinely uncovered; same gap as the bash original.
- **F9 (missing fstring negative guard)** — C_v=70. Cosmetic — both detection paths run on every input; cross-coverage is partial but not zero.
- **F10 ("tests" path-component semantics)** — C_v=80, but coverage is sufficient for the current use case. The boundary tests would document semantics without protecting against a real regression.

### Enforcement diagnostics

- `[taxonomy-path-drift]` `skills/fix/review-classes.yml`:`0` — Skill spec at `skills/fix/SKILL.md:Phase 0` references `skills/fix/review-classes.yml` but the canonical file lives at `skills/code-review/review-classes.yml`; used the actual location. Worth raising upstream.

### Next

PR ready for re-review. The CI gate is restored without violating the memory rule (because the rule doesn't apply at this layer). Suggest squash-merge after CI green.

PR #1190, opened 2026-05-13
fix(nats): codec assert_never + Text v2 triplet branches; PR-template DoD

MickaelV0:
## Summary

Closes the Slice 2 (#1099) codec gap that crashed prod streaming today: `NatsRenderEventCodec.encode` had no branch for `TextStart/Delta/End/Chunk` events, so when `StreamProcessor` emitted them the encoder raised `TypeError: Unsupported RenderEvent subtype: TextStartRenderEvent`, the iterator was drained, and adapters showed "no display text" to users.

Sister fix that makes this class of bug **catchable by pyright at review time**, not at runtime in prod.

## Why this slipped through CI

- Codec has 14 `isinstance` branches + a `raise TypeError` final fall-through with `# pyright: ignore[reportUnreachable]` — pyright was muted, so the encoder was not enforced exhaustive
- All Slice 2 tests were in-process (`test_render_events.py`, `test_stream_processor.py`, `test_shared.py`) — none cross the `NatsChannelProxy` wire boundary
- The codec test file referenced only `TextRenderEvent` (v1) — zero coverage of v2 Text events
- Sibling slices 1/3/4 all updated the codec ✅; Slice 2 (#1173) did not

## Changes

- `src/lyra/nats/render_event_codec.py`
  - `encode()`: 4 new branches (`text_start`, `text_delta`, `text_end`, `text_chunk`)
  - `encode()`: replace `raise TypeError` + `pyright: ignore[reportUnreachable]` with `assert_never(event)`
  - `decode()`: 4 new `event_type` branches with schema-version check
  - Wire-format docstring lists the full event_type set
- `tests/nats/test_render_event_codec.py`: 9 new tests (encode shape + round-trip + schema floor)
- `tools/file_exemptions.txt`: bump codec cap 332→403 with #1099 rationale
- `.github/PULL_REQUEST_TEMPLATE.md`: new "RenderEvent changes" DoD section so future render-event PRs cannot omit the codec + test + docstring checklist

## Mechanical guarantee going forward

```python
# new RenderEvent subclass added to the union without a codec branch:
$ pyright src/lyra/nats/render_event_codec.py
error: Argument of type "NewRenderEvent" cannot be assigned to parameter "arg" of type "Never"
```

The encoder is now sum-type exhaustive: adding to the `RenderEvent` union without handling it in the codec fails pyright. Same mechanism `_shared_streaming_emitter.py` already uses on the consumer side.

## Type of change

- [x] Bug fix
[...]

PR #1173, opened 2026-05-13
feat(render-events): #1099 add v2 Text triplet (Slice 2 of #1096)

MickaelV0:
## Summary

- Adds v2 `TextStart/Delta/End/Chunk` RenderEvents alongside v1 `TextRenderEvent` (dual emission); `StreamingSession` routes the 4 new types to a per-platform `_on_text_v2` override seam (no-op default — parity preserved by unchanged v1 path).
- Per-block `message_id` (not per-turn) supports interleaved `text → tool → text` patterns; `TextEnd` closes the block at all 4 boundaries (ToolUse start, Result, truncation, exception).
- `_drain_fallback` harvests v2 deltas (v2-preferred elif to avoid double-count under dual emission).
- `TextChunkRenderEvent` is defined but not emitted — experimental, shape revisable until first consumer wires up.

Slice 2 of epic #1096 (RenderEvent v2 — back-port AG-UI modeling). Slices 1 (#1098 Run lifecycle) + 3 (#1100 ToolCall split) already merged; this slice instantiates the same template for the Text family. Schema bump cadence: additive only — no v1 floor raise (deferred to Slice 5 #1102).

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1099: refactor(render-events): split TextRenderEvent into Start/Delta/End + Chunk | OPEN |
| Frame | [1099-text-render-event-triplet-frame.mdx](artifacts/frames/1099-text-render-event-triplet-frame.mdx) | approved |
| Spec | [1099-text-render-event-triplet-spec.mdx](artifacts/specs/1099-text-render-event-triplet-spec.mdx) | approved |
| Plan | [1099-text-render-event-triplet-plan.mdx](artifacts/plans/1099-text-render-event-triplet-plan.mdx) | 11 micro-tasks, 5 waves |
| Implementation | 11 commits on `feat/1099-text-render-event-triplet` | Complete |
| Verification | ruff ✅ · pyright ✅ · pytest 3798 passed ✅ · import-linter (7/7) ✅ | Passed |

Tier: **F-lite**. Analyze skipped per process.

## Test Plan

- [ ] Unit: `tests/core/messaging/test_render_events.py` — 36 cases × 4 dataclasses (instantiation, frozen, schema_version default)
[...]

Issue #1096, opened 2026-05-06
epic: RenderEvent v2 — adopt AG-UI modeling (back-port, no wire integration)


MickaelV0:
## Context

AG-UI (https://github.com/ag-ui-protocol/ag-ui) ships a richer event model than Lyra's current `RenderEvent` hierarchy. This epic back-ports the 4 modeling improvements that are strictly better for Lyra's use case **without** adopting AG-UI as a wire format. Internal model becomes as rich as AG-UI; if/when AG-UI integration ships later, edge translation is mechanical.

Foundation being extended: ADR-032 (`docs/architecture/adr/032-llmevent-streamprocessor-renderevent-hexagonal-streaming.mdx`).

## Back-ports (in scope)

1. **Run lifecycle** — explicit `RunStarted/Finished/Error` events (vs implicit turn boundaries)
2. **Text triplet** — `TextStart/Delta/End` + `Chunk` convenience (vs single delta + done bit)
3. **ToolCall split** — `ToolCallStart/Args/End/Result` streamed (vs post-hoc `ToolSummaryRenderEvent`)
4. **Reasoning typed** — `ReasoningStart/Delta/End` stream (vs ⏳-prefix text hack via `show_intermediate`)

## Deferred (out of scope)

- StateSnapshot/Delta — no consumer until roxabi-dashboard exists
- ActivitySnapshot/Delta — same
- AG-UI as HTTP/SSE wire format / 5th channel adapter — separate decision

## Kept (rejected from AG-UI)

- `InboundMessage` richness — AG-UI's flat `Message` would lose multi-channel metadata
- History authority server-side (TurnStore + Pool) — AG-UI's "client transmits history" wrong for multi-channel + persistent memory
- `schema_version` per envelope — AG-UI has no version field
- Static plugin/skill model (ADR-010) — AG-UI's frontend-declared `tools[]` is orthogonal

## Slicing rule

Each slice = vertical thin slice: type def + StreamProcessor change + driver change (if any) + adapter consumption + tests. No "types only" slices.

## Dependency graph

```
ADR ──► S1 (Run lifecycle) ──► S2 (Text triplet) ──┐
                            └─► S3 (ToolCall split) ┼──► S5 (Cleanup + schema floor)
                            └─► S4 (Reasoning typed) ┘
```

MickaelV0:
All 6 children (#1097-#1102) closed. Epic complete.

PR #1193, opened 2026-05-14
test(nats): hub→NATS→adapter render-event wire round-trip (#1191)

### LINKED ISSUE #1191: test(nats): hub→NATS→adapter render-event integration test (Slice 5b of #1096)
## Goal

Slice 5b of epic #1096. Wire-boundary integration test for outbound render-event streaming. Truly orthogonal from sibling #1102a (codec/cutover) — can land in any order.

## Scope

Spin up an in-process NATS + hub channel proxy + adapter outbound listener. For every `RenderEvent` subclass in the union, send → assert decoded value equals original (round-trip wire fidelity).

Catches the exact wire-boundary class of bug that the in-process Slice 2 tests missed (PR #1173 → prod incident "no display text" → hotfix PR #1190).

## Acceptance

- Every member of `typing.get_args(RenderEvent)` has a round-trip test
- Test uses real NATS messaging (in-process server), not mocked
- Fails closed if a new `RenderEvent` type is added without a round-trip case (use the same `typing.get_args` enumeration trick that the registry-completeness test will use)

## Relationship to other tests

- Sibling of #632 (not replacement): #632 covers `lyra_cli` driver path (`NatsCliClient → CliAdapterStandalone → CliPool`); this covers the **outbound render-event streaming** path.
- Complements #1102a's registry-completeness test: completeness test guards Python-side dispatch; this guards wire-format round-trip.

## References

- Epic: #1096
- Umbrella: #1102
- Sibling: #1102a (codec/cutover)
- Wire-boundary incident: PR #1173 → hotfix PR #1190

MickaelV0:
## Summary
- Wire-boundary integration test (Slice 5b of #1096) that round-trips every member of `typing.get_args(RenderEvent)` through a real `nats-server` subprocess via `NatsChannelProxy.send_streaming()` → `NatsRenderEventCodec.decode()`.
- Companion completeness guard fails closed when `_SAMPLE_BY_TYPE` drifts from the `RenderEvent` union — same `typing.get_args` enumeration trick the upcoming registry-completeness test (#1102a) uses on Python-side dispatch.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1191: test(nats): hub→NATS→adapter render-event integration test | OPEN |
| Implementation | 1 commit on `feat/1191-hub-nats-adapter-render-event-test` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (17 new: 16 round-trips + 1 completeness) | Passed |

## Why this test

The in-process Slice 2 unit tests added by PR #1173 only exercised `NatsRenderEventCodec.encode/decode` directly on hand-crafted dicts — they never crossed a real NATS subject. When #1173 added `TextStart/Delta/End/Chunk` to the `RenderEvent` union without adding the corresponding `encode()` branches, the gap went undetected until prod, where `StreamProcessor` started emitting the new events and the encoder raised `TypeError` → adapters received zero events ("no display text" incident → hotfix PR #1190).

This test closes that wire-boundary gap and the fail-closed completeness check prevents the regression class from recurring: any new `RenderEvent` subclass must be added to `_SAMPLE_BY_TYPE` or the suite goes red.

## Test Plan
- [x] Locally validated against a real `nats-server v2.10.22` subprocess — 17/17 pass in ~3.5 s
- [x] Tested fail-closed: removing a sample from `_SAMPLE_BY_TYPE` correctly trips the completeness guard
- [ ] CI runs against `nats-server v2.10.22` (already installed by `.github/workflows/ci.yml`)
- [ ] If a future PR adds a new `RenderEvent` subclass without a sample, the completeness guard fails before merge

Closes #1191

---

### DIFF
--- a/tests/nats/test_render_event_wire_roundtrip.py
+++ b/tests/nats/test_render_event_wire_roundtrip.py
@@ -1,0 +1,238 @@
+"""Wire-boundary integration test for outbound render-event streaming.
+
+Slice 5b of #1096. Spins up a real ``nats-server`` (via the session-scoped
+fixture in ``conftest.py``), publishes one instance of every member of
+``typing.get_args(RenderEvent)`` through :class:`NatsChannelProxy.send_streaming`,
+subscribes from a real NATS client, decodes via :class:`NatsRenderEventCodec`,
+and asserts ``decoded == original`` for every member.
+
+Catches the exact wire-boundary class of bug that the in-process Slice 2
+codec unit tests missed (PR #1173 added ``TextStart/Delta/End/Chunk`` to the
+``RenderEvent`` union but not to the codec's ``encode()`` branches → prod
+incident "no display text" → hotfix PR #1190). The completeness check below
+fails closed when a new ``RenderEvent`` subclass is added without a sample —
+the ``typing.get_args(RenderEvent)`` enumeration is the same trick the
+upcoming registry-completeness test (#1102a) uses on the Python-side
+dispatch.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import typing
+from datetime import datetime, timezone
+
+import pytest
+from nats.aio.client import Client as NATS
+
+from lyra.core.auth.trust import TrustLevel
+from lyra.core.messaging.message import InboundMessage, Platform
+from lyra.core.messaging.render_events import (
+    FileEditSummary,
+    ReasoningDeltaRenderEvent,
+    ReasoningEndRenderEvent,
+    ReasoningStartRenderEvent,
+    RenderEvent,
+    RunErrorRenderEvent,
+    RunFinishedRenderEvent,
+    RunStartedRenderEvent,
+    SilentCounts,
+    TextChunkRenderEvent,
+    TextDeltaRenderEvent,
+    TextEndRenderEvent,
+    TextRenderEvent,
+    TextStartRenderEvent,
+    ToolCallArgsRenderEvent,
+    ToolCallEndRenderEvent,
+    ToolCallResultRenderEvent,
+    ToolCallStartRenderEvent,
+    ToolSummaryRenderEvent,
+)
+from lyra.nats.nats_channel_proxy import NatsChannelProxy
+from lyra.nats.render_event_codec import NatsRenderEventCodec
+from tests.nats.conftest import requires_nats_server
+
+pytestmark = [requires_nats_server]
+# asyncio mode is set to ``auto`` in pyproject.toml, so async tests get the
+# asyncio marker automatically — applying it globally would falsely tag the
+# sync completeness test below.
+
+
+# ---------------------------------------------------------------------------
+# Sample table — one realistic instance per RenderEvent subclass.
+# ---------------------------------------------------------------------------
+# Adding a new member to ``RenderEvent`` requires adding a sample here. The
+# ``test_sample_table_matches_render_event_union`` guard fails closed when
+# this table drifts from the union.
+
+_SAMPLE_BY_TYPE: dict[type, RenderEvent] = {
+    TextRenderEvent: TextRenderEvent(text="hello world", is_final=True, is_error=False),
+    TextStartRenderEvent: TextStartRenderEvent(message_id="msg-1"),
+    TextDeltaRenderEvent: TextDeltaRenderEvent(message_id="msg-1", delta="he"),
+    TextEndRenderEvent: TextEndRenderEvent(message_id="msg-1"),
+    TextChunkRenderEvent: TextChunkRenderEvent(message_id="msg-1", delta="hi"),
+    ToolSummaryRenderEvent: ToolSummaryRenderEvent(
+        files={
+            "/tmp/x.py": FileEditSummary(path="/tmp/x.py", edits=["+1"], count=1),
+        },
+        bash_commands=["ls"],
+        web_fetches=["https://example.com"],
+        agent_calls=["sub-agent"],
+        silent_counts=SilentCounts(reads=2, greps=1, globs=0),
+        is_complete=True,
+    ),
+    RunStartedRenderEvent: RunStartedRenderEvent(run_id="run-1"),
+    RunFinishedRenderEvent: RunFinishedRenderEvent(run_id="run-1", outcome="success"),
+    RunErrorRenderEvent: RunErrorRenderEvent(run_id="run-1", message="boom", code=None),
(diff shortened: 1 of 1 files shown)


MickaelV0:
## Code Review

Single-file diff (238 lines, `tests/nats/test_render_event_wire_roundtrip.py`). Fresh-context review by **tester** + **security-auditor** in parallel. Verdict: **Approve with comments** — no blockers for the #1191 scope; one diagnostic improvement worth picking up, two follow-ups noted.

**suggestion(non-blocking):** `_handler` silently catches malformed wire frames → asyncio timeout hides cause
  `tests/nats/test_render_event_wire_roundtrip.py:196`
  -- tester (78%) ∪ security-auditor (68%)
  Root cause: `json.loads(msg.data.decode(...))` runs bare inside the NATS subscribe callback. If the codec ever publishes a non-JSON frame, the exception propagates into the nats.py dispatch loop — depending on version, that either drops subsequent messages silently (the test then hangs to the 5 s timeout with no diagnostic) or surfaces as an unrelated event-loop error. The test is *designed* to catch codec wire-format regressions, so it should fail loudly with the offending bytes.
  Solutions:
    1. Wrap `json.loads` in `try/except json.JSONDecodeError` inside `_handler` and call `pytest.fail(f"malformed NATS frame: {msg.data!r}")` (recommended)
    2. Add a post-loop sanity check `assert all(isinstance(c, dict) for c in received)` before the length assertion
  Confidence: 73%

**suggestion(non-blocking):** `pytest.fail` at line 183 is unreachable — every parametrized `event_type` is already in `_UNION_MEMBERS`
  `tests/nats/test_render_event_wire_roundtrip.py:183`
  -- tester
  Root cause: `event_type` is drawn from `_UNION_MEMBERS`, so `event_type not in _SAMPLE_BY_TYPE` can only fire when a union member has no sample — the case `test_sample_table_matches_render_event_union` already blocks at collection time. The inline guard is harmless defense-in-depth but the comment at lines 64–67 implies it is the active mechanism, which is misleading.
  Solutions:
    1. Drop the inline guard and let `test_sample_table_matches_render_event_union` be the single source of fail-closed behaviour
    2. Keep the guard but mark it `# defensive — test_sample_table_matches_render_event_union fires first at collection`
  Confidence: 70%

**suggestion(non-blocking):** `nc` fixture is function-scoped — 16 NATS connections per test run (one per parametrized case)
  `tests/nats/test_render_event_wire_roundtrip.py:171`
  -- tester
  Root cause: `nats_server_url` is session-scoped but `nc` is function-scoped, so every parametrized round-trip opens and drains a fresh client against the same server. All round-trips are independent and stateless, so per-test isolation buys nothing.
  Solutions:
    1. Override `nc` locally with a module-scoped fixture in this test file
    2. Leave as-is — total runtime is ~3.5 s, isolation cost is negligible
  Confidence: 65%

### Follow-ups (out of scope for #1191)

**thought:** No coverage for the `outbound != None` branch of `send_streaming` (the `stream_start` header path) — `tests/nats/test_render_event_wire_roundtrip.py:212`. Out of scope for this PR (issue #1191 is specifically "render-event round-trip"); worth a separate issue alongside the existing mocked unit tests in `test_nats_channel_proxy.py`. Confidence: 80%.

**thought:** Security-auditor flagged a forward-looking concern in production code (not in this diff): `NatsChannelProxy` validates `bot_id` against `[A-Za-z0-9_-]+` but does not sanitise `Platform.value` before subject interpolation. Currently inert (the two enum values are clean tokens) but a future third platform with `.`/`*`/`>` in its value would silently produce malformed subjects. Out of scope — file a follow-up against `src/lyra/nats/nats_channel_proxy.py:99,120,207,233,281`. Confidence: 65%.

### Praise

**praise:** Completeness guard structured as a bidirectional set-difference (`missing` + `extra`) catches *both* additions to the union without samples *and* stale sample entries when a subclass is removed. This is the precise dual-sided drift check #1191's acceptance criteria asked for, and separating it from the async parametrized test sidesteps the asyncio-marker contamination issue documented at line 57. — `tests/nats/test_render_event_wire_roundtrip.py:114`

**praise:** Publishing via the real `NatsChannelProxy.send_streaming` (rather than calling the codec directly) and subscribing on the same connection with an explicit `nc.flush()` before publish (closing the SUB/PUB race) is the textbook real-NATS integration pattern. Exercises the full encode → publish → wire → receive → decode chain — exactly the scope PR #1190 demonstrated was uncovered. — `tests/nats/test_render_event_wire_roundtrip.py:201`

### Summary

| Severity | Count |
|---|---|
| Blockers | 0 |
| Suggestions | 3 |
| Follow-ups | 2 |
| Praise | 2 |

**Verdict: Approve with comments.** Wire-boundary regression guard is correct, complete (16/16 members + fail-closed), and runs in ~3.5 s. The JSON-swallow diagnostic improvement is the most valuable suggestion — pick up either in this PR or as a tiny follow-up.

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 2
**Skipped:** 1
**Failed:** 0

### Applied (1b1)

- [applied] **suggestion(non-blocking):** Surface malformed NATS frames via a `parse_errors` list + post-loop assertion (tester 78% ∪ security-auditor 68%) — `tests/nats/test_render_event_wire_roundtrip.py:196`. `_handler` now catches `UnicodeDecodeError`/`json.JSONDecodeError`, records the offending bytes, and signals `done` so the post-loop assertion fails with the raw frame instead of the test hanging to the 5 s timeout.
- [applied] **suggestion(non-blocking):** Drop unreachable inline `pytest.fail` guard (tester 70%) — `tests/nats/test_render_event_wire_roundtrip.py:183`. `test_sample_table_matches_render_event_union` is the single source of fail-closed enforcement; comment near `_SAMPLE_BY_TYPE` updated to reflect that.

### Skipped

- **suggestion(non-blocking):** Module-scoped `nc` fixture (tester 65%) — `tests/nats/test_render_event_wire_roundtrip.py:171`. Total suite runtime is ~3.5 s already; the per-test isolation cost is negligible and overriding the shared fixture locally would introduce a foot-gun for future tests in this module. Keeping `nc` function-scoped as defined in `tests/nats/conftest.py`.

### Verification

- `uv run pytest tests/nats/test_render_event_wire_roundtrip.py` → 17/17 pass in 3.55 s
- `uv run ruff check tests/nats/test_render_event_wire_roundtrip.py` → clean
- `uv run pyright tests/nats/test_render_event_wire_roundtrip.py` → 0 errors

Issue #1253, opened 2026-05-19
fix(cli-nats-driver): sanitize str(exc) in LlmResult.error (sibling of #1212/#1219)


MickaelV0:
## Context

Surfaced during /code-review on PR #1250 (closes #1219). Backend-dev flagged a fourth structurally identical bus-bound leak in `src/lyra/llm/drivers/cli_nats.py:161` — the `CliNatsDriver.complete()` error path. #1212 closed `lyra.nats.nats_llm_client` but missed `lyra.llm.drivers.cli_nats`, which has the same wire-error pattern.

## Current state

`src/lyra/llm/drivers/cli_nats.py:161`:

```python
except nats.errors.Error as exc:
    log.warning("cli_nats: complete() transport error [pool:%s]: %s: %s",
                pool_id, type(exc).__name__, exc)
    return LlmResult(
        error=f"NATS transport error: {exc}",  # ← bus-bound `str(exc)`
        retryable=True,
    )
```

`nats.errors.Error.__str__` can embed server addresses (e.g. `goose-logarithm.ts.net:4222`) and connection metadata. `LlmResult.error` is forwarded to the hub and potentially into user-visible error renders.

## Fix pattern

Apply the same #1212/#1215/#1219 pattern:

```python
except nats.errors.Error as exc:
    log.warning(
        "cli_nats: complete() transport error [pool:%s]: %r",
        pool_id, exc,  # %r for repr-style log, never on bus
    )
    return LlmResult(
        error=f"NATS transport error: {type(exc).__name__}",
        retryable=True,
    )
```

Also change the existing log format from `%s: %s` (calls `__str__`) to `%r` for parity with the sibling pattern.

## Acceptance

- [ ] `cli_nats.py:161` switched to `type(exc).__name__` form
- [ ] log.warning uses `%r` for `exc` (matching siblings — log-only impact but established convention)
- [ ] 1 regression test injecting a sentinel into the NATS error and asserting it does NOT surface in `LlmResult.error`
- [ ] Falsification check: test must fail if reverted to `{exc}`

## Cross-refs

- Surfaced by: PR #1250 /code-review (backend-dev, C=98%)
- Siblings: #1212 (`lyra.nats.nats_llm_client`), #1215 (`clipool_worker`), #1219 (`cli_streaming_parser` path b), [linked above] (cli_streaming_parser path a)

Issue #1212, opened 2026-05-15
fix(nats): sanitize str(exc) in error_text — bus-visible info-leak via RunErrorRenderEvent (#1210 iter-2 security)


MickaelV0:
## Context

PR #1210 review iter-2 (security-auditor) surfaced a pre-existing info-leak that PR #1210 made bus-visible.

In \`src/lyra/nats/nats_llm_client.py\`, 5 sites construct user-facing error messages by embedding \`str(exc)\` directly:

\`\`\`
nats_llm_client.py:461   error_msg = f"NATS no responders: {exc}"
nats_llm_client.py:482   error_msg = f"LLM request payload too large: {exc}"
nats_llm_client.py:496   error_msg = f"NATS transport error: {exc}"
nats_llm_client.py:519   error_msg = f"LLM stream timed out: {exc}"
nats_llm_client.py:538   error_msg = f"LLM stream: malformed chunk: {exc}"
\`\`\`

These strings flow into \`ResultLlmEvent(is_error=True, error_text=error_msg)\`. NATS / asyncio / httpx exceptions routinely embed server addresses, subject names, SSL cert paths, and connection metadata in their \`__str__\` representation.

### How PR #1210 changes the threat model

Before #1210, \`error_text\` was consumed only by local adapter-side rendering. PR #1210 (B1 fix) now forwards \`error_text\` into \`RunErrorRenderEvent.message\` in the StreamProcessor's soft-error path. \`RunErrorRenderEvent.message\` is published on the NATS bus where any subscriber can read it.

## Fix pattern

Apply the same discipline already used at \`stream_processor.py\` RunErrorRenderEvent emission site (line 380-381) and \`classify_stream_error\` (post-#1210):

\`\`\`python
# Before
error_msg = f"NATS no responders: {exc}"

# After
log.warning("NATS no responders: %r", exc)  # full str only in logs
error_msg = f"NATS no responders: {type(exc).__name__}"  # user-facing
\`\`\`

## Affected sites

| Line | Pattern | Exception types (typical) |
|---|---|---|
| 461 | \`f"NATS no responders: {exc}"\` | \`NoRespondersError\` |
| 482 | \`f"LLM request payload too large: {exc}"\` | \`nats.errors.MaxPayloadError\` |
| 496 | \`f"NATS transport error: {exc}"\` | \`nats.errors.Error\`, \`nats.errors.ConnectionClosedError\` |
[...]

PR #1259, opened 2026-05-19
fix(cli-nats-driver): sanitize str(exc) in _stream_gen_llm (#1256)

### LINKED ISSUE #1256: fix(cli-nats-driver): sanitize str(exc) in streaming path _stream_gen_llm (sibling of #1253)
## Context

Surfaced during /code-review on PR #1255 (closes #1253). Backend-dev flagged the **streaming sibling** of the leak just fixed in `complete()`: `CliNatsDriver._stream_gen_llm` delegates to `self._stream_gen(self.SUBJECT_CMD, payload)` with no `except nats.errors.Error` guard.

#1253 was explicitly scoped to `complete()` (the synchronous request path). The streaming path in the same file has the same vulnerability class — 5th sibling in the #1212/#1215/#1219/#1253 chain.

## Current state

`src/lyra/llm/drivers/cli_nats.py:102`:

```python
async def _stream_gen_llm(self, pool_id, text, model_cfg, system_prompt) -> AsyncIterator[LlmEvent]:
    payload = self._build_cmd_payload(pool_id, text, model_cfg, system_prompt, stream=True)
    async for chunk in self._stream_gen(self.SUBJECT_CMD, payload):  # ← no nats.errors.Error guard
        ...
```

If `_stream_gen` (base class method) raises `nats.errors.Error`, the exception propagates uncaught. `str(exc)` carries server addresses (e.g. `goose-logarithm.ts.net:4222`) into:

1. Tracebacks (potentially logged + bus-forwarded depending on caller)
2. Higher-level error handlers that may stringify the exception onto user-visible surfaces

Unlike `complete()`, this is a generator — the fix shape is "yield a sanitized error event + return", not "return a sanitized LlmResult".

## Fix pattern

Mirror `nats_llm_client.py:430–482` (the streaming sibling of #1212):

```python
async def _stream_gen_llm(self, ...):
    payload = self._build_cmd_payload(...)
    try:
        async for chunk in self._stream_gen(self.SUBJECT_CMD, payload):
            ...
    except nats.errors.Error as exc:
        log.warning(
            "cli_nats: _stream_gen_llm() transport error [pool:%s]: %r",
            pool_id, exc,
        )
        yield ResultLlmEvent(
            is_error=True,
            error_text=f"NATS transport error: {type(exc).__name__}",
            ...
        )
        return
```

## Acceptance

- [ ] `_stream_gen_llm` wraps the `async for` in `try/except nats.errors.Error`
- [ ] `error_text` on the yielded `ResultLlmEvent` uses `type(exc).__name__`, never `str(exc)` / `{exc}`
- [ ] `log.warning` uses `%r` for `exc` (sibling parity)
- [ ] 1 regression test injecting a sentinel into the NATS error and asserting it does NOT surface in any yielded `ResultLlmEvent.error_text`
- [ ] Falsification check: test must fail if reverted to a raw `str(exc)` form

## Cross-refs

- Surfaced by: PR #1255 /code-review (backend-dev, C=88%)
- Siblings: #1212 (nats_llm_client complete), #1215 (clipool_worker), #1219 (cli_streaming_parser path b), #1252 (cli_streaming_parser path a), #1253 (cli_nats complete) ← parent

MickaelV0:
## Summary
- `_stream_gen_llm` wraps the `async for` over `_stream_gen` in `try/except nats.errors.Error`; on transport error, yields a terminal `ResultLlmEvent(is_error=True, error_text=f"NATS transport error: {type(exc).__name__}")` and returns.
- 5th sibling in the #1212/#1215/#1219/#1253 sanitization chain — `nats.errors.Error.__str__` can embed server addresses / connection metadata, which previously could surface to user-facing renderers via the event bus when `_stream_gen` raised.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1256: fix(cli-nats-driver): sanitize str(exc) in streaming path _stream_gen_llm (sibling of #1253) | OPEN |
| Implementation | 1 commit on `fix/1256-sanitize-stream-gen-llm` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (1 new regression test, 4097 passed full suite) | Passed |

## Test Plan
- [x] New regression test `test_stream_transport_error_sanitizes_exc_message` injects a sentinel (`goose-logarithm.ts.net:4222-LEAK-SENTINEL`) into the raised `nats.errors.Error` and asserts:
  - `ResultLlmEvent.error_text` does **not** contain the sentinel,
  - `error_text == "NATS transport error: Error"` (sanitized class-name form — falsifies any revert to `{exc}` / `str(exc)`).
- [x] Existing 42 cli_nats driver tests remain green.

## Notes

- `cli_nats.py` is added to `tools/file_exemptions.txt` at 308 lines (was 289). Most of the +19 lines is mechanical ruff re-wrap of the loop body at +4 indent inside the new `try:`; only the `except` block (log.warning + sanitized `yield ResultLlmEvent` + `return`) is genuinely new code.

Closes #1256

---


MickaelV0:
## Code Review

3 fresh agents (security-auditor, backend-dev, tester) reviewed `origin/staging...HEAD` on PR #1259 (S-tier sibling fix in the #1212/#1215/#1219/#1253 chain). 3 files, 161 diff lines, no chunking needed.

### Verdict: Approve with comments

No recall findings, no blockers introduced by this PR. The sanitization fix itself is correct (`type(exc).__name__` only, `%r` in logs, terminal `ResultLlmEvent(is_error=True)`). The regression test's falsification is real and non-tautological — it would fail under `{exc}`, `str(exc)`, `repr(exc)`, and uncaught-propagation regressions.

---

### Suggestions (non-blocking — defer or follow up)

**suggestion(non-blocking): `worker_error` envelope absent on the new terminal event**
  `src/lyra/llm/drivers/cli_nats.py:151–155`
  -- backend-dev
  Root cause: `ResultLlmEvent.error_text`'s docstring states it is "always co-populated with `worker_error` on terminal failure events" (`events.py:115`). The sibling `nats_llm_client.py` yields `WorkerError(code="transport.error", retryable=True, …)` on every error path. This PR yields `error_text` only — but so does the **pre-existing** worker-reported-error branch a few lines above (the result-chunk handler), so adopting `worker_error` here without harmonising the sibling site would just trade one inconsistency for another. Recommend a separate, broader follow-up that fixes both yield sites at once.
  Confidence: 92% (gap is real) / scope: deferred to follow-up

**suggestion(non-blocking): parametrize over a real `NoRespondersError` subclass**
  `tests/llm/drivers/test_cli_nats_driver.py:241`
  -- tester
  Root cause: Test hard-codes the base-class name (`"NATS transport error: Error"`). In production the most likely transport errors are `nats.errors.NoRespondersError` and `nats.errors.TimeoutError`; a parametrized case would prove the class-name encoding survives subclass dispatch.
  Confidence: 80%

**nitpick: log.warning `%r` format is the only safeguard for the log-side leak**
  `tests/llm/drivers/test_cli_nats_driver.py:241`
  -- tester
  Root cause: A revert from `%r` to `%s` in the log call would re-introduce `str(exc)` into the log record (a separate, lower-impact info-leak vector). The sibling `complete()` test also omits this check, so the gap is convention-consistent — flagging for awareness, not blocking.
  Confidence: 75%

**nitpick: mid-stream failure path is untested**
  `tests/llm/drivers/test_cli_nats_driver.py:241`
  -- tester
  Root cause: The test raises before yielding. Adding a case that yields one `TextLlmEvent` then raises would confirm partial-output + terminal-error semantics — a real observable behaviour. Out-of-scope for the sanitization issue but worth a follow-up.
  Confidence: 70%

**nitpick: exemption-comment line accounting**
  `tools/file_exemptions.txt:20`
  -- backend-dev
  Root cause: The comment attributes ~4 lines to ruff re-wrap; counting the diff, the re-indented body is line-neutral and the `except` block contributes ~10 lines. The comment is informative but slightly miscounts.
  Confidence: 78%

---

### Notes (¬action)

- **`complete()` still leaks `str(exc)` at lines 175/180** — flagged by both security-auditor (C=82%) and backend-dev (C=97%). Confirmed **not introduced by this PR**; this is the explicit scope of **#1253** (the parent of #1256 in the chain). The exemption-line comment cross-references #1253 correctly. Leaving for that issue.
- **try/except scope** is correct — `_build_cmd_payload` is local Pydantic only, no NATS I/O. ✓
- **`yield` then `return` in async-generator `except`** is safe per CPython semantics, matches the sibling pattern in `nats_llm_client.py`. ✓
- **`nats.errors.Error` base catch** subsumes `NoRespondersError` / `TimeoutError` — sufficient for sanitization scope; the lost `code` discriminator only matters once `worker_error` is added (follow-up).

---

### Praise

praise: regression test's falsification is exact-equality on `"NATS transport error: Error"` — falsifies revert to `{exc}`, `str(exc)`, `repr(exc)`, and uncaught propagation.
  -- tester / security-auditor

MickaelV0:
## Review Fixes Applied

**Applied (1b1, batched):** 5 finding(s)
**Deferred (out of scope):** 1 (complete() leak — owned by #1253)

Pushed in 298181ed.

### Applied
- [applied] suggestion(non-blocking): **worker_error envelope on the new yield** — `src/lyra/llm/drivers/cli_nats.py:151-161` (backend-dev, C=92%) — added `WorkerError(code="transport.error", message=error_msg, retryable=True)` alongside `error_text`, matching the `nats_llm_client.py:484` pattern and the `ResultLlmEvent` "always co-populated" contract. Scope kept tight: only the new yield is touched; the pre-existing parallel result-chunk yield is left to a broader follow-up (per backend-dev's note "harmonise both sites at once").
- [applied] suggestion(non-blocking): **parametrize over `NoRespondersError`** — `tests/llm/drivers/test_cli_nats_driver.py` (tester, C=80%) — test now runs against both `nats.errors.Error` (base) and `nats.errors.NoRespondersError` (the most-frequent transport error in production).
- [applied] nitpick: **log-side `%r` assertion** — same test file (tester, C=75%) — added `mock_log.warning.assert_called_once()`, verified `"%r" in log_format`, and verified `sentinel not in log_format`. A silent `%r → %s` revert would now fail the test.
- [applied] nitpick: **mid-stream failure test** — `test_stream_transport_error_after_partial_yield` (tester, C=70%) — yields a `TextLlmEvent` then raises; asserts the partial event surfaces *and* the sanitized terminal event follows. Confirms `try/except` scope wraps the full `async for` body.
- [applied] nitpick: **exemption-comment line accounting** — `tools/file_exemptions.txt:20` (backend-dev, C=78%) — rewritten to credit the full +17 lines to the except block (log + sanitized yield + WorkerError envelope + return); re-indented loop body recorded as line-neutral. File now 315 lines (up from 308 after WorkerError addition).

### Deferred (out of scope)
- **`complete()` `str(exc)` leak at lines 175/180** — flagged by security-auditor (C=82%) and backend-dev (C=97%). This is the explicit scope of **#1253** (parent of #1256 in the chain); not introduced by this PR. The exemption-line comment already cross-references #1253. Leaving for that issue to land.

### Verification
- `uv run ruff check . && uv run pyright src/lyra/llm/drivers/cli_nats.py tests/llm/drivers/test_cli_nats_driver.py`: green.
- `uv run pytest`: **4099 passed**, 31 skipped, 2 xfailed (was 4097 before; +2 from the parametrize expansion + mid-stream test).

Re-running `/code-review` next.

MickaelV0:
## Review Fixes Applied (iter 2)

Fresh re-review on iter-1 fix (commit 298181ed) caught a real tautology I introduced and a missing assertion. Both fixed in 4d3e7ccb.

### Applied
- [applied] **issue (test-tautology): `NoRespondersError` sentinel guard was vacuous** — tester
  Root cause: `nats.errors.NoRespondersError.__str__` returns the hardcoded `"nats: no responders available for request"` and ignores constructor args; same for `nats.errors.TimeoutError`. The sentinel-not-in-error_text assertion was therefore vacuously true for those parameterised cases regardless of sanitization — only the `nats.errors.Error` base-class case actually falsified a `{exc}` revert.
  Fix: split into two tests:
    1. `test_stream_transport_error_sanitizes_exc_message` — base class only (sentinel assertion remains real).
    2. `test_stream_transport_error_encodes_subclass_name` — parametrized over `(NoRespondersError, TimeoutError)`, asserts only what is real for them (class-name dispatch via `type(exc).__name__`).
- [applied] **issue: mid-stream test missing `worker_error` assertions** — tester
  Root cause: `test_stream_transport_error_after_partial_yield` covered `is_error`/`error_text`/sentinel but skipped the WorkerError envelope asserts present in the sibling test. Asymmetric coverage would let a future refactor silently drop the envelope on the partial-yield branch.
  Fix: added `code` / `message` / `retryable` assertions matching the sibling.
- [applied] **nitpick: exemption-line count was off by 3** — backend-dev
  Root cause: comment said "~17 lines"; actual except block from `except` through `return` is 20 lines. Declared cap (315) was already correct.
  Fix: updated rationale to "20 lines" with breakdown.

### Verification
- `uv run ruff check . && uv run ruff format --check tests/llm/drivers/test_cli_nats_driver.py`: green.
- `uv run pytest tests/llm/drivers/test_cli_nats_driver.py`: **46 passed** (+1 over iter-1; split test produced 1 base + 2 subclass cases vs prior 2 parametrized cases, and the mid-stream test still counts as 1).

### Confirmations from re-review (no action)
- WorkerError construction matches `nats_llm_client.py:484` sibling pattern exactly.
- Scope contained — parallel result-chunk yield and `complete()` untouched.
- Import order alphabetic, layer-clean.
- AAA structure delineated.
- `mock_log.warning.assert_called_once()` catches silent-drop; `%r in log_format` catches `%r → %s` revert. Both load-bearing.

PR #1324, opened 2026-05-22
perf(image): split lyra runtime into svc + agent stages, slim hub/telegram/discord (#1095)

### LINKED ISSUE #1095: perf(image): slim non-clipool services from the Claude-agent base (~5GB pull savings)
## Update 2026-05-18 — Reframed as hygiene debt

**Initial framing was wrong.** This is not "park until bandwidth pressure". The bots (Telegram, Discord) **inherit a ~5GB Claude-agent base image** they don't need. Every deploy/restart pulls this excess. It's a permanent operational waste, not a conditional optimization.

**Where the fix lives:** The base image is built in the **`roxabi-container` repo** (out-of-tree from lyra). The Lyra side is just consumption — each `deploy/quadlet/lyra-*.container` references the published image tag.

**Concrete action plan:**
- **In `roxabi-container`:** decide between (a) split into 2 base images (heavy `claude-agent` for clipool worker + light `lyra-runtime` for bots/hub), or (b) multi-stage Dockerfile that produces both as variants from one source.
- Option (a) is simpler ops; option (b) keeps a single source-of-truth.
- **In `lyra` (this issue's locus):** once the slim image is published from `roxabi-container`, update `deploy/quadlet/lyra-hub.container`, `lyra-telegram.container`, `lyra-discord.container` to reference the slim variant. Keep `lyra-clipool.container` on the heavy variant.
- Track this as a cross-repo coordination — file or link the companion issue in `roxabi-container` when work begins.

**Effort:** F-lite cross-repo (most work in `roxabi-container`).

**Status:** KEEP open. The "revisit when bandwidth pressure" framing in the original body is **superseded** — this is hygiene infra debt, schedule when next slot in `roxabi-container` opens.

---

## Context

Lyra's runtime image `ghcr.io/roxabi/lyra:staging` is **2.43 GB**. Profiled post-#1078 (commit `26f4780c`):

| Layer | Size | Source |
|---|---|---|
| `bun` (npm -g) | **573 MB** | base image `ghcr.io/roxabi/base:latest` |
| `claude-code` CLI (npm -g) | **323 MB** | base image |
| `uv python 3.12` | 102 MB | base image |
| apt deps (`git`, `gh`, `ripgrep`, `jq`, `procps`, …) | 71 MB | base image |
| `uv` binary | 58 MB | base image |
| corepack (yarn + pnpm) | 23 MB | base image |
| **Lyra Python venv (`/app`)** | 82 MB | Lyra Dockerfile |
| Lyra additions (gh-token tools, socat, helper user) | ~3 MB | Lyra Dockerfile |

## Problem

The base `ghcr.io/roxabi/base:latest` is purpose-built as a **Claude Code agent base** — it ships `bun`, `claude-code`, and a full dev toolbox because `lyra-clipool` runs `claude` as a subprocess. Right-sized for clipool.

But the **same image** runs `lyra-hub`, `lyra-telegram`, `lyra-discord` — none of which need `claude`, `bun`, or the dev toolchain. They're paying ~1.7 GB of dead weight each.

Cumulative pull cost on a fresh M₁ provision: ~2.43 GB × 4 services ≈ **10 GB**, of which ~7 GB could be saved on the 3 non-clipool services.

### (a) Split base into two images

Add `ghcr.io/roxabi/base-svc:latest` (no Claude/bun/dev toolbox, ~700 MB target) alongside the existing agent base.

- Lyra Dockerfile multi-stages: `FROM base-svc AS svc-runtime`, `FROM base-agent AS agent-runtime`
- Pull `:staging-svc` for hub/telegram/discord, `:staging-agent` for clipool
- **Pros:** clean separation; svc image stays minimal forever; agent image evolves with Claude/bun churn independently
- **Cons:** 2 base image registries to maintain; CI builds 2 tags; `roxabi-container` repo grows

### (b) Multi-stage runtime in Lyra Dockerfile only

Single base image, but Lyra Dockerfile produces two final stages (`svc-runtime` strips Claude/bun/etc; `agent-runtime` keeps them). Publish 2 image tags from one build.

- **Pros:** no `roxabi-container` change needed; one place to maintain
- **Cons:** `RUN rm -rf /usr/local/lib/node_modules/{bun,@anthropic-ai}` etc. is brittle (npm-managed paths can shift); image layers are append-only, so deletes don't reclaim space without `--squash`

### (c) Status quo

Accept 2.43 GB everywhere. Re-pulls are cheap on the same host (Podman dedupes layers).

- **Pros:** zero work
- **Cons:** fresh provisions and CI runners pay 4× the size; cold starts slower

## Recommendation

**(a) Split base.** Cleanest separation, durable, follows least-privilege/least-bytes principle for service vs agent runtimes. ~1 day of work in `roxabi-container` + 1-line change per Lyra service `.container` Quadlet (different image tag).

(c) is fine for solo-dev today; revisit when:
- Bandwidth-constrained host joins (M₂, future cloud node)
- CI minutes pressure from large image pulls
- Pull frequency × image size starts hurting deploy velocity

## Acceptance

When un-deferred, option (a):
1. `ghcr.io/roxabi/base-svc:latest` ≤ 800 MB (no node, no claude-code, no bun, minimal apt set: `git`, `ca-certificates`, `tini`, `curl`, `uv`, `python3.12`)
2. `ghcr.io/roxabi/base-agent:latest` = current contents (renamed from `base:latest`)
3. Lyra service Quadlets target `:staging-svc` (hub/telegram/discord) and `:staging-agent` (clipool)
4. Cold-pull on a fresh host: ≤ 1 GB total for the 3 svc containers (vs ~7.3 GB today)
5. Smoke test: hub/telegram/discord launch + minimum healthcheck pass
6. Migration path documented: per-service Quadlet `Image=` flip + `make quadlet-install`

## Cross-references

- #1078 — parent (revealed the bloat during post-merge profiling)
- `Roxabi/roxabi-container` repo — base image source
- `docs/ops/container-publishing.md` — current publish pattern

MickaelV0:
## Summary
- Split `Dockerfile` single `runtime` stage into `svc-runtime` (FROM `base-svc:latest`, no claude/bun/lyra-gh tooling) and `agent-runtime` (FROM `base:latest`, full toolchain, last stage = default)
- Add `publish-svc` inline CI job to publish `ghcr.io/roxabi/lyra:staging-svc` in parallel with the existing reusable workflow; uses GHA cache scope `lyra-svc` to avoid key collision
- Flip `lyra-hub`, `lyra-telegram`, `lyra-discord` Quadlets to `Image=ghcr.io/roxabi/lyra:staging-svc`; `lyra-clipool` and `lyra-gh-helper` unchanged

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1095: perf(image): slim non-clipool services | Open |
| Frame | [1095-slim-non-clipool-services-frame.mdx](artifacts/frames/1095-slim-non-clipool-services-frame.mdx) | Approved |
| Spec | [1095-slim-non-clipool-services-spec.mdx](artifacts/specs/1095-slim-non-clipool-services-spec.mdx) | Approved |
| Plan | [1095-slim-non-clipool-services-plan.mdx](artifacts/plans/1095-slim-non-clipool-services-plan.mdx) | Approved |
| Implementation | 3 commits on `feat/1095-slim-non-clipool-services` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (4219 passed) | Passed |

## Test Plan
- [ ] Push to staging → CI publishes both `ghcr.io/roxabi/lyra:staging` and `ghcr.io/roxabi/lyra:staging-svc` (requires `base-svc:latest` available in GHCR first)
- [ ] `docker build --target svc-runtime` locally once `base-svc:latest` is published
- [ ] `docker run --rm <svc-image> which bun claude node lyra-gh` → all exit non-zero (tools absent)
- [ ] `docker inspect --format='{{.Size}}' <svc-image>` → ≤ 838860800 bytes
- [ ] `lyra config validate` passes on hub/telegram/discord with svc image (HEALTHCHECK)
- [ ] clipool + lyra-gh-helper smoke test unchanged

> ⚠️ **Upstream gate:** `ghcr.io/roxabi/base-svc:latest` must be published from `roxabi-container` before merging — the `publish-svc` CI job will fail without it. Merge as a unit once the companion issue lands.

Closes #1095

---

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,3 +12,24 @@ jobs:
     with:
       image_name: ghcr.io/roxabi/lyra
       release_please_component: lyra
+  publish-svc:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: docker/setup-buildx-action@v3
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: docker/build-push-action@v5
+        with:
+          context: .
+          target: svc-runtime
+          push: true
+          tags: ghcr.io/roxabi/lyra:staging-svc
+          cache-from: type=gha,scope=lyra-svc
+          cache-to: type=gha,mode=max,scope=lyra-svc
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -34,3 +34,7 @@ jobs:
           tags: ghcr.io/roxabi/lyra:staging-svc
           cache-from: type=gha,scope=lyra-svc
           cache-to: type=gha,mode=max,scope=lyra-svc
+      - name: Verify forbidden binaries absent (SC-3)
+        run: |
+          docker run --rm ghcr.io/roxabi/lyra:staging-svc \
+            sh -c 'which bun claude corepack node lyra-gh 2>/dev/null && exit 1 || exit 0'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -13,19 +13,20 @@ jobs:
       image_name: ghcr.io/roxabi/lyra
       release_please_component: lyra
   publish-svc:
+    if: github.ref_type == 'branch'
     runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
     steps:
-      - uses: actions/checkout@v4
-      - uses: docker/setup-buildx-action@v3
-      - uses: docker/login-action@v3
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
+      - uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - uses: docker/build-push-action@v5
+      - uses: docker/build-push-action@ca052bb54ab0790a636c9b5f226502c73d547a25 # v5
         with:
           context: .
           target: svc-runtime
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ COPY src/ src/
 RUN uv sync --frozen --no-dev
 
 # ── Slim service runtime (hub, telegram, discord) ───────────────────────────
+# TODO: pin base-svc by digest — track alongside base:latest pinning issue
 FROM ghcr.io/roxabi/base-svc:latest AS svc-runtime
 
 USER root
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,27 @@ COPY packages/ packages/
 COPY src/ src/
 RUN uv sync --frozen --no-dev
 
-# ── Runtime stage ────────────────────────────────────────────────────────────
-FROM ghcr.io/roxabi/base:latest AS runtime
+# ── Slim service runtime (hub, telegram, discord) ───────────────────────────
+FROM ghcr.io/roxabi/base-svc:latest AS svc-runtime
+
+USER root
+
+# UID 1500 pinned per ADR-053 (Quadlet container UID stability)
+RUN useradd -u 1500 -m lyra
+
+COPY --from=builder --chown=lyra:lyra /app /app
+
+WORKDIR /app
+
+ENV PATH="/app/.venv/bin:$PATH"
+
+USER lyra
+
+HEALTHCHECK --interval=30s --timeout=10s --start-period=30s --retries=3 \
+  CMD lyra config validate || exit 1
+
+# ── Agent runtime (clipool — full gh_token tooling) ───────────────────────────
+FROM ghcr.io/roxabi/base:latest AS agent-runtime
 
 USER root
 
--- a/deploy/quadlet/lyra-discord.container
+++ b/deploy/quadlet/lyra-discord.container
@@ -10,7 +10,7 @@ StartLimitBurst=5
 [Container]
 # Image digest pinning for localhost/lyra:latest tracks on a follow-up issue
 # (CI tag-by-SHA strategy).
-Image=ghcr.io/roxabi/lyra:staging
+Image=ghcr.io/roxabi/lyra:staging-svc
 Label=io.containers.autoupdate=registry
 ContainerName=lyra-discord
 Network=roxabi.network
--- a/deploy/quadlet/lyra-hub.container
+++ b/deploy/quadlet/lyra-hub.container
@@ -9,7 +9,7 @@ StartLimitBurst=5
 # Image digest pinning for localhost/lyra:latest tracks on a follow-up issue
 # (CI tag-by-SHA strategy). Upstream registry images (e.g. nats.container)
 # are pinned by digest directly.
-Image=ghcr.io/roxabi/lyra:staging
+Image=ghcr.io/roxabi/lyra:staging-svc
 Label=io.containers.autoupdate=registry
 ContainerName=lyra-hub
 Network=roxabi.network
--- a/deploy/quadlet/lyra-telegram.container
+++ b/deploy/quadlet/lyra-telegram.container
@@ -10,7 +10,7 @@ StartLimitBurst=5
 [Container]
 # Image digest pinning for localhost/lyra:latest tracks on a follow-up issue
 # (CI tag-by-SHA strategy).
-Image=ghcr.io/roxabi/lyra:staging
+Image=ghcr.io/roxabi/lyra:staging-svc
 Label=io.containers.autoupdate=registry
 ContainerName=lyra-telegram
 Network=roxabi.network
--- a/docs/ops/container-publishing.md
+++ b/docs/ops/container-publishing.md
@@ -30,6 +30,12 @@ caller workflow that feeds project-specific inputs.
   > silently drop this instruction. The reusable workflow sets `oci-mediatypes=false` on the
   > `docker/build-push-action` step to force Docker v2 schema 2, so `HEALTHCHECK` is preserved
   > in the published image. No action needed in the Dockerfile or caller workflow.
+
+  > **svc-runtime (`staging-svc`) requires `config.toml` bind-mount:** `lyra config validate`
+  > opens `config.toml` from `WORKDIR /app` and exits 1 on `FileNotFoundError`. In production
+  > Quadlets the file is bind-mounted, so this is fine. Running `docker run` without the mount
+  > (local dev, CI smoke tests) will immediately mark the container unhealthy — this is expected
+  > behaviour, not a bug.
 - **OCI labels** — do not set `org.opencontainers.image.*` labels in the Dockerfile. They are
   injected at build time by `docker/metadata-action@v5` in the reusable workflow, ensuring labels
   always match the actual pushed tag and commit SHA.
--- a/docs/ops/container-publishing.md
+++ b/docs/ops/container-publishing.md
@@ -129,10 +129,11 @@ containers. No manual intervention is needed after a staging merge.
 
 | Container | Image | AutoUpdate |
 |---|---|---|
-| lyra-hub | `ghcr.io/roxabi/lyra:staging` | registry |
-| lyra-telegram | `ghcr.io/roxabi/lyra:staging` | registry |
-| lyra-discord | `ghcr.io/roxabi/lyra:staging` | registry |
+| lyra-hub | `ghcr.io/roxabi/lyra:staging-svc` | registry |
+| lyra-telegram | `ghcr.io/roxabi/lyra:staging-svc` | registry |
+| lyra-discord | `ghcr.io/roxabi/lyra:staging-svc` | registry |
 | lyra-clipool | `ghcr.io/roxabi/lyra:staging` | registry |
+| lyra-gh-helper | `ghcr.io/roxabi/lyra:staging` | registry |
 | voicecli-tts | `ghcr.io/roxabi/voicecli-tts:staging` | registry |
 | voicecli-stt | `ghcr.io/roxabi/voicecli-stt:staging` | registry |
 | lyra-nats | pinned by digest | none (pinned) |
@@ -158,6 +159,11 @@ podman auto-update
   `podman login --get-login ghcr.io`.
 - Podman does not auto-rollback on startup failure. A bad image enters a restart loop
   (`Restart=on-failure`). Check with `podman ps` or `journalctl --user -u <unit>`.
+- **Parallel publish inconsistency window (known):** `publish` (`:staging`) and `publish-svc`
+  (`:staging-svc`) jobs run in parallel — if one fails, the other may still push. During that
+  window, autoupdate can pull one variant while the other is stale. Accepted risk at staging;
+  long-term fix is a single `docker buildx bake` job that pushes both targets atomically —
+  tracked in #1325.
 
 ---
 


MickaelV0:
## Code Review

### Blockers

**issue(blocking): `publish-svc` fires on `lyra/v*` tag pushes and always writes `:staging-svc` — no versioned svc image on release**
  `.github/workflows/publish.yml` — `publish-svc` job, `tags:` field
  -- devops + product-lead
  Root cause: The workflow `on:` trigger covers both `branches: [staging]` and `tags: ['lyra/v*']`. `publish-svc` hardcodes `tags: ghcr.io/roxabi/lyra:staging-svc`, so a `lyra/v1.2.3` release push overwrites `:staging-svc` with the release-tagged build while never publishing `:1.2.3-svc`. Consumers that track `:staging-svc` by digest after a release get unexpected content; the release is also inconsistent (`:staging` gets a versioned tag via the reusable workflow; `:staging-svc` does not).
  Solutions:
    1. Add `if: github.ref_type == 'branch'` to `publish-svc` so it only fires on staging branch pushes (recommended)
    2. Publish a versioned svc tag on tag events via a separate `if: github.ref_type == 'tag'` step or by extending the reusable workflow
  Confidence: 95%

**issue(blocking): `publish-svc` uses floating action tags — inconsistent with SHA-pinned pattern used in other workflows**
  `.github/workflows/publish.yml` — `checkout@v4`, `setup-buildx-action@v3`, `login-action@v3`, `build-push-action@v5`
  -- devops + security-auditor
  Root cause: Other workflows in this repo pin GitHub Actions to full commit SHAs. The new `publish-svc` job uses floating major-version tags for all four actions. This is a supply-chain risk on a job that writes to GHCR with `packages: write` — a compromised action can inject code into the image publish pipeline.
  Solutions:
    1. Pin each action to its full commit SHA with an inline version comment (e.g., `docker/build-push-action@<sha> # v5.x.y`) (recommended)
    2. Add Dependabot/Renovate SHA auto-update rules if not already configured
  Confidence: 97%

---

### Warnings

**issue: HEALTHCHECK uses `lyra config validate` — exits non-zero when `config.toml` is absent (no bind-mount)**
  `Dockerfile` — `svc-runtime` HEALTHCHECK line
  -- devops + product-lead
  Root cause: `config validate` opens `config.toml` from `WORKDIR /app` and exits 1 on `FileNotFoundError`. In production Quadlets the file is bind-mounted, so it works. But in `docker run` without the mount (CI smoke tests, local dev), the container is immediately unhealthy. This is identical behaviour to `agent-runtime`; the concern is whether svc-image consumers will get adequate guidance, and whether any post-build CI smoke test will be written.
  Solutions:
    1. Document explicitly that running svc-runtime without `config.toml` mounted will fail the HEALTHCHECK; add to deployment runbook (recommended)
    2. Replace with a lighter probe that doesn't require config (e.g., `python -c "import lyra"`)
  Confidence: 88%

**issue: `svc-runtime` does not scaffold `.claude/` directories — hub/telegram/discord previously had these from `agent-runtime`**
  `Dockerfile` — `svc-runtime` stage, `RUN useradd -u 1500 -m lyra`
  -- devops
  Root cause: `agent-runtime` (the former `runtime` stage) creates `/home/lyra/.claude/{projects,plugins,skills,shared,.git}` directories. `svc-runtime` only creates the home dir via `useradd -m`. If hub, telegram, or discord code paths reference `.claude/` subdirectories (plugin discovery, vault CLI), they'll get `FileNotFoundError` at runtime. Needs auditing before deploy.
  Solutions:
    1. Grep hub/telegram/discord startup paths for `.claude/` references; add `mkdir -p` to `svc-runtime` for any required dirs (recommended)
    2. Confirm via `lyra hub --dry-run` or equivalent against the svc image that no `.claude/` subdirs are accessed
  Confidence: 75%

**suggestion(non-blocking): SC-3 (svc-runtime must not contain bun/claude/node/lyra-gh) has no CI enforcement**
  `.github/workflows/publish.yml` — `publish-svc` job
  -- product-lead
  Root cause: SC-3 is a negative criterion that can only be verified post-build. No step in `publish-svc` checks for forbidden binaries. A future `base-svc` update that reintroduces a binary would silently violate SC-3.
  Solutions:
    1. Add a post-push step: `docker run --rm ghcr.io/roxabi/lyra:staging-svc sh -c 'which bun claude corepack node lyra-gh 2>/dev/null && exit 1 || exit 0'` (recommended)
    2. Defer to deployment-time manual check — accept the risk
  Confidence: 88%

**suggestion(non-blocking): `publish-svc` and `publish` run in parallel with no `needs:` — if `publish` fails, `:staging-svc` is still pushed**
  `.github/workflows/publish.yml` — `publish-svc` job
  -- devops
  Root cause: Spec mandates parallel (no `needs:`) for speed. Trade-off is that on a publish failure, `:staging-svc` can be pushed while `:staging` is not, resulting in an inconsistent deployment window where autoupdate pulls `:staging-svc` from a build that was never fully validated. Acceptable given the staging context but worth documenting.
  Solutions:
    1. Accept as designed — document the inconsistency window in the deployment runbook (recommended)
    2. Add `needs: [publish]` to serialize and ensure consistency
  Confidence: 80%

**suggestion(non-blocking): `base-svc:latest` is an unpinned mutable tag — no digest, no tracking issue**
  `Dockerfile` — `FROM ghcr.io/roxabi/base-svc:latest AS svc-runtime`
  -- devops + security-auditor
  Root cause: Same class of issue as the pre-existing `base:latest` in `agent-runtime`. Net-new dependency with no digest pin and no tracking issue referenced. A push to `base-svc:latest` is reflected in the next build without review.
  Solutions:
    1. Add `# TODO: pin base-svc by digest — follow-up issue (alongside existing base:latest tracking issue)` comment and open a tracking issue (recommended)
    2. Pin immediately to `@sha256:<digest>`
  Confidence: 85%

---

### Praise

**praise: cache scope isolation (`scope=lyra-svc`) correctly prevents GHA cache collision with the reusable workflow's default key**
  `.github/workflows/publish.yml` — `cache-from/cache-to` in `publish-svc`
  -- devops
  Root cause: Without explicit scope, both jobs would write to the same GHA cache key and the second job's push would overwrite the first's cache. The explicit `scope=lyra-svc` is the right call.
  Confidence: 100%

**praise: `agent-runtime` correctly positioned last — reusable workflow builds it without `--target`, preserving backwards compatibility**
  `Dockerfile` — stage ordering
  -- devops
  Root cause: The reusable `publish-container.yml@v1` has no `--target` input. Keeping `agent-runtime` last ensures the existing caller workflow continues to work without modification.
  Confidence: 100%

---

### Summary

| Category | Count |
|---|---|
| Blockers | 2 |
| Warnings | 4 |
| Suggestions (non-blocking) | 2 |
| Praise | 2 |

**Verdict: Request changes** — fix blockers before merging. Warnings are strongly recommended for operational safety.

> Note: This PR is also gated on upstream `ghcr.io/roxabi/base-svc:latest` being published from `roxabi-container`. Blocker fixes can be made on the feature branch while awaiting the upstream.

---
_Generated by `/code-review`_

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 2 finding(s)
**Deferred (issues created):** 1 finding(s)
**Skipped:** 1 finding(s) — verified safe
**Failed:** 0

### Auto-Applied

- [applied] `issue(blocking)`: `publish-svc` fires on `lyra/v*` tag events → added `if: github.ref_type == 'branch'` (95%)
- [applied] `issue(blocking)`: floating GHA action tags → pinned all 4 actions to full commit SHAs (97%)
- [applied] `issue`: HEALTHCHECK mount requirement undocumented → added note to `docs/ops/container-publishing.md` (88%)
- [applied] `suggestion`: `base-svc:latest` unpinned mutable tag → added TODO comment in Dockerfile; tracking in #1325 (85%)

### Applied (1b1)

- [applied] `suggestion`: SC-3 has no CI enforcement → added post-push binary absence check step to `publish-svc` (88%)
- [applied] `suggestion`: parallel publish inconsistency window → documented in runbook Caveats with link to #1325 (`buildx bake` long-term fix) (80%)

### Deferred

- `issue`: `svc-runtime` `.claude/` directory scaffold — verified safe to skip: `grep` found zero `.claude/` references in hub/telegram/discord boot paths (75%)

### Follow-up

- #1325 — `perf(ci): unify publish + publish-svc into single buildx bake job` (atomic multi-target push, blocked by #1095)

MickaelV0:
## Code Review (iteration 2)

### Blockers

**issue(blocking): SC-3 binary-absence check has a false-negative — `which` exits 1 when any arg is missing, causing `|| exit 0` to fire even when some forbidden binaries are still present**
  `.github/workflows/publish.yml` — "Verify forbidden binaries absent (SC-3)" step
  -- security-auditor + devops + architect + product-lead
  Root cause: GNU `which bun claude corepack node lyra-gh` exits 1 if ANY of the five names is not found (even if others are found). The `&& exit 1` arm never fires, `|| exit 0` fires instead, and the step reports success. A svc image containing `bun` but not `claude` passes the gate silently. The intent is "fail if any one is present" but the implementation reads "fail only if all five are simultaneously present."
  Solutions:
    1. Loop per-binary with `command -v`: `for b in bun claude corepack node lyra-gh; do command -v "$b" 2>/dev/null && { echo "FORBIDDEN: $b"; exit 1; }; done` (recommended — POSIX, no `which` dependency, each check independent)
    2. Five separate `docker run` invocations, one per binary
  Confidence: 92%

**issue(blocking): `publish-svc` does not set `oci-mediatypes=false` — HEALTHCHECK may be silently dropped from the published `:staging-svc` image**
  `.github/workflows/publish.yml` — `build-push-action` step in `publish-svc`
  -- architect
  Root cause: `docs/ops/container-publishing.md` explicitly requires `oci-mediatypes=false` to preserve HEALTHCHECK (Docker v2 schema 2 vs OCI manifest). The reusable `publish-container.yml@v1` sets this; the inline `publish-svc` job does not. Depending on BuildKit defaults, `:staging-svc` may be published as OCI with HEALTHCHECK silently stripped — breaking SC-9 without any build error.
  Solutions:
    1. Add `outputs: type=image,oci-mediatypes=false` to the `build-push-action` inputs (recommended — mirrors reusable workflow)
    2. Add `provenance: false` — partial mitigation, less explicit
  Confidence: 90%

---

### Informational (not blockers)

- **SHA pin quality verified** — all 4 action SHAs are legitimate commit objects (not tag objects). Supply chain concern from iteration 1 is fully resolved.
- **`if: github.ref_type == 'branch'` is correct** — workflow-level `on.push.branches: [staging]` already prevents feature-branch triggers; the `if:` guard correctly excludes `lyra/v*` tag events only. (devops agent false positive — not a bug)
- **Parallel publish window** — documented in this PR, tracked in #1325. No action in this iteration.
- **`base-svc:latest` float** — acknowledged via TODO comment, same as pre-existing `base:latest`. No action required here.
- **nitpick:** `container-publishing.md` manual pull section still references `:staging` for hub/telegram/discord — these units now use `:staging-svc`. Low risk at staging but could mislead during degraded-state recovery.

---

### Praise

**praise: iteration 1 blockers correctly resolved** — tag-event guard and SHA pins are both sound.

---

### Summary

| Category | Count |
|---|---|
| Blockers | 2 |
| Informational | 4 |
| Nitpick | 1 |
| Praise | 1 |

**Verdict: Request changes** — fix 2 blockers before merging.

---
_Generated by `/code-review` (iteration 2)_

MickaelV0:

**Auto-applied (C≥80, 2+ agents):** 2 finding(s)

### Auto-Applied

- [applied] `issue(blocking)`: SC-3 binary check false-negative → replaced `which` multi-arg with per-binary `command -v` loop; each binary checked independently, step fails on first hit (92%)
- [applied] `issue(blocking)`: `oci-mediatypes=false` missing → switched from `push: true` to `outputs: type=image,push=true,oci-mediatypes=false` to force Docker v2 manifest format and preserve HEALTHCHECK in `:staging-svc` (90%)

### Applied (1b1)

- [applied] `nitpick`: manual pull section referenced `:staging` for hub/telegram/discord → updated to pull both `:staging-svc` (hub/telegram/discord) and `:staging` (clipool/lyra-gh-helper); added lyra-gh-helper to restart + verify commands (95%)

PR #1326, opened 2026-05-22
perf(ci): unify publish + publish-svc into single buildx bake job

### LINKED ISSUE #1325: perf(ci): unify publish + publish-svc into single buildx bake job (atomic multi-target push)
## Problem

\`publish\` and \`publish-svc\` jobs in \`.github/workflows/publish.yml\` run in parallel with no coordination. If \`publish\` fails, \`:staging-svc\` can still be pushed — leaving hub/telegram/discord on a new svc image while clipool/gh-helper remain on a stale \`:staging\`. podman auto-update polls independently, so the inconsistency window is real.

## Goal

Replace both jobs with a single \`docker buildx bake\` job that builds \`svc-runtime\` and \`agent-runtime\` targets atomically — either both images are pushed or neither is.

## What this requires

- Write \`docker-bake.hcl\` with two targets (\`svc-runtime\`, \`agent-runtime\`)
- Replace reusable-workflow call (\`publish-container.yml@v1\`) with an inline job that uses bake
- Preserve: versioned tags on \`lyra/v*\` pushes, OCI label injection, GHA cache per target, GHCR auth
- Verify: \`Label=io.containers.autoupdate=registry\` still works after the switch

## Origin

PR #1324 review (deferred per \`/fix\` walkthrough). Surfaced as non-blocking finding — parallel inconsistency is accepted at staging but the right long-term fix is atomic multi-target push.

MickaelV0:
## Summary
- Replace two parallel CI jobs (`publish` via reusable workflow + inline `publish-svc`) with a single `docker buildx bake` job that builds both `svc-runtime` and `agent-runtime` atomically — either both images push or neither does, eliminating the partial-push inconsistency window
- Add `docker-bake.hcl` with per-target GHA cache scopes; preserve versioned tags on `lyra/v*`, OCI label injection, and SC-3 forbidden-binary check

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1325: perf(ci): unify publish + publish-svc into single buildx bake job | Open |
| Implementation | 1 commit on `feat/1325-unify-publish-buildx-bake` | Complete |
| Verification | Lint ✅ Typecheck ✅ | Passed |

## Test Plan
- [ ] Push to `staging` branch → verify both `:staging` and `:staging-svc` tags appear on GHCR atomically
- [ ] Verify SC-3 step runs and passes (no forbidden binaries in `svc-runtime`)
- [ ] Push a `lyra/v*` tag → verify `:vX.Y.Z`, `:X`, `:latest` and svc variants all pushed
- [ ] Verify `io.containers.autoupdate=registry` label still present on pushed images (set in Quadlet `.container` files, unaffected by this change)
- [ ] Confirm GHA cache hits on second run (separate scopes: `lyra-agent`, `lyra-svc`)

Closes #1325

---

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -28,6 +28,7 @@ jobs:
               echo "agent-runtime.tags=${REG}:staging"
               echo "svc-runtime.tags=${REG}:staging-svc"
               echo "EOF"
+              echo "svc_tag=${REG}:staging-svc"
             } >> "$GITHUB_OUTPUT"
           else
             full="${GITHUB_REF_NAME}"
@@ -41,6 +42,7 @@ jobs:
               echo "svc-runtime.tags=${REG}:${ver}-svc"
               echo "svc-runtime.tags=${REG}:${major}-svc"
               echo "EOF"
+              echo "svc_tag=${REG}:${ver}-svc"
             } >> "$GITHUB_OUTPUT"
           fi
 
@@ -55,7 +57,6 @@ jobs:
             *.labels.org.opencontainers.image.version=${{ github.ref_name }}
 
       - name: Verify forbidden binaries absent (SC-3)
-        if: github.ref == 'refs/heads/staging'
         run: |
-          docker run --rm ghcr.io/roxabi/lyra:staging-svc \
+          docker run --rm ${{ steps.tags.outputs.svc_tag }} \
             sh -c 'which bun claude corepack node lyra-gh 2>/dev/null && exit 1 || exit 0'
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,16 +8,7 @@ permissions:
   packages: write
 jobs:
   publish:
-    uses: Roxabi/.github/.github/workflows/publish-container.yml@v1
-    with:
-      image_name: ghcr.io/roxabi/lyra
-      release_please_component: lyra
-  publish-svc:
-    if: github.ref_type == 'branch'
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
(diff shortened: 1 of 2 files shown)


MickaelV0:
## Code Review

### Blockers

**issue(blocking): SC-3 binary check not run on tag (`lyra/v*`) release builds**
`.github/workflows/publish.yml:57`
— security-auditor
Root cause: `if: github.ref == 'refs/heads/staging'` skips SC-3 entirely on tag pushes. Both `svc-runtime` and `agent-runtime` are built and pushed for prod without verifying forbidden binaries absent. The `:staging-svc` tag doesn't exist on tag runs, but the fix is to pull the versioned svc tag (`:${ver}-svc`) rather than skip the check.
Class: [parallel-path-drift]
Raw callsites: `{file: ".github/workflows/publish.yml", line: 57}`
Solutions:
1. Remove the `if:` guard; output a dedicated `svc_tag` from the tags step and reference it in SC-3 so the check always pulls the image that was just pushed (recommended)
2. Split into two jobs (branch vs tag) each with its own SC-3 step
Confidence: 88%

---

**issue: Multi-line `agent-runtime.tags=` in `set:` relies on undocumented bake-action array-append behavior**
`.github/workflows/publish.yml:38-40`
— devops
Root cause: `bake-action` splits `set:` on `\n` and passes each line as a separate `--set` flag. Repeated `--set target.tags=X` calls accumulate into an array in v7.2.0 but this is internal, not a stable API contract. CSV form (`target.tags=X,Y,Z`) is the explicit, version-stable alternative.
Solutions:
1. Collapse to CSV in the shell step: `echo "agent-runtime.tags=${REG}:${ver},${REG}:${major},${REG}:latest"` — single `--set` call, survives bake-action refactors (recommended)
2. Keep multi-line, accept implicit behavior (SHA already pinned)
Confidence: 75%

---

### Suggestions

**suggestion(non-blocking): Hardcoded `EOF` delimiter in GITHUB_OUTPUT multiline block**
`.github/workflows/publish.yml:27`
— devops
GitHub docs recommend a random delimiter (`openssl rand -hex 8`) to prevent early termination if a value ever matches the delimiter. Values here are controlled image refs so no current exploit path.
Solutions:
1. `DELIM=$(openssl rand -hex 8)` → `echo "bake_set<<${DELIM}"` / `echo "${DELIM}"` (recommended)
2. Add comment acknowledging accepted risk
Confidence: 85%

**nitpick: `agent-runtime` cache scope rename (`unscoped` → `lyra-agent`) will silently cause a cold rebuild**
`docker-bake.hcl:9`
— devops
Old reusable workflow used `type=gha` without `scope=` (defaults to `buildkit`). New `scope=lyra-agent` is a different cache key. First run = full rebuild with no log explaining why. Confidence: 90%

**nitpick: `svc-runtime` gets no `:latest-svc` tag on release — asymmetric with `agent-runtime`**
`.github/workflows/publish.yml:41-42`
— devops
Add a comment noting `:latest-svc` is intentionally omitted, or add it if any Quadlet unit pulls unversioned. Confidence: 70%

---

### Praise

- GITHUB_OUTPUT multiline block and brace-grouped redirect are correctly implemented
- `if: github.ref == 'refs/heads/staging'` guard on SC-3 correctly avoids pulling `:staging-svc` on tag runs where that tag doesn't exist — fix is parameterization, not removal

---

**Verdict: Request changes** — 2 blocking findings (SC-3 gap on tag builds · bake `tags` array-append reliance)

MickaelV0:
## Review Fixes Applied

**Applied via 1b1 (consensus):** 2 findings
**Skipped:** 2 findings (F3, F4 — verifier-confirmed non-issues)
**Deferred:** 0
**Failed:** 0

### Auto-Applied
- [applied] `issue(blocking)`: SC-3 binary check skipped on tag releases — removed `if:` guard, added `svc_tag` output so SC-3 always pulls the image just pushed (parallel-path-drift, C=88%+99%)

### Applied (1b1 via /consensus — unanimous panel)
- [applied] `issue`: Collapse multi-line `agent-runtime.tags=` to CSV — eliminates reliance on undocumented bake-action array-append behavior (C=75%, panel: Fix Sol 1)
- [applied] `nitpick`: Add comment noting `:latest-svc` intentionally omitted on release tags — Quadlet units pin to specific versions (C=70%, panel: Fix Sol 1)

### Skipped (consensus: unanimous Skip)
- `suggestion`: Hardcoded `EOF` delimiter — verifier (C_v=30) confirmed guidance applies to stop-commands, not GITHUB_OUTPUT; values are hardcoded image refs with no exploit path
- `nitpick`: Cache scope cold rebuild — verifier (C_v=20) confirmed `agent-runtime` had no prior cache key in this repo; cold build was always the baseline

PR #1358, opened 2026-05-25
feat(blobstore): V8 HTTP-fronted service + HttpBlobStore client

MickaelV0:
## Summary

- V8 of the BlobStore ecosystem plan (#1061): exposes the existing `FsBlobStore` over HTTP on `:8449` from a new `lyra-blobstore` Quadlet service so cross-host Protocol clients on M₂ (llm-worker, image-worker, future voice-worker) can read/write blobs persisted on M₁.
- 5 vertical slices: (S1) service skeleton + bearer auth + healthz/metrics; (S2) PUT/GET/HEAD/DELETE handlers + `HttpBlobStore` client + Protocol-parametrized equivalence test; (S3) `BlobAuditEvent` contract + degraded-mode sink + JetStream KV readiness announce; (S4) full Quadlet unit + secret bootstrap + cross-host integration scaffold; (S5) module CLAUDE.mds + rotation/backup runbooks + ADR-067 HEAD-row amendment.
- Security audit run mid-flight (`artifacts/reviews/1330-security-audit.md`) flagged 2 findings (missing 401 audit emission; HEAD/exists Literal mismatch) — both resolved in the S4 commit before merge.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1330: feat(blobs): V8 — HTTP-fronted BlobStore service + HttpBlobStore client | Open |
| Frame | [1330-v8-http-fronted-blobstore-frame.mdx](artifacts/frames/1330-v8-http-fronted-blobstore-frame.mdx) | Present (approved) |
| Analysis | [1330-v8-http-fronted-blobstore-analysis.mdx](artifacts/analyses/1330-v8-http-fronted-blobstore-analysis.mdx) | Present |
| Spec | [1330-v8-http-fronted-blobstore-spec.mdx](artifacts/specs/1330-v8-http-fronted-blobstore-spec.mdx) | Present (approved) |
| Plan | [1330-v8-http-fronted-blobstore-plan.mdx](artifacts/plans/1330-v8-http-fronted-blobstore-plan.mdx) | Present (26 micro-tasks, 13 waves) |
| Security audit | [1330-security-audit.md](artifacts/reviews/1330-security-audit.md) | APPROVED-WITH-FOLLOWUP (F-1, F-2 fixed) |
| Implementation | 5 feat/docs commits on `1330-blobstore-http` | Complete |
[...]

PR #1364, opened 2026-05-25
chore(blobstore): V8.1 hardening + cleanup pass

### LINKED ISSUE #1362: chore(blobstore): V8.1 hardening + cleanup pass (parking-lot from PR #1358)
## Origin

PR #1358 (closes #1330 — V8 HTTP-fronted BlobStore service) — parking-lot collected during `/code-review` Phase 3 multi-domain review.

Each item flagged by a single agent OR C<80% OR design tradeoff (not a bug). None blocks production. Grouped here as a single hardening + cleanup pass to avoid letting them rot in a PR comment.

## Scope

12 non-structural improvements to `src/lyra/blobstore/`, `packages/roxabi-blobs/`, `deploy/quadlet/lyra-blobstore.container`. Sort by descent expected value below; cherry-pick what fits the 30-min budget.

### High value

- [ ] **Tailscale `ExecStartPre=` fail-closed guard** (`deploy/quadlet/lyra-blobstore.container`)
  - Today: `${TAILSCALE_IPV4}` empty → `PublishPort=:8449:8449` silently binds 0.0.0.0
  - Fix: `ExecStartPre=/bin/sh -c '[ -n "${TAILSCALE_IPV4}" ] || exit 1'` so missing IP fails the unit instead of widening exposure
  - Class: residual-risk hardening

### Medium value

- [ ] **`BlobAuditSink.provision()` method** (`src/lyra/blobstore/audit_sink.py`)
  - Mirror `JetStreamAuditSink.provision(nc)` lifecycle pattern (sink owns NATS attach, not lifespan)
  - Cohérence with sibling sink in `lyra.security`
- [ ] **`connect_retry_max_s` no-op param** (`packages/roxabi-blobs/src/roxabi_blobs/http_store.py`)
  - Parameter accepted + documented but currently unused — either implement (in-band retry on connect refused) or remove from signature + docstring

### Low value (clean-up, doc, perf-marginal)

- [ ] **HEAD triple-DB-hit** (`src/lyra/blobstore/_handlers.py`) — remove redundant intermediate query in HEAD handler
- [ ] **`audit_sink` double-assignment in lifespan** (`src/lyra/blobstore/serve.py` ~line 98 + ~line 102) — assigned then immediately overwritten
- [ ] **DELETE 204↔404 TOCTOU window** — RFC 9110 §9.3.5 says safe for idempotent ops; document or accept
- [ ] **`_conn()` skips lock** (`packages/roxabi-blobs/src/roxabi_blobs/fs_store.py`) — pattern divergence vs sibling code; add comment justifying WAL-read safety
- [ ] **Deferred import in `HttpBlobStore.exists`** (`from datetime import UTC, datetime` on hot path) — hoist to module
- [ ] **`shared-modules-upper-boundary` contract** (`.importlinter`) — does not include `lyra.blobstore`; verify intent
- [ ] **ASGI lifespan drive in `HttpBlobStore._start_asgi_lifespan` undocumented** — add docstring covering the in-process test seam
- [ ] **HEAD content_hash/store_key dual-lookup undocumented** in `src/lyra/blobstore/CLAUDE.md` — document why both lookup paths exist
- [ ] **Volume path divergence vs analysis table** (`artifacts/analyses/1330-v8-http-fronted-blobstore-analysis.mdx`) — analysis table stale; runbook (`docs/QUADLET-DEPLOYMENT.md`) is authoritative

## Acceptance

- All checked items either applied (commit reference) or explicitly skipped with one-line rationale in PR description
- No regression in existing 460-test blobstore suite
- `uv run pyright`, `uv run ruff check`, `uv run lint-imports` all clean

## Out of scope

- Pre-read 413 oversized-blob gate — already deferred per `src/lyra/blobstore/CLAUDE.md` "V8 contract"
- Protocol v2 streaming get / multipart — tracked in #1334
- Dedicated `ghcr.io/roxabi/blobstore` image — three-strikes rule per ADR-073

MickaelV0:
## Summary
- Drain the PR #1358 parking-lot — 12 non-structural items from the V8 BlobStore multi-domain review (1 hardening, 2 medium API drift, 9 low cleanup/docs).
- Production-relevant: Tailscale fail-closed `ExecStartPre=` guard on the Quadlet unit; `BlobAuditSink.provision()` mirrors `JetStreamAuditSink` lifecycle (sink owns its NATS attach, removes the SLF001 `sink._js = js` poke); `HttpBlobStore.__init__` signature is honest (dead `connect_retry_max_s` param removed); HEAD handler 3→≤2 DB lookups; `.importlinter` peer-layer comment + 2 CLAUDE.md updates + stale analysis table superseded.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1362: V8.1 hardening + cleanup pass (parking-lot from PR #1358) | OPEN |
| Frame | [1362-v8-1-hardening-cleanup-frame.mdx](artifacts/frames/1362-v8-1-hardening-cleanup-frame.mdx) | Approved |
| Spec | [1362-v8-1-hardening-cleanup-spec.mdx](artifacts/specs/1362-v8-1-hardening-cleanup-spec.mdx) | Approved (8 grouped criteria) |
| Plan | [1362-v8-1-hardening-cleanup-plan.mdx](artifacts/plans/1362-v8-1-hardening-cleanup-plan.mdx) | Approved (16 micro-tasks, 6 agents, 3 waves) |
| Implementation | 1 commit on `feat/1362-v8-1-hardening-cleanup` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (2 new) Import contracts ✅ | Passed |

## Test Plan
- [ ] Inspect `deploy/quadlet/lyra-blobstore.container` and confirm the `ExecStartPre=` line is present and references `TAILSCALE_IPV4`.
- [ ] `uv run pytest tests/blobstore/test_quadlet_unit.py -v` — H1 snapshot test passes.
- [ ] `uv run pytest tests/blobstore/test_serve_api.py::TestHeadBlob::test_head_does_not_call_store_exists -v` — L1 SQL-trace assertion passes (the live `app.state.store.exists` is never reached during HEAD).
- [ ] `grep -rn 'connect_retry_max_s' packages/roxabi-blobs/` — 0 hits (M2 dead-knob removal).
- [ ] `grep -nE 'async def provision' src/lyra/blobstore/audit_sink.py` — 1 line (M1).
- [ ] `grep -c 'sink._js' src/lyra/blobstore/serve.py` — 0 (SLF001 poke gone).
- [ ] `uv run lint-imports` — 8 contracts kept, 0 broken (L6 peer-layer comment doesn't add `lyra.blobstore` to `shared-modules-upper-boundary`).
- [ ] Full blobstore suite: `uv run pytest tests/blobstore/ packages/roxabi-blobs/tests/` — 125 passed, no new failures.

## Reviewer notes
- M1 expanded beyond the original review item: also removes the `SLF001`-noqa private-field mutation in `_provision_nats` (caught by /spec architect review).
- H1 semantic: the `ExecStartPre=` guard fires *after* Podman has already constructed the publish-port spec — so the operational outcome is "container never serves traffic" rather than "bind decision averted" (clarified in `# RESIDUAL RISK` comment and in the spec).
- L4 `_conn_ro()` is a new method, no callers yet — added as the WAL-doc anchor since `_conn` is an attribute on `FsBlobStore`, not a method. Worth a sanity check on whether to keep as a clean read-only API or replace with an inline comment near `self._conn`.
- L2 fix landed inline during /implement: the lifespan-body default `app.state.audit_sink = BlobAuditSink()` was being overwritten on the success path; moved into the `else` branch so it's set only when `nc is None`.

Closes #1362

---

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -109,6 +109,9 @@ forbidden_modules =
     lyra.bootstrap.wiring
 allow_indirect_imports = true
 
+# NOTE: lyra.blobstore is intentionally absent from source_modules below — it is a
+# peer-layer service (own systemd unit + NATS attach, ADR-073), not a shared floating
+# module. See src/lyra/blobstore/CLAUDE.md "peer-of-adapters" placement.
 [importlinter:contract:shared-modules-upper-boundary]
 name = Shared floating modules must not import bootstrap, adapters, or infrastructure
 type = forbidden
--- a/artifacts/analyses/1330-v8-http-fronted-blobstore-analysis.mdx
+++ b/artifacts/analyses/1330-v8-http-fronted-blobstore-analysis.mdx
@@ -195,23 +195,7 @@ Step 1 must happen **before** step 2 (else a blob_refs row may point to a hash w
 
 Cloned from `deploy/quadlet/lyra-hub.container` with the following deltas:
 
-| Setting | Value | Rationale |
-|---------|-------|-----------|
-| `Description=` | `Lyra BlobStore HTTP service` | `[Unit]` stanza — `systemctl status` clarity |
-| `ContainerName=` | `lyra-blobstore` | Quadlet DNS handle for same-host clients (`http://lyra-blobstore:8449`) |
-| `Network=` | `roxabi.network` | shared Podman network with hub/adapters/NATS |
-| `Image=` | `ghcr.io/roxabi/lyra:staging-svc` | svc-runtime variant — requires `config.toml` bind-mount; same as hub/telegram/discord |
-| `Exec=` | `lyra blobstore serve` | typer subcommand |
-| `Environment=NATS_URL=` | `nats://lyra-nats:4222` | audit sink + readiness KV target — matches hub convention |
-| `PublishPort=` | bind to Tailscale IPv4 (resolved at provision time via `tailscale ip -4`) — `<ts-ipv4>:8449:8449` | LAN exposure is unintentional surface — bearer token is the only auth; restrict to Tailnet. `0.0.0.0:8449` is the fallback if `tailscale0` is absent, accepted as documented residual risk. |
-| `HealthCmd=` | `pgrep -f "lyra blobstore serve"` | mirrors hub pattern |
-| `CPUQuota=` | `100%` | cold-path I/O; doesn't need hub's `200%` |
-| `After=` / `Wants=` | `lyra-nats.service` (soft) | audit sink degrades to logger if NATS unreachable; service can start without NATS |
-| `Volume=` | `%h/.lyra/blobs:/data/lyra/blobs:z` + `%h/.lyra/config.toml:/app/config.toml:ro,z` | content dir bind + config bind |
-| `Secret=` | `lyra_blobstore_token,type=mount,target=lyra_blobstore_token,uid=1500,gid=1500,mode=0400` | ADR-054 |
-| `Label=` | `io.containers.autoupdate=registry` | same autoupdate cadence as svc-runtime peers |
-| `EnvironmentFile=` | `%h/.lyra/env/blobstore.env` | per Quadlet env convention |
-| Hardening | `NoNewPrivileges=true`, `ReadOnly=true`, `DropCapability=all`, `UserNS=keep-id:uid=1500,gid=1500` | clones hub |
+| _(superseded — see [docs/QUADLET-DEPLOYMENT.md](../../docs/QUADLET-DEPLOYMENT.md))_ |
 
 ### Cross-host integration test
 
--- a/artifacts/analyses/1362-v8-1-hardening-cleanup-consensus.mdx
+++ b/artifacts/analyses/1362-v8-1-hardening-cleanup-consensus.mdx
@@ -1,0 +1,83 @@
+---
+title: "PR #1364 residual findings — Expert Consensus"
+issue: 1362
+status: consensus-reached
+date: 2026-05-26
+panel: architect, devops, product-lead
+confidence: high
+---
+
+## Problem
+
+PR #1364 (V8.1 BlobStore hardening cleanup) received a first auto-applied commit
+draining B1 (Tailscale guard env scope), B2 (BlobAuditSink dead exception),
+and N1 (dead `_conn_ro`). 6 residual review findings remain (N2-N7). N5 was
+auto-resolved by the B2 fix. The panel evaluated whether each remaining finding
+warrants in-PR action, deferral to a follow-up issue, or skip.
+
+## Panel
+
+| Agent | Focus |
+|-------|-------|
+| architect | encapsulation, layer boundaries, Protocol v2 readiness |
+| devops | Quadlet correctness, audit forensic completeness, package release hygiene |
+| product-lead | scope discipline, parking-lot drain principle, F-lite tier shape |
+
+## Consensus ρ
+
+| Finding | Verdict | Vote | Solution |
+|---------|---------|------|----------|
+| N2 (double `nc.jetstream()`) | **SKIP** | 3-0 unanimous | — (idempotent in nats-py; both fix options trade encapsulation for zero behavioral gain) |
+| N3 (tautological snapshot test) | **APPLY** | 2-1 majority | Tighten regex to require `${VAR}` sigil + `-n` test |
+| N4 (HttpBlobStore sentinel `content_hash`) | **DEFER** | 2-1 majority | Follow-up issue tied to roxabi-blobs package release boundary + Protocol v2 ambiguity |
+| N5 (provision async w/ zero await) | **RESOLVED** | — | B2 fix already added `await self._js.account_info()` |
+| N6 (HEAD audit drops `content_hash`) | **APPLY** | 3-0 unanimous | Add `content_hash=str(row[0])` to `_emit_audit` call (already in scope) |
+| N7 (whitespace `TAILSCALE_IPV4` bypass) | **DEFER** | 3-0 unanimous | Follow-up — Podman's parse error provides accidental defense in depth; correct fix is a behavioral guard change |
+
+### Rationale
+
+The two unanimous applies (N3, N6) are both single-line changes with data
+already in scope. N3 locks the guard's `${VAR}` sigil shape so a future
+refactor of the H1 ExecStartPre cannot regress vacuously. N6 reinstates the
+content_hash field that the L1 lookup-collapse silently dropped from HEAD
+audit events — a forensic regression vs the pre-V8.1 path.
+
+The two deferrals (N4, N7) both cross genuine scope lines: N4 touches the
+`packages/roxabi-blobs/` release boundary and may be obviated by Protocol v2;
+N7 changes a security-adjacent guard behavior and warrants its own tracked
+PR rather than a rushed shell-pipeline change inside a 30-min budget.
+
+N2 was unanimously skipped: `nc.jetstream()` is documented idempotent in
+nats-py (the client caches the JetStream context), and the proposed fixes
+either widen the encapsulation surface of `BlobAuditSink` or introduce
+SLF001 noqa for a private attribute — both worse than the status quo.
+
+### Trade-offs
+
+- **In-PR scope**: 2 trivial applies (N3, N6) keep PR within F-lite shape; deferrals add ticket noise but preserve cleanup-tier discipline
+- **Parking-lot drain**: 4 of 6 findings get terminal action (apply or defer with traceable follow-up issue); N5 self-resolved; N2 skipped with rationale — no residual debt
+- **Architect dissent on N4**: silent type confusion in `BlobRef.content_hash` could compromise future dedup callers — recorded as known risk in the deferred issue body
+
+## Alternatives
+
+| ω | Proposed by | Rejected because |
+|---|-------------|------------------|
+| Apply N4 (content_hash="") in-PR | architect | crosses roxabi-blobs package release boundary; Protocol v2 may restructure exists() entirely |
+| Apply N7 (tr -d whitespace) in-PR | (none) | shell pipeline in ExecStartPre is brittle; Podman's downstream parse error provides defense in depth |
+| Defer N3 to follow-up | product-lead | test-tautology class is hard to revisit later because nobody hits it as a runtime bug; 1-line regex fix is well-defined |
+
+## Dissent
+
+- **N3**: product-lead preferred deferral (test-quality, not correctness) but conceded the regex tighten is 1-line and locks in the sigil shape that N7's eventual fix will rely on. 2-1 APPLY stands.
+- **N4**: architect argued that silent type confusion in `BlobRef.content_hash` (storing a store_key in a sha256-typed field) could corrupt future dedup logic; recommended in-PR fix `content_hash=""`. devops + product-lead both deferred citing package release hygiene + Protocol v2 ambiguity. 2-1 DEFER stands, but architect's risk note is preserved in the N4 follow-up issue.
+
+## Implementation Notes
+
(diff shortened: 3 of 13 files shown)


MickaelV0:
## Code Review

7 fresh agents (security-auditor, architect, product-lead, backend-dev, devops, tester, axial-adr-review). Cross-agent convergence on 2 blocking issues + several non-blocking improvements.

### 🔴 Blockers

**B1. H1 guard never fires — `TAILSCALE_IPV4` not in scope of `ExecStartPre=`** *(security-auditor C=92%, confirmed by repo inspection)*
- `deploy/quadlet/lyra-blobstore.container:62`
- Root cause: `[Container] EnvironmentFile=` translates to Podman's `--env-file` (container-scope only). The generated `[Service]` carries no systemd-level `EnvironmentFile=`, so `${TAILSCALE_IPV4}` is unset for `ExecStartPre=` which runs in systemd scope BEFORE the container starts. The guard's shell test `[ -n "" ]` then fails unconditionally → service is permanently broken post-deploy. Sibling units (`lyra-clipool.container`, `lyra-hub.container`) only use systemd specifiers like `%h` in `ExecStartPre=`, never `${VAR}` from the env file — this PR is the first to attempt it, which is why no one hit the trap before.
- Class: `[missing-input-validation, candidate/quadlet-env-scope-trap]`
- Fix: add `EnvironmentFile=%h/.lyra/env/blobstore.env` in `[Service]` (Quadlet passes `[Service]` through verbatim). Alternative: `ExecStartPre=/bin/sh -c '. %h/.lyra/env/blobstore.env && [ -n "${TAILSCALE_IPV4}" ] || exit 1'`.
- Note: devops agent disagreed (97%) but confused `[Service] EnvironmentFile=` vs `[Container] EnvironmentFile=` scoping; security-auditor (92%, citing Podman discussion #20642) and repo inspection both confirm the trap is real.

**B2. `BlobAuditSink.provision()` dead exception handler — `_degraded` never set** *(security-auditor C=85%, backend-dev C=88%)*
- `src/lyra/blobstore/audit_sink.py:42-49`
- Root cause: `nc.jetstream()` is a synchronous constructor (`return nats.js.JetStreamContext(self, **opts)`), it never raises `nats.errors.Error`. The `try/except nats.errors.Error` block is dead code — `self._degraded` is never set via `provision()`. Degradation is only detected on the first `emit()` call. This means `if sink._degraded: return` in `_provision_nats:59` is permanently False, so the KV-skip branch is unreachable; and one audit event may be lost before per-event degradation kicks in.
- Diverges from `JetStreamAuditSink.provision()` which catches *real* async errors from `js.add_stream(cfg)`.
- Class: `[bare-except, missing-error-handling]` (subsumed → tag `bare-except`)
- Fix options:
  1. Live probe inside `provision()` — e.g. `await self._js.account_info()` — wrap in `except nats.errors.Error` (real I/O can fail).
  2. Widen catch to `except Exception` (with `BLE001`) — accept that JetStream construction is currently infallible but want defensive guard.
  3. Remove the dead try/except + document "degradation detected on first `emit()`" explicitly in the docstring.

### 🟡 Non-blocking (recommended)

**N1. `_conn_ro()` is dead code** *(security-auditor C=72%, architect C=92%, backend-dev C=85%, axial-adr-review C=72%) — 4 agents independently flagged*
- `packages/roxabi-blobs/src/roxabi_blobs/fs_store.py:119`
- No callers. Spec L4 asked for a comment near `_conn`; agent added a new method instead (because `_conn` is an attribute, not a method). Risk: future cleanup removes "dead method" along with its WAL docstring, losing the rationale.
- Class: `[parallel-path-drift]` (candidate/dead-api-surface)
- Fix: delete `_conn_ro()`, move the WAL explanation to a comment on `self._conn = conn` in `__aenter__` OR onto `_require_open()`. OR wire `_conn_ro()` into 1+ existing read callers (e.g. `serve.py._blob_count` at line 40 uses `store._conn` directly with `SLF001` noqa).

**N2. Double `nc.jetstream()` call in `_provision_nats`** *(security-auditor C=78%, architect C=82%)*
- `src/lyra/blobstore/serve.py:57,62`
- `sink.provision(nc)` calls `nc.jetstream()` internally → `self._js = …`. Then `serve.py:62` calls `js = nc.jetstream()` again for the KV path. Idempotent in nats-py today but two independent contexts; bakes in the assumption silently.
- Class: `[parallel-path-drift]`
- Fix: expose JS handle from `provision()` (return it OR add `BlobAuditSink.js` property), then reuse: `js = sink._js`.

**N3. H1 snapshot test is tautological** *(tester C=88%)*
- `tests/blobstore/test_quadlet_unit.py:10-25`
- Asserts literal `ExecStartPre=` + `TAILSCALE_IPV4` strings — both planted by the PR. Inert stub like `ExecStartPre=/bin/sh -c 'exit 0' # TAILSCALE_IPV4` would also pass.
- Class: `[test-tautology]`
- Fix: tighten regex to require the variable-expansion sigil + the `-n` test: `assert re.search(r'\[\s*-n.*\$\{?TAILSCALE_IPV4\}?', line)` for each `ExecStartPre=` line.

**N4. `HttpBlobStore.exists()` sentinel `BlobRef.content_hash` carries the input arg** *(backend-dev C=78%)*
- `packages/roxabi-blobs/src/roxabi_blobs/http_store.py:145-152`
- The Protocol parameter is named `content_hash`, but per CLAUDE.md the HTTP variant treats it as a `store_key`. The synthesized sentinel sets `content_hash=content_hash` — so the field carries a store_key, not a sha256. Callers depending on `BlobRef.content_hash` for dedup/integrity would silently get a wrong-typed value.
- Fix: set `content_hash=""` (or `"unknown"`) in the sentinel; document in the method docstring that the field is unavailable via HEAD.

**N5. `BlobAuditSink.provision()` is `async def` with zero `await`** *(backend-dev C=72%)*
- Misleading signature: callers `await` a coroutine that does sync object construction. Either keep async (with comment "async for future stream-assertion step") OR make it sync (and update the one callsite in `serve.py`).

**N6. HEAD audit event drops the available `content_hash`** *(security-auditor C=68%)*
- `src/lyra/blobstore/_handlers.py:208`
- Forensic value reduction vs the old 3-lookup path. `_emit_audit(...)` could include `content_hash=str(row[0])`. Non-critical.

**N7. Whitespace-only `TAILSCALE_IPV4` passes the H1 guard** *(devops C=82%)*
- `[ -n " " ]` is true. Podman catches it downstream with a parse error, so fail-closed is preserved by accident. Tighten with a non-empty-after-trim test if you care about the comment's stated semantics.

**N8. L2 audit_sink assignment count (verification mismatch, not a bug)** *(product-lead C=92%)*
- Spec criterion reads "exactly 1 inside the lifespan body". Post-fix, `lifespan` has 1 (line 123 in `else` branch); `_provision_nats` (called by lifespan, defined separately) has 1 (line 58); `build_app` has 1 (line 158). The original review pattern (assigned-then-overwritten) is gone — no double-write on any path. The spec's grep is overly literal; runtime correctness is intact.

### 🟢 Praise

- **H1 guard string is injection-free** (security-auditor C=99%) — `[ -n "${TAILSCALE_IPV4}" ]` cannot be exploited via env-file values.
- **Bearer auth uses `hmac.compare_digest`** (security-auditor C=99%) — pre-existing but worth highlighting.
- **`.importlinter` peer-layer comment faithfully invokes ADR-073** (axial-adr-review C=97%) — placement is consistent with the layers contract.
- **CLAUDE.md HEAD-handler section introduces no axial drift** (axial-adr-review C=95%) — three-strikes safeguard correctly retained.
- **`ExecStartPre=` placement in `[Service]` matches sibling pattern** (devops C=97%) — `lyra-clipool.container` uses the same idiom.
- **`ExecStartPre=` abort semantics comment is accurate** (devops C=97%) — non-zero exit → unit in `failed` state, no `ExecStart=`.
- **Product spec coverage**: 10/12 acceptance criteria verified met from diff (M1, M2 met *in diff*; L2 with the grep-count caveat above; L4 with the dead-method caveat above). 12-item parking-lot fully drained, no silently dropped items.

Blocking: B1 (H1 guard is non-functional in prod), B2 (M1 degraded-mode dead code).

Both fixes are small (≤5 lines each). Recommend addressing B1 + B2 + N1 (dead method) in a fixup, leaving N4, N5, N6, N7 as either follow-up tickets or low-priority commits in the same PR.

— /code-review · 7 agents · single chunk · 399 diff lines

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents or cross-verified):** 3 finding(s) — commit `691c739b`
**Applied via /consensus + 1b1:** 2 finding(s) — commit `031618a6`
**Deferred (issues created):** 2 finding(s) — #1367, #1368
**Skipped:** 1 finding (N2) + praise items
**Auto-resolved:** 1 finding (N5)
**Failed:** 0
**Enforcement diagnostics:** 0

### Auto-Applied (commit `691c739b`)
- [applied] **B1** issue(blocking): H1 Tailscale guard `${TAILSCALE_IPV4}` was unset in systemd scope — added `[Service] EnvironmentFile=%h/.lyra/env/blobstore.env` so the env var resolves before the ExecStartPre guard fires. `deploy/quadlet/lyra-blobstore.container:55-62` (92%, cross-verified)
- [applied] **B2** issue(blocking): `BlobAuditSink.provision()` had a dead `try/except nats.errors.Error` — added `await self._js.account_info()` live probe so the catch is meaningful. `src/lyra/blobstore/audit_sink.py:42-49` (88%, 2 agents)
- [applied] **N1** suggestion: `_conn_ro()` was a dead method with no callers — deleted it and moved the WAL-read-isolation docstring onto `_require_open()`. `packages/roxabi-blobs/src/roxabi_blobs/fs_store.py:113-120` (92%, 4 agents)

### Applied (1b1 + /consensus — commit `031618a6`)
- [applied] **N3** suggestion(class:test-tautology): tightened `test_quadlet_unit.py` to assert the shell expression shape `[ -n "${TAILSCALE_IPV4}" ]` (not just string presence) + added a second test asserting `[Service] EnvironmentFile=` is declared so B1 cannot regress silently. `tests/blobstore/test_quadlet_unit.py` (88%, consensus 2-1)
- [applied] **N6** suggestion: HEAD audit event now includes `content_hash=str(row[0])` — restores the pre-collapse 3-lookup forensic record. `src/lyra/blobstore/_handlers.py:208-214` (68%, consensus 3-0)

### Deferred
- **N4** → **#1367** — `HttpBlobStore.exists()` sentinel `BlobRef.content_hash` silently carries a `store_key` (architect risk: silent type confusion; deferred for `roxabi-blobs` release boundary + Protocol v2 ambiguity)
- **N7** → **#1368** — Whitespace-only `TAILSCALE_IPV4` passes H1 guard (Podman's downstream parse error provides accidental defense in depth; deferred so behavioral guard change gets its own tracked PR)

### Skipped
- **N2** (double `nc.jetstream()` call): unanimous SKIP — `nc.jetstream()` is documented idempotent in nats-py; both proposed fixes trade encapsulation for zero behavioral gain.

### Auto-Resolved
- **N5** (`provision()` async with zero await): the B2 fix added `await self._js.account_info()` — provision() is no longer signature-dishonest.

### Consensus

3-expert panel (architect, devops, product-lead) was convened for the residual findings before 1b1 — full record in [`artifacts/analyses/1362-v8-1-hardening-cleanup-consensus.mdx`](artifacts/analyses/1362-v8-1-hardening-cleanup-consensus.mdx) (commit `f33c8452`). Unanimous on N2 (skip), N6 (apply), N7 (defer); 2-1 on N3 (apply) and N4 (defer).

— /fix · 5 applied · 2 deferred · 1 skipped · 1 auto-resolved

PR #1386, opened 2026-05-26
fix(deploy): #1368 blobstore Tailscale guard whitespace-safe

### LINKED ISSUE #1368: fix(deploy): blobstore Tailscale guard whitespace-safe
**Origin:** PR #1364 review [comment-4537497690](https://github.com/Roxabi/lyra/pull/1364#issuecomment-4537497690) (deferred per /fix walkthrough + /consensus 2026-05-26).

## Problem

`deploy/quadlet/lyra-blobstore.container:62` guard:

```ini
ExecStartPre=/bin/sh -c '[ -n "${TAILSCALE_IPV4}" ] || exit 1'
```

`[ -n " " ]` is **true** — whitespace-only \`TAILSCALE_IPV4\` (e.g. \`TAILSCALE_IPV4="   "\`) passes the guard. Podman then constructs \`PublishPort=   :8449:8449\` which is malformed and rejected at runtime — so fail-closed is preserved **by accident** via Podman's downstream parse error, not by the guard's intent.

## Why deferred (consensus 2026-05-26 — devops self-assessment)

- Podman's parse-rejection provides real defense in depth — the operational fail-closed property holds.
- The correct fix `[ -n "$(echo "${TAILSCALE_IPV4}" | tr -d '[:space:]')" ]` is a behavioral change to a security-adjacent guard — warrants its own tracked PR rather than a rushed shell pipeline inside a 30-min parking-lot budget.
- PR #1364 N3 (snapshot test) now locks in the \`\${VAR}\` sigil form so any future refactor must keep the variable reference visible.

## Action

1. Tighten guard to strip whitespace before `-n` test, e.g.:
   ```sh
   v="${TAILSCALE_IPV4}"; v="${v// /}"; [ -n "$v" ] || exit 1
   ```
   (pure POSIX sh, no subshell pipe, simpler than `tr -d`).
2. Update the snapshot test in `tests/blobstore/test_quadlet_unit.py` to reject whitespace-only values explicitly (synthetic guard test).
3. Document the whitespace-safe semantics in `deploy/CLAUDE.md §Known residual risk — blobstore PublishPort Tailscale fallback`.

## References

- `deploy/CLAUDE.md` §Known residual risk — blobstore PublishPort Tailscale fallback
- PR #1364 — H1 guard initial implementation
- PR #1364 commit 691c739b — Service-scope EnvironmentFile fix (B1)

MickaelV0:
Closes #1368

## Summary

- Replace `[ -n "${TAILSCALE_IPV4}" ]` with POSIX-sh `tr -d "[:space:]"` strip before the `-n` test, making the guard the authoritative rejection point for empty, unset, and whitespace-only values
- Prior behavior: `[ -n "   " ]` is TRUE in POSIX sh — whitespace-only values passed the guard and fail-closed relied on Podman's downstream parse error (accident, not intent)
- New behavior: systemd-layer rejection via `ExecStartPre=` before the container ever enters running state

## POSIX-sh compatibility

`${v// /}` (bash parameter substitution with global replace) is bash-specific and fails on dash (`/bin/sh` on some hosts). `printf %s … | tr -d "[:space:]"` is fully POSIX: `tr`, `printf`, `$(…)`, and `[ ]` are all specified by POSIX.1-2017.

## Test plan

- [x] `uv run pytest tests/blobstore/test_quadlet_unit.py -xvs` — 3 tests pass (snapshot guard shape + behavioral + env-file scope)
- [x] Test shape: **(b) behavioral** — `subprocess.run(["/bin/sh", "-c", <extracted script>])` with `TAILSCALE_IPV4="   "` asserts exit 1; with `TAILSCALE_IPV4=""` asserts exit 1; with `TAILSCALE_IPV4="100.64.1.2"` asserts exit 0
- [x] `uv run pytest -x --timeout=60 -q` — full suite green (pre-existing flaky NATS e2e excluded)
- [x] `uv run ruff check tests/blobstore/test_quadlet_unit.py` — no errors
- [x] `uv run pyright tests/blobstore/test_quadlet_unit.py` — 0 errors, 0 warnings
- [x] `grep "tr -d" deploy/quadlet/lyra-blobstore.container` — confirmed
- [x] `grep "whitespace" tests/blobstore/test_quadlet_unit.py` — confirmed

### DIFF
--- a/deploy/CLAUDE.md
+++ b/deploy/CLAUDE.md
@@ -102,6 +102,12 @@ is absent at container start, Podman falls back to `0.0.0.0:8449` (LAN-exposed).
 token (`lyra_blobstore_token`) is then the **sole** auth boundary. Accepted for V8; Phase 2
 (network policy / per-identity tokens) will address this systematically.
 
+The `ExecStartPre=` guard strips all whitespace before the `-n` test (POSIX `tr -d`) so
+empty, unset, **and whitespace-only** values are all rejected at the systemd layer (#1368).
+Prior to #1368, `[ -n "   " ]` was TRUE in POSIX sh — a whitespace-only value passed the
+guard and Podman's downstream parse error provided fail-closed behaviour by accident, not
+by design. The guard is now the authoritative rejection point.
+
 ### Known residual risk — clipool `core.hooksPath` override (tracked #1245)
 
 The clipool unit sets `core.hooksPath = /opt/lyra-gh/hooks` via `GIT_CONFIG_GLOBAL`
--- a/deploy/quadlet/lyra-blobstore.container
+++ b/deploy/quadlet/lyra-blobstore.container
@@ -59,13 +59,18 @@ HealthInterval=30s
 # through verbatim, so this loads the file twice — once for systemd (here), once for
 # Podman (above) — both intentional.
 EnvironmentFile=%h/.lyra/env/blobstore.env
-# Fail-closed guard: refuse to start when TAILSCALE_IPV4 is empty/unset.
+# Fail-closed guard: refuse to start when TAILSCALE_IPV4 is empty, unset,
+# or whitespace-only. `[ -n "   " ]` is TRUE in POSIX sh, so a bare -n test
+# passes on whitespace-only values — Podman would then construct a malformed
+# PublishPort spec and reject it downstream (fail-closed by accident, not by
+# intent). The tr-based strip removes all whitespace before the -n test so
+# the guard is the authoritative rejection point, not Podman's parser.
 # Note: the `PublishPort=${TAILSCALE_IPV4}:8449:8449` line above is resolved
 # by Podman BEFORE this ExecStartPre fires — so the publish-port spec is
 # already constructed (and could read as 0.0.0.0). This guard prevents the
 # CONTAINER from entering running state, so no traffic is ever served.
 # Operationally equivalent to closing the port (nothing accepts traffic).
-ExecStartPre=/bin/sh -c '[ -n "${TAILSCALE_IPV4}" ] || exit 1'
+ExecStartPre=/bin/sh -c 'v=$(printf %s "${TAILSCALE_IPV4}" | tr -d "[:space:]"); [ -n "$v" ] || exit 1'
 Restart=on-failure
 RestartSec=10
 # Cold-path I/O service — 1 full core is sufficient; prevents it from starving
--- a/tests/blobstore/test_quadlet_unit.py
+++ b/tests/blobstore/test_quadlet_unit.py
@@ -1,17 +1,24 @@
 """Snapshot test for deploy/quadlet/lyra-blobstore.container — H1 (#1362)."""
+
 from __future__ import annotations
 
 import pathlib
 import re
+import subprocess
 
 REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
 QUADLET_UNIT = REPO_ROOT / "deploy" / "quadlet" / "lyra-blobstore.container"
 
-# Match the shell expression shape `[ -n "${TAILSCALE_IPV4}" ]` (with optional
-# braces / whitespace around the bracket). An inert stub like
-# `ExecStartPre=/bin/sh -c 'exit 0' # TAILSCALE_IPV4` (string-only mention)
-# fails this pattern — the test is no longer tautological.
-_GUARD_PATTERN = re.compile(r'\[\s*-n\s+"\$\{?TAILSCALE_IPV4\}?"\s*\]')
+# Match the whitespace-safe guard shape: tr -d strips whitespace then -n tests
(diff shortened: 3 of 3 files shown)


MickaelV0:
## Code Review

**Verdict:** Approve with comments — 1 test-brittleness issue + coverage suggestions; core fix is correct.

> Note: 2 of the 7 review agents reported "fix is not present" — that was a false positive from reading the main repo (on `staging`, not yet merged) instead of the PR branch. Filtered out. The diff (`gh pr diff 1386`) confirms the new POSIX `tr -d` guard + behavioral test are correctly applied.

### Issues

**issue(brittleness):** Extraction regex greedy capture can silently over-match.

- `tests/blobstore/test_quadlet_unit.py` `_extract_exec_start_pre_shell`
- Pattern: `r"/bin/sh\s+-c\s+'(.+)'"` — greedy `(.+)` captures everything up to the LAST single-quote on the line.
- If a future `.container` change adds an inline comment with a `'`, or a second quoted segment, the regex captures the wrong string. The behavioral test runs a malformed script → may exit 0 (vacuous pass) or non-zero for the wrong reason. Failure mode is silent.
- Class: `test-tautology` (latent — currently functioning, will silently rot)
- **Solutions:**
  1. Non-greedy or character-class: `r"/bin/sh\s+-c\s+'([^']+)'"` — stops at first closing `'` (recommended).
  2. `shlex.split` on the full directive value.
- Confidence: high (tester)

### Suggestions

**suggestion(coverage):** Behavioral test misses tab, newline, and mixed-whitespace cases — all fixed by `tr -d "[:space:]"` but never asserted.

- `tests/blobstore/test_quadlet_unit.py::test_whitespace_only_tailscale_ipv4_rejected`
- Covered: `"   "` (3 spaces), `""`, valid IP.
- Missing: `"\t"` (tab-only), `"\n"` (newline), `" \t  "` (mixed). All exit 1 in practice (verified by tester agent), but evidence-from-test is weaker than evidence-from-behavior.
- **Solution:** Add 3 more `subprocess.run` cases mirroring the existing block. Drop-in, ~12 lines.
- Confidence: high (tester)

**suggestion(non-blocking):** `subprocess.run(env={"TAILSCALE_IPV4": ..., "PATH": ...})` replaces (not augments) parent env. Current script uses only sh builtins + `tr`/`printf` resolved via `PATH`, so harmless. If the guard is ever extended to call an external utility not on `/bin:/usr/bin`, the test silently strips its context.

- **Solution:** `env=os.environ | {"TAILSCALE_IPV4": ...}` (Python 3.9+) to inherit then override.
- Confidence: 72% (security)

### Nitpicks

**nitpick:** Unset `TAILSCALE_IPV4` (key absent from env) is not a distinct test case. POSIX expands unset to empty so the guard catches it transitively, but explicit coverage documents the contract and protects against a future refactor to `${TAILSCALE_IPV4:?}`. Verified: unset currently exits 1.

**nitpick:** `_STRIPPED_TEST_PATTERN = re.compile(r'\[\s*-n\s+"\\?\$v"\s*\]')` — the `\\?` makes the backslash optional. Intent: match both `"$v"` and `"\$v"` (systemd may escape `$` in some serializations). A 1-line comment explaining this would prevent a future "cleanup" from removing the backslash-optional. Confidence: tester.

### Summary

- 1 issue (test-brittleness, low priority)
- 2 suggestions (coverage gap, env isolation)
- 2 nitpicks

Core fix (`tr -d "[:space:]"` POSIX guard) is correct and well-documented. Tests close the original tautology gap (#1364 N3) and add behavioral verification of the rejection path. Acceptable to merge as-is and file follow-up for test hardening; or fold the regex fix + tab/newline cases in this PR (small batch).

🤖 Generated with /code-review

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 5 finding(s)
**Skipped:** 0 finding(s)
**Failed:** 0 finding(s)

### Applied (1b1)

- [applied] issue(brittleness): `_extract_exec_start_pre_shell` — greedy `(.+)` → non-greedy `[^']+` in `tests/blobstore/test_quadlet_unit.py:62`
- [applied] suggestion(coverage): `test_whitespace_only_tailscale_ipv4_rejected` — added tab-only, newline-only, mixed-whitespace cases in `tests/blobstore/test_quadlet_unit.py`
- [applied] suggestion(non-blocking): `subprocess.run` env — `os.environ | {...}` pattern applied to all 6 calls; dropped explicit `PATH` entries; added `import os` in `tests/blobstore/test_quadlet_unit.py`
- [applied] nitpick: unset `TAILSCALE_IPV4` case — explicit test added (parent env minus the key) in `tests/blobstore/test_quadlet_unit.py`
- [applied] nitpick: `_STRIPPED_TEST_PATTERN` — 2-line comment explaining `\\?` backslash-optional added above the definition in `tests/blobstore/test_quadlet_unit.py`

Commit: c64bc045 — all quality gates passed (38 blobstore + 4373 full suite).

PR #1400, opened 2026-05-26
fix(nats): #1391 auth.conf placeholder sentinel + install.sh guard

### LINKED ISSUE #1391: ops(nats): auth.conf placeholder sentinel + install.sh guard against accidental mount
**Origin:** PR #1389 review consensus (B5, deferred per `/consensus`).

The committed `deploy/nats/auth.conf` uses placeholder pubkeys (prefix `UDET*`) intended for the in-repo template. Real nkeys live on M₁ at `~/.lyra/nkeys/*.seed` and are rendered into the live `auth.conf` by `make nats-regen-authconf`. Nothing in CI or provision scripts asserts the placeholder file isn't mounted as-is on a fresh provision.

**Action:**
1. Add `# PLACEHOLDER — regenerate with make nats-regen-authconf before use` comment on each `nkey:` placeholder line.
2. Add guard in `deploy/install.sh` (or a provision step): `! grep -q '^.*nkey:.*UDET' deploy/nats/auth.conf` before mount.

**Class:** secret-leak, missing-input-validation

MickaelV0:
## Summary

- Closes #1391
- `deploy/nats/auth.conf`: added `# PLACEHOLDER — regenerate with: make nats-regen-authconf` comment before each of the 11 `UDET*` placeholder nkey lines, so the template's provisional state is visible in-file
- `deploy/install.sh`: added an early guard (section 0, before nkeys-dir check) that aborts with a clear error message if `auth.conf` still contains any `UDET*` placeholder, preventing accidental provision from the unrendered template

## Test plan

- [ ] `./deploy/install.sh` on a fresh clone (before `make nats-regen-authconf`) exits 1 with the expected ERROR message
- [ ] After `make nats-regen-authconf` (real nkeys rendered), `./deploy/install.sh` proceeds past the guard
- [ ] `--dry-run` is not short-circuited by the guard (grep runs on the source file, not a dry-run artifact)

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -45,19 +45,21 @@ run() {
 log() { echo "==> $*"; }
 warn() { echo "WARN: $*" >&2; }
 
-# ── 0. Refuse to install with placeholder nkeys still in auth.conf ───────────
-
-if grep -qE '^[[:space:]]*nkey:[[:space:]]+UDET' "${SCRIPT_DIR}/nats/auth.conf"; then
-  echo "ERROR: deploy/nats/auth.conf still contains UDET* placeholder pubkeys." >&2
-  echo "       Run: make nats-regen-authconf  (renders nkeys from ~/.lyra/nkeys/*.seed)" >&2
-  exit 1
-fi
-
 # ── 1. Verify nkeys dir ──────────────────────────────────────────────────────
 
 log "Checking ~/.lyra/nkeys/ ..."
 if [[ ! -d "${NKEYS_DIR}" ]]; then
   echo "ERROR: ${NKEYS_DIR} not found. Run: make nats-setup" >&2
+  exit 1
+fi
+
+# Refuse to install with placeholder nkeys still in the live auth.conf.
+# Skipped under --dry-run so the operator can still preview install actions.
+if [[ "$DRY_RUN" -eq 0 ]] \
+  && [[ -f "${NKEYS_DIR}/auth.conf" ]] \
+  && grep -qE '^[[:space:]]*nkey:[[:space:]]+"UDET' "${NKEYS_DIR}/auth.conf"; then
+  echo "ERROR: ${NKEYS_DIR}/auth.conf still contains UDET* placeholder pubkeys." >&2
+  echo "       Run: make nats-regen-authconf  (renders nkeys from ${NKEYS_DIR}/*.seed)" >&2
   exit 1
 fi
 
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -45,6 +45,14 @@ run() {
 log() { echo "==> $*"; }
 warn() { echo "WARN: $*" >&2; }
 
+# ── 0. Refuse to install with placeholder nkeys still in auth.conf ───────────
+
+if grep -qE '^[[:space:]]*nkey:[[:space:]]+UDET' "${SCRIPT_DIR}/nats/auth.conf"; then
+  echo "ERROR: deploy/nats/auth.conf still contains UDET* placeholder pubkeys." >&2
+  echo "       Run: make nats-regen-authconf  (renders nkeys from ~/.lyra/nkeys/*.seed)" >&2
+  exit 1
+fi
+
 # ── 1. Verify nkeys dir ──────────────────────────────────────────────────────
 
 log "Checking ~/.lyra/nkeys/ ..."
--- a/deploy/nats/auth.conf
+++ b/deploy/nats/auth.conf
@@ -10,6 +10,7 @@ authorization {
   }
   users: [
     {
+      # PLACEHOLDER — regenerate with: make nats-regen-authconf
       nkey: "UDETHUB"
       # hub
       permissions: {
@@ -19,6 +20,7 @@ authorization {
       }
     }
     {
+      # PLACEHOLDER — regenerate with: make nats-regen-authconf
       nkey: "UDETTELEGRAMADAPTER"
       # telegram-adapter
       permissions: {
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Code Review

**Verdict:** Request changes — **the guard is dead code**. Two independent critical bugs make it non-functional today, and a third gap breaks `--dry-run`. Both reviewers (security-auditor + devops) flagged the same two blockers at 100% confidence with reproducer commands.

---

### Blockers

**issue(blocking):** Guard regex never matches — placeholder check is permanently bypassed
- `deploy/install.sh:50`
- Root cause: actual nkey lines are `      nkey: "UDETHUB"` — the character between the space and `UDET` is `"`. The pattern `[[:space:]]+UDET` requires whitespace immediately before `UDET`, but `"` is not whitespace, so the class never spans `: "`. Reproduce: `printf '      nkey: "UDETHUB"\n' | grep -qE '^[[:space:]]*nkey:[[:space:]]+UDET'; echo $?` → `1`. The guard exits 0 (not-found → no-error) on every run; the entire block is dead code. Confirmed empirically against the worktree — `bash deploy/install.sh --dry-run` proceeds past the guard.
- Class: `missing-input-validation`
- Raw callsites: `[deploy/install.sh:50]`
- Solutions:
  1. **(recommended)** Insert the literal `"` into the pattern: `'^[[:space:]]*nkey:[[:space:]]+"UDET'`
  2. Broaden to `'nkey:.*UDET'` — tolerant to any quoting style.
- Confidence: 100%

**issue(blocking):** Guard checks the wrong file — the in-repo template is always a placeholder by design
- `deploy/install.sh:50`
- Root cause: the guard targets `${SCRIPT_DIR}/nats/auth.conf` (in-repo template, always placeholder). The Podman secret `lyra-nats-auth` is sourced from `${NKEYS_DIR}/auth.conf` = `~/.lyra/nkeys/auth.conf` (line 65 of the same script). `make nats-regen-authconf` writes to the latter, not the former — so the repo file is never updated and never validates anything meaningful. Even with the regex fixed, this guard either (a) fires permanently on every clone — blocking the install entirely — or (b) silently passes the actual mount source. Both branches defeat the issue's intent ("placeholder file isn't mounted as-is on a fresh provision").
- Class: `missing-input-validation`
- Raw callsites: `[deploy/install.sh:50, deploy/install.sh:65]`
- Solutions:
  1. **(recommended)** Move the guard after line 59 (post `! -d ${NKEYS_DIR}` check), target the actual secret source:
     ```bash
     if grep -qE '^[[:space:]]*nkey:[[:space:]]+"UDET' "${NKEYS_DIR}/auth.conf"; then
       echo "ERROR: ~/.lyra/nkeys/auth.conf still contains UDET* placeholder pubkeys." >&2
       echo "       Run: make nats-regen-authconf" >&2
       exit 1
     fi
     ```
  2. Keep both checks (repo template + live path), demoting the repo check to `WARN:` — "documents the intent that the template stays placeholder-only". Live-path check remains the blocker.
- Confidence: 99%

---

### Warning

**suggestion(blocking):** Guard runs unconditionally before `--dry-run` / `--secrets-only` mode-gating
- `deploy/install.sh:50` (vs flag parsing at `:28-35`)
- Root cause: under `--dry-run` the contract is "show what would happen without side-effects or hard stops". Exiting 1 before printing a single dry-run line breaks that contract. `--secrets-only` has the same issue.
- Solutions:
  1. **(recommended)** Wrap in `[[ "$DRY_RUN" -eq 0 ]]`, OR emit `WARN:` and continue under dry-run — mirrors the `run()` pattern that already gates all side-effecting ops on `$DRY_RUN`.
  2. Document explicitly that `--dry-run` does not bypass the guard (intentional safety property), and emit a dry-run-aware message before exiting.
- Confidence: 95%

---

### Thought

**thought:** After fixing the file-path, the error message phrasing will need a small tweak
- `deploy/install.sh:51-52`
- Root cause: current message blames "deploy/nats/auth.conf still contains UDET*" — but once the guard checks `~/.lyra/nkeys/auth.conf`, the more accurate framing is "nkeys not rendered yet" rather than "placeholder in repo template". Minor; the remediation command (`make nats-regen-authconf`) is correct either way.
- Confidence: 90%

---

### Praise

**praise:** PLACEHOLDER comment placement in `auth.conf` is well-executed
- `deploy/nats/auth.conf:13,23,33,…` (11 entries, all consistent)
- Each placeholder line is preceded by `# PLACEHOLDER — regenerate with: make nats-regen-authconf` — comment text exactly matches the install.sh error message, co-located at the point of confusion, no entry missed. (C=100%)

---

🤖 Reviewed by `/code-review` (security-auditor + devops, fresh context). Both agents converged independently on the same two blockers; the regex bug is empirically reproducible.

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 3 finding(s)
**Skipped:** 0 finding(s)
**Failed:** 0 finding(s)

### Applied (1b1)

- [applied] issue(blocker): regex `[[:space:]]+UDET` fails to match quoted value `"UDETHUB"` — fixed to `[[:space:]]+"UDET` (includes literal `"`)
- [applied] issue(blocker): guard checked `${SCRIPT_DIR}/nats/auth.conf` (in-repo placeholder, always placeholder by design) — moved to `${NKEYS_DIR}/auth.conf` (the live mount source)
- [applied] suggestion: guard fired under `--dry-run`, exiting before any preview output — gated entire block on `[[ "$DRY_RUN" -eq 0 ]]`

The dead-code Section 0 block is removed; the corrected guard lives inside Section 1 after the `NKEYS_DIR` directory check, with a `-f` test so a missing file falls through to the existing MISSING-seeds loop.

Commit: `4e51b53b`

PR #1389, opened 2026-05-26
feat(typing): #1376 T1 — lyra.typing.* plane scaffold

MickaelV0:
## Summary

- Foundational machinery for the **3rd NATS plane** (typing-indicator): `WorkScope` + `TypingEvent` Pydantic wire + ref-counted `TypingPublisher` (env-gated, swallow-on-error) + `TypingListener` (stage-axis, subscribes `lyra.typing.<platform>.<bot_id>`).
- New `src/lyra/typing/` stage-axis module between `lyra.core` and `lyra.streaming` (importlinter layer added). Per-platform module-level `_<platform>_scope_resolver` functions in adapters (AC8 — import-only testable). Bootstrap wires `TypingPublisher` on hub and `TypingListener` per platform/bot; `LYRA_TYPING_ENABLED=false` default = zero NATS traffic on `lyra.typing.*`.
- ACL plane regenerated: `hub` publishes `lyra.typing.>`, `{telegram,discord}-adapter` subscribes `lyra.typing.{telegram,discord}.>` (matrix-driven `auth.conf` + spec mdx).
- **No call-site migration** — T1 ships machinery only. T2 (separate issue) will replace the 8 imperative `_start_typing`/`_cancel_typing` sites and close #1374 by construction.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1376 — feat(typing): T1 (parent epic #1375) | In Progress |
| Frame | [1376-typing-plane-t1-frame.mdx](artifacts/frames/1376-typing-plane-t1-frame.mdx) | Approved |
| Spec | [1376-typing-plane-t1-spec.mdx](artifacts/specs/1376-typing-plane-t1-spec.mdx) | Approved |
| Plan | [1376-typing-plane-t1-plan.mdx](artifacts/plans/1376-typing-plane-t1-plan.mdx) | Approved |
| Implementation | 22 plan-tasks across 5 waves, 7 agent instances | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (13 new) Layer-contracts 8/8 kept | Passed |

## Acceptance Criteria

- [x] **AC1** Plane reachable end-to-end (publisher → NATS Core → listener → manager.start/cancel) — `tests/typing/test_integration_nats.py`
- [x] **AC2** Ref-counted publisher (2× started → 1 wire emit; LIFO ended balances) — `tests/transport/test_typing_publisher.py`
[...]

Issue #1376, opened 2026-05-26
feat(typing): T1 — WorkScope + TypingPublisher + TypingListener + lyra.typing.* plane


MickaelV0:
**Parent epic:** #1375

## Scope

Land the foundational machinery for the typing-lifecycle NATS plane. **No migration of existing call sites in this issue** — T2 (#TBD) does that.

### Code

- `src/lyra/transport/work_scope.py` — internal dataclass `WorkScope(platform, bot_id, scope_id, trace_id)`. **NOT in `roxabi-contracts`** — Pool-internal for T1/T2, promotion deferred to worker integration (#1044 follow-up).
- `src/lyra/transport/typing_publisher.py` — twin of `turn_publisher.py`:
  - `publish_started(scope) / publish_ended(scope)` API
  - **Ref-counted state** `dict[scope_key, int]` — emit `started` on 0→1, `ended` on 1→0. Sub-scope opt-out gratis (no `initiator` field).
  - PoolObserver-style failure mode: log + swallow on NATS publish error.
  - `LYRA_TYPING_ENABLED` env flag — default `false`, no-op when disabled.
- `src/lyra/typing/` — new **stage-axis** module (mirror `src/lyra/inbound/`):
  - `__init__.py`
  - `types.py` — `ScopeResolver` Protocol (`resolve(scope: WorkScope) -> int`), `TypingEvent` types
  - `listener.py` — shared `TypingListener`: subscribe to `lyra.typing.<platform>.<bot_id>` NATS Core, decode, dispatch to existing `TypingTaskManager`. `try/except` + `cancel(scope)` in `finally` on the subscription handler.
- Per-platform `ScopeResolver` impls **inline in `adapter.py`** (≤10 LOC each, no new `<platform>_typing.py` files):
  - Discord: returns `channel.id` (parent), never `thread.id`
  - Telegram: returns `chat_id`

### Wiring / config

- `.importlinter` — add `lyra.typing` to the clean-architecture layers contract between `streaming` and `core`. Same PR (avoid drift with #1287/#1289).
- Bootstrap registration of `TypingPublisher` (hub-side) + `TypingListener` (adapter-side) — follow current DI pattern; if #1283 lands first, align with post-#1283 DI cleanup.

### NATS ACL

- Update `artifacts/specs/706-per-role-nkeys-acls-spec.mdx`:
  - `hub` publish.allow += `lyra.typing.>`
[...]

PR #1404, opened 2026-05-26
feat(typing): WorkScope.trace_id non-empty __post_init__ guard

### LINKED ISSUE #1392: feat(typing): WorkScope.trace_id non-empty __post_init__ guard (T2)
**Origin:** PR #1389 review consensus (S2, deferred per `/consensus`).

Spec requires `WorkScope.trace_id` to be non-empty (threaded from inbound msg.id when available, else uuid4().hex). Currently `WorkScope` is a frozen dataclass with bare `trace_id: str` — no validation. Empty trace_id would silently break downstream tracing.

**Trigger:** land alongside T2 call-site wiring (when `Pool.process_one()` constructs WorkScope from inbound messages — that's the boundary where the invariant becomes operationally load-bearing).

**Action:** add `__post_init__` to `src/lyra/transport/work_scope.py`:
```python
def __post_init__(self) -> None:
    if not self.trace_id:
        raise ValueError("WorkScope.trace_id must be non-empty")
```

**Class:** missing-input-validation

MickaelV0:
## Summary
- Add `__post_init__` to `WorkScope` raising `ValueError` when `trace_id` is empty (#1392 T2).
- Lands before T2 call-site wiring (`Pool.process_one()`) so the invariant is operationally load-bearing the moment it gets threaded.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | [#1392](https://github.com/Roxabi/lyra/issues/1392): WorkScope.trace_id non-empty __post_init__ guard (T2) | Open |
| Implementation | 1 commit on `worktree-1392-workscope-trace-id-guard` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (2 new) | Passed |

S-tier — no analysis/spec artifacts (origin: PR #1389 review consensus, S2 deferred).

## Test Plan
- [x] `WorkScope(trace_id="abc")` constructs successfully (happy path).
- [x] `WorkScope(trace_id="")` raises `ValueError: WorkScope.trace_id must be non-empty`.
- [x] All existing `WorkScope(...)` call-sites already pass non-empty `trace_id` (verified via grep — no regressions).
- [ ] CI green on `worktree-1392-workscope-trace-id-guard`.

## Notes
- One pre-existing flake observed: `tests/typing/test_integration_nats.py::test_publisher_to_listener_e2e_discord` fails under loadgroup xdist but passes in isolation. Not introduced by this change (commit history shows the test landed in #1376 T1 d673c993). Out of scope for this guard.

Closes #1392

---

### DIFF
--- a/src/lyra/transport/work_scope.py
+++ b/src/lyra/transport/work_scope.py
@@ -9,5 +9,5 @@ class WorkScope:
     trace_id: str
 
     def __post_init__(self) -> None:
-        if not self.trace_id:
+        if not self.trace_id.strip():
             raise ValueError("WorkScope.trace_id must be non-empty")
--- a/src/lyra/transport/work_scope.py
+++ b/src/lyra/transport/work_scope.py
@@ -7,3 +7,7 @@ class WorkScope:
     bot_id: str
     scope_id: int
     trace_id: str
+
+    def __post_init__(self) -> None:
+        if not self.trace_id:
+            raise ValueError("WorkScope.trace_id must be non-empty")
--- a/tests/transport/test_work_scope.py
+++ b/tests/transport/test_work_scope.py
@@ -8,13 +8,12 @@ from lyra.transport.work_scope import WorkScope
 
 
 def test_workscope_constructs_with_non_empty_trace_id() -> None:
-    scope = WorkScope(
+    WorkScope(
         platform="telegram",
         bot_id="bot-1",
         scope_id=42,
         trace_id="abc123",
     )
-    assert scope.trace_id == "abc123"
 
 
 def test_workscope_rejects_empty_trace_id() -> None:
@@ -24,4 +23,14 @@ def test_workscope_rejects_empty_trace_id() -> None:
             bot_id="bot-1",
             scope_id=42,
             trace_id="",
+        )
+
+
+def test_workscope_rejects_whitespace_only_trace_id() -> None:
+    with pytest.raises(ValueError, match="trace_id must be non-empty"):
+        WorkScope(
+            platform="telegram",
+            bot_id="bot-1",
+            scope_id=42,
+            trace_id="   ",
         )
--- a/tests/transport/test_work_scope.py
+++ b/tests/transport/test_work_scope.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import dataclasses
+
 import pytest
 
 from lyra.transport.work_scope import WorkScope
@@ -34,3 +36,14 @@ def test_workscope_rejects_whitespace_only_trace_id() -> None:
             scope_id=42,
             trace_id="   ",
         )
+
+
+def test_workscope_replace_rejects_empty_trace_id() -> None:
+    scope = WorkScope(
+        platform="telegram",
+        bot_id="bot-1",
+        scope_id=42,
+        trace_id="abc123",
+    )
+    with pytest.raises(ValueError, match="trace_id must be non-empty"):
+        dataclasses.replace(scope, trace_id="")
--- a/tests/transport/test_work_scope.py
+++ b/tests/transport/test_work_scope.py
@@ -1,0 +1,27 @@
+"""Unit tests for lyra.transport.work_scope.WorkScope (#1392)."""
+
+from __future__ import annotations
+
+import pytest
+
+from lyra.transport.work_scope import WorkScope
+
+
+def test_workscope_constructs_with_non_empty_trace_id() -> None:
+    scope = WorkScope(
+        platform="telegram",
+        bot_id="bot-1",
+        scope_id=42,
+        trace_id="abc123",
+    )
+    assert scope.trace_id == "abc123"
+
+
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Code Review

**Verdict:** Approve with comments — 0 blockers, 3 suggestions, 1 praise.

The change is correct, minimal, and matches the spec from PR #1389 review consensus (S2). 3 fresh agents (security-auditor, tester, backend-dev) reviewed in parallel; all three independently flagged the same whitespace bypass below — strong cross-agent signal.

---

### Suggestions

**suggestion: `not self.trace_id` accepts whitespace-only / control-char strings** _(consensus: security-auditor, tester, backend-dev — C=97%)_
- `src/lyra/transport/work_scope.py:12`
- **Root cause:** Python's `not "   "` is `False`. A `trace_id` of three spaces or `"\x00"` bypasses the guard but is semantically empty for the downstream NATS bus consumers (`TurnPublisher` → `lyra.turns.write`, `TypingPublisher`). The pattern already used in `pool_observer.py:123` (`(message_id or "").strip() or uuid.uuid4().hex`) confirms whitespace is treated as empty elsewhere; the boundary should be uniform.
- **Class:** `missing-input-validation`
- **Solutions:**
  1. (recommended) `if not self.trace_id.strip(): raise ValueError("WorkScope.trace_id must be non-empty")` — aligns guard with semantic intent. Zero call-site impact (all current sites pass `uuid4().hex` or real message IDs).
  2. Leave as-is + document the gap in a guard comment (acceptable only if whitespace is considered impossible by convention).

**suggestion: missing test for whitespace-only `trace_id`** _(security-auditor + tester, C=97%)_
- `tests/transport/test_work_scope.py`
- **Root cause:** Test suite covers empty string + happy path but not `"   "` or control chars. Without this case, the guard's true contract boundary isn't pinned by tests.
- **Class:** `missing-input-validation`
- **Solutions:**
  1. (recommended) Add `test_workscope_rejects_whitespace_only_trace_id` with `trace_id="   "` once the production guard is tightened to `.strip()`.

**nitpick: happy-path assert is tautological** _(tester, C=85%)_
- `tests/transport/test_work_scope.py:17` — `assert scope.trace_id == "abc123"`
- **Root cause:** For `@dataclass(frozen=True)`, the assignment round-trips trivially regardless of `__post_init__`. The test would still pass with the guard deleted. The meaningful contract of the happy-path test is "construction does not raise" — the call alone proves that.
- **Class:** `test-tautology`
- **Solutions:**
  1. (recommended) Drop the `assert` line; keep the construction line as the implicit "does not raise" assertion.
  2. Replace with a structural equality assertion across all fields.

**nitpick: no test for `dataclasses.replace` path** _(backend-dev, C=75%)_
- `tests/transport/test_work_scope.py`
- **Root cause:** `dataclasses.replace(scope, trace_id="")` re-invokes `__post_init__` and should also raise — but this isn't pinned. Low-signal but the cheapest "invariant is visible" test.
- **Solutions:**
  1. Add `test_workscope_replace_rejects_empty_trace_id` asserting `dataclasses.replace(scope, trace_id="")` raises.

---

### Praise

**praise: `frozen=True, slots=True` + `__post_init__` is the correct pattern here.** No pitfalls — `__post_init__` on a frozen+slots class can safely read fields (only writes would need `object.__setattr__`). The guard runs on every construction path, including `dataclasses.replace`, providing transport-plane invariant enforcement without framework coupling. — backend-dev

---

**Summary:** the central suggestion is the whitespace tighten — three agents converged on it independently. Recommend applying solution 1 of the first finding (one-character change: `not self.trace_id` → `not self.trace_id.strip()`) plus the matching whitespace test. The other two test nitpicks are optional polish.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/code-review` (3 parallel agents)

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 3 finding(s)
**Skipped:** 1 (praise)
**Failed:** 0

### Auto-Applied
- [applied] suggestion: whitespace bypass in `src/lyra/transport/work_scope.py:12` (97%) — tightened guard to `not self.trace_id.strip()` (b7fbbc98)
- [applied] suggestion: missing whitespace test in `tests/transport/test_work_scope.py` (97%) — added `test_workscope_rejects_whitespace_only_trace_id` (b7fbbc98)
- [applied] nitpick: tautological happy-path assert in `tests/transport/test_work_scope.py:17` (85% → 99% after verifier) — dropped the redundant `assert scope.trace_id == "abc123"` line (b7fbbc98)

### Applied (1b1)
- [applied] nitpick: missing `dataclasses.replace` test (75%) — added `test_workscope_replace_rejects_empty_trace_id` to pin the invariant on the `replace` path (dec7deb1)

### Skipped
- praise: `frozen=True, slots=True` + `__post_init__` is the correct pattern — no action needed.

---

Label `reviewed` applied. CI pending on the 2 new commits.

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/fix`

PR #1405, opened 2026-05-26
refactor(typing): promote CoroFactory/FactoryBuilder to types.py

### LINKED ISSUE #1394: refactor(typing): promote CoroFactory/FactoryBuilder to lyra.typing.types
**Origin:** PR #1389 review consensus (S5, deferred per `/consensus`).

`CoroFactory = Callable[[], Coroutine[Any, Any, None]]` and `FactoryBuilder = Callable[[int], CoroFactory]` are currently local aliases in `src/lyra/typing/listener.py` (lines 18-19). T2 callers (Pool, adapter wrappers) will need to annotate against these types — they'll either re-declare or import from listener.py (a deep import).

**Trigger:** land when T2 actually imports these types from outside `lyra.typing.listener`.

**Action:** move both type aliases to `src/lyra/typing/types.py` next to the existing `ScopeResolver` + `TypingManagerProtocol`. Re-export from `src/lyra/typing/__init__.py`. Update `lyra.typing.listener` to import from `types.py`.

MickaelV0:
## Summary
- Move `CoroFactory` and `FactoryBuilder` type aliases from `src/lyra/typing/listener.py` to `src/lyra/typing/types.py` alongside the existing `ScopeResolver` + `TypingManagerProtocol`.
- Re-export both aliases from `lyra.typing` so T2 callers (Pool, adapter wrappers) can annotate against them without a deep import into the listener module.
- Pure relocation — no behavior change.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1394: refactor(typing): promote CoroFactory/FactoryBuilder to lyra.typing.types | Open |
| Implementation | 1 commit on `worktree-1394-promote-corofactory-factorybuilder-to-lyra-typing-types` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (existing — refactor adds no new behavior) | Passed |

## Test Plan
- [ ] `uv run python -c "from lyra.typing import CoroFactory, FactoryBuilder; print(CoroFactory, FactoryBuilder)"` resolves both names.
- [ ] `tests/typing/` continues to pass (`uv run pytest tests/typing/ -n0`).
- [ ] `lyra.typing.listener` import path still re-exports `FactoryBuilder` transitively (back-compat — though no callers rely on it yet).

Closes #1394

---

### DIFF
--- a/src/lyra/typing/__init__.py
+++ b/src/lyra/typing/__init__.py
@@ -1,4 +1,15 @@
 from lyra.typing.listener import TypingListener
-from lyra.typing.types import ScopeResolver, TypingManagerProtocol
+from lyra.typing.types import (
+    CoroFactory,
+    FactoryBuilder,
+    ScopeResolver,
+    TypingManagerProtocol,
+)
 
-__all__ = ["ScopeResolver", "TypingListener", "TypingManagerProtocol"]
+__all__ = [
+    "CoroFactory",
+    "FactoryBuilder",
+    "ScopeResolver",
+    "TypingListener",
+    "TypingManagerProtocol",
+]
--- a/src/lyra/typing/listener.py
+++ b/src/lyra/typing/listener.py
@@ -2,11 +2,14 @@ from __future__ import annotations
 
 import logging
 import os
-from collections.abc import Callable, Coroutine
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 from lyra.transport.typing_event import TypingEvent
-from lyra.typing.types import ScopeResolver, TypingManagerProtocol
+from lyra.typing.types import (
+    FactoryBuilder,
+    ScopeResolver,
+    TypingManagerProtocol,
+)
 
 if TYPE_CHECKING:
     from nats.aio.client import Client as NATS
@@ -14,9 +17,6 @@ if TYPE_CHECKING:
     from nats.aio.subscription import Subscription
 
 log = logging.getLogger(__name__)
-
-CoroFactory = Callable[[], Coroutine[Any, Any, None]]
-FactoryBuilder = Callable[[int], CoroFactory]
 
 
 class TypingListener:
--- a/src/lyra/typing/types.py
+++ b/src/lyra/typing/types.py
@@ -3,6 +3,9 @@ from typing import Any, Protocol, runtime_checkable
 
 from lyra.transport.work_scope import WorkScope
 
+CoroFactory = Callable[[], Coroutine[Any, Any, None]]
+FactoryBuilder = Callable[[int], CoroFactory]
+
 
 @runtime_checkable
 class ScopeResolver(Protocol):
--- a/src/lyra/typing/types.py
+++ b/src/lyra/typing/types.py
@@ -29,7 +29,7 @@ class TypingManagerProtocol(Protocol):
     def start(
         self,
         target: int,
-        coro_factory: Callable[[], Coroutine[Any, Any, None]],
+        coro_factory: CoroFactory,
     ) -> None: ...
 
     def cancel(self, target: int) -> None: ...


MickaelV0:
## Code Review

**Verdict:** Approve with comments — single non-blocking DRY suggestion. No blockers, no security findings. Pure type-alias relocation is clean; the one comment surfaces an opportunity now that both pieces live in the same file.

---

**suggestion(non-blocking):** `TypingManagerProtocol.start()` signature uses an inline `Callable[[], Coroutine[Any, Any, None]]` instead of the newly-defined `CoroFactory` alias
  `src/lyra/typing/types.py:29`
  -- architect
  Root cause: The alias and the Protocol were merged into the same file in this PR, but the Protocol method signature was not updated to reference the alias. The inline expression is now a copy of the definition two lines above it, which is DRY-violating and will silently diverge if `CoroFactory` is ever narrowed (e.g. `AsyncGenerator` instead of bare `Coroutine`).
  Solutions:
    1. Replace the inline `Callable[[], Coroutine[Any, Any, None]]` on line 29 with `CoroFactory` (recommended) — same file, zero import cost, makes intent explicit.
    2. Leave inline and add a `# == CoroFactory` comment as weak documentation.
  Confidence: 90%

---

### Thoughts (advisory)

**thought:** `CoroFactory` is re-exported from `lyra.typing.__init__` but has zero external callsites and is not referenced inside `listener.py` after the PR
  `src/lyra/typing/__init__.py:3` / `src/lyra/typing/types.py:6`
  -- architect
  Root cause: `CoroFactory` was never used as a standalone annotation in `listener.py` (only `FactoryBuilder` was). Promoting it pre-emptively is reasonable (issue explicitly calls out T2 callers), but landing it in `__all__` with zero callsites means any static-analysis tool that flags unused re-exports will fire on the next pass.
  Solutions:
    1. Accept the pre-emptive re-export per issue rationale; add a `# T2 forward-export` comment next to the `CoroFactory` line in `__init__.py` to signal intentional.
    2. Defer `CoroFactory` to the first T2 issue that actually uses it.
  Confidence: 80%

**thought:** Bootstrap continues to import `TypingListener` directly from `lyra.typing.listener` rather than the package's public API
  `src/lyra/bootstrap/standalone/adapter_standalone.py:120` / `:257`
  -- architect
  Root cause: The deferred deep import is a deliberate cycle-avoidance device; lazy-import inside function body. No correctness concern — out of scope for this PR.
  Solutions:
    1. Leave as-is.
    2. Switch to `from lyra.typing import TypingListener` if/when the deferred-import pattern is standardised (out of scope).
  Confidence: 75%

---

### Nitpicks

**nitpick:** `__init__.py` imports `TypingListener` from `lyra.typing.listener` rather than from `lyra.typing.types`, making the import order visually inconsistent
  `src/lyra/typing/__init__.py:1`
  -- architect
  Root cause: `TypingListener` is a class, not a type alias, so it correctly lives in `listener.py`. The split is correct by architecture; the visual asymmetry is cosmetic.
  Solutions:
    1. No change needed. The split-source import is correct.
    2. Add a brief inline comment to prevent future incorrect consolidation.
  Confidence: 85%

---

### Praise

**praise:** Clean consolidation — aliases placed before the Protocols they feed, `__all__` is alpha-sorted and explicit, and `listener.py` sheds its `Callable`/`Coroutine`/`Any` stdlib imports entirely. Layer contract preserved: `types.py` already imported from `lyra.transport.work_scope`, so adding two `Callable`-based aliases introduces no new dependency arc.
  `src/lyra/typing/types.py:1-7`
  -- architect
  Confidence: 95%

**praise:** Clean type-alias relocation — no new attack surface introduced
  `src/lyra/typing/types.py:6-7`, `src/lyra/typing/__init__.py:1-15`
  -- security-auditor
  Root cause: `CoroFactory`/`FactoryBuilder` are pure `Callable` type aliases with zero runtime behavior. The move + re-export does not widen the callable API surface, expose previously internal values, add a new import chain through security-relevant code, or introduce transitive imports beyond what already existed (`collections.abc.Callable`, `collections.abc.Coroutine`, `typing.Any` — all stdlib). No secrets, tokens, credentials, or injectable call paths anywhere in the diff or its one-level transitive closure.
  Confidence: 97%

---

### Agents

- architect (1 suggestion, 2 thoughts, 1 nitpick, 1 praise)
- security-auditor (1 praise)

Skipped: tester (no logic change), product-lead (no spec), frontend-dev (no FE), backend-dev (covered by architect), devops (no infra), axial-adr-review (`typing/` not in `{infrastructure, adapters, domains, stages}`).

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 0 findings
**Deferred (issues created):** 0 findings
**Skipped:** 5 findings (2 thoughts not actionable + 2 praise + 1 nitpick with recommended Solution 1 = "no change")
**Failed:** 0 findings
**Enforcement diagnostics:** 0 subsumption violations

### Auto-Applied
- [applied] suggestion(non-blocking): use `CoroFactory` alias in `TypingManagerProtocol.start()` signature — `src/lyra/typing/types.py:32` (architect 90% + backend-dev verifier 95%)

### Skipped
- nitpick: `__init__.py` `TypingListener` import asymmetry (`src/lyra/typing/__init__.py:1`) — architect's recommended Solution 1 was "no change needed; split-source import is correct"
- thought: pre-emptive `CoroFactory` re-export with zero callsites (`src/lyra/typing/__init__.py:3`) — advisory, not actionable
- thought: bootstrap deep import of `TypingListener` (`src/lyra/bootstrap/standalone/adapter_standalone.py:120,257`) — explicitly noted as out of scope for this PR
- praise (×2): clean refactor hygiene + no new attack surface

Label `reviewed` applied.

Issue #1530, opened 2026-05-30
epic(arch): architecture knowledge SSoT — generated current-state + doc-drift gate


MickaelV0:
## Problem

Lyra/Roxabi repos evolve faster than docs+issues are maintained → **systemic silent doc-rot**. 2026-05-30 audit of `docs/architecture/`: **67% live-ref hit-rate** (5/15 sampled `src/…` paths & `CamelCase` symbols dead). Examples: `CliNatsDriver` (deleted #1281) still shown LIVE in `docs/ARCHITECTURE.md` (2d old), `target-architecture.md`, `messaging.md`; ADR-058 `Proposed` symbols (`LyraUserError`/`ErrorBoundaryMiddleware`/`NullMessageManager`) described as implemented. These propagated into the #1490 harness diagram.

Root cause: architecture knowledge has 3 strata with different rot rates — INVENTORY (names/paths/subjects: fast+silent) / STRUCTURE (layers/ownership: medium) / INTENT (why: slow). Mixing them in one artifact lets fast rot poison slow truth → docs untrusted → fallback to guessing or expensive full code-reads.

## Decision (Option 3, 2026-05-30 — see ADR-080)

**2-tier SSoT + doc-drift gate:**
- **Tier 1 — GENERATED current-state ("what is")**: derived in CI from `.importlinter`, `deploy/nats/acl-matrix.json`, `deploy/quadlet.toml`+`cli.py`, `roxabi-contracts` SUBJECTS. Never hand-typed.
- **Tier 2 — hand-authored INTENT ("why")**: ADR domain pages, purged of inventory, `status:`-stamped.
- **Bridge — doc-drift CI gate**: fail CI when a doc cites a symbol/path that no longer resolves.

**Retrieval ladder:** code → generated artifacts → ccc → domain-pages (intent only) → **never** specs/issues for current-state.

`roxabi-live` stays the issue-dependency SSoT (NOT code architecture).

## Children (recommended sequence 1→2→3→4)
1. doc-drift gate (keystone; lands ADR-080)
2. Tier-1 generated snapshot
3. check_subject_literals.py
4. ccc auto-refresh in worktree-setup.sh

## Acceptance
- Silent doc-rot becomes a **loud CI failure**.
- A **generated current-state artifact** exists and is CI-refreshed.
- Humans + agents have a reliable **ordered retrieval source**.

Trigger: #1490 (harness diagram errors).

Issue #1281, opened 2026-05-20
feat(refactor): Phase 4 — LLM drivers thinned to codec


MickaelV0:
**📖 Reference document**: [`artifacts/analyses/1277-stage-axis-refactor-strategy.mdx`](https://github.com/Roxabi/lyra/blob/staging/artifacts/analyses/1277-stage-axis-refactor-strategy.mdx) — full strategic context (root-cause analysis, axis-of-decomposition insight, debt drain map, success criteria, methodology).

---

## Scope

After Phase 1, `cli_nats.py` becomes thin domain on top of `NatsTransport`. Phase 4 generalizes:

- `src/lyra/llm/codec.py` — Protocol
- Per-driver: `claude_codec.py`, `cli_nats_codec.py`, others
- `LlmClient(transport, codec)` composition

HTTP-based drivers (claude direct, etc.) get a sibling `HttpTransport` alongside `NatsTransport`.

## Debt drained

- re-export-init (drivers)
- boundary-broad-catch (driver sites)

## Closed as subsumed

- #1201 (see epic body)

## Success criteria

- Per-driver file ≤ 80 lines
- 0 transport plumbing in driver files
- All providers (Claude direct, NATS-clipool, etc.) function identically

## Estimated effort

~1 week. 1 PR. F-lite.

Parent: #1277
Blocked-by: #1278 (Phase 1)

Issue #1533, opened 2026-05-30
feat(ci): check_subject_literals.py — every lyra.* literal in src/ resolves to acl-matrix or contracts


MickaelV0:
Part of #1530 (pairs with the Tier-1 snapshot child). Grep `src/` for `"lyra\.\w"` string literals; verify each against `deploy/nats/acl-matrix.json` publish/subscribe entries + `roxabi-contracts` SUBJECTS; **fail on orphan subjects** (~15 internal subjects currently scattered as raw literals). Follows the existing `scripts/check_request_reply_flows.py` pattern.

**Acceptance:**
- a raw `lyra.*` literal in `src/` absent from acl-matrix + contracts → CI fails, lists the orphan
- every declared subject passes
- runs in CI

---
**Re-scoped — ADR-081 (2026-05-30):** verify `lyra.*` literals in `src/` against the **CodeInventory** oracle's `subjects` set (`tools/code_inventory.py`, sourced from `acl-matrix.json` + contracts) via `oracle.resolve(subject)`, NOT a bespoke resolver. Consumes the shared oracle.

Issue #1532, opened 2026-05-30
feat(tools): Tier-1 generated architecture snapshot (layers + subjects + process topology)


MickaelV0:
Part of #1530 (recommended after the doc-drift gate child). Emit machine+human-readable **current-state** from existing sources — never hand-typed:
- `.importlinter` → layer/boundary map
- `deploy/nats/acl-matrix.json` → NATS subjects + process identities
- `deploy/quadlet.toml` + `cli.py` → process topology
- `roxabi-contracts` SUBJECTS → typed subject constants

Output a GENERATED doc (e.g. `docs/architecture/CURRENT.generated.md`), refreshed in CI, clearly marked **never hand-edit**.

**Acceptance:**
- `docs/architecture/CURRENT.generated.md` exists: layer map + NATS subjects/identities + process topology
- regenerated in CI (stale = fail or auto-commit); header marks it **never hand-edit**
- excluded from the #1531 doc-drift gate (no self-gating)

---
**Coordination (blocked-by #1531):** this issue emits `docs/architecture/CURRENT.generated.md`, which falls inside the #1531 doc-drift gate scan scope. The gate must **exclude** the generated snapshot (same treatment as `adr/` + the baseline) or it gates its own output. `blocked-by #1531` enforces the gate exists first.

---
**Re-scoped — ADR-081 (2026-05-30):** the Tier-1 snapshot = the **CodeInventory** oracle *serialized* (built in #1531). Do NOT build a separate extractor — `CodeInventory.build(root)` already yields modules/symbols/subjects; this issue serializes that (+ the `.importlinter` layer map) to `docs/architecture/CURRENT.generated.md`, CI-refreshed, marked never-hand-edit. Consumes the shared oracle.

MickaelV0:
📌 **Forward-fit from #1535 review** (oracle keystone — now merged). Inputs for the Tier-1 snapshot that consumes `CodeInventory.to_dict()`:

- **`to_dict()` conflates defined vs imported symbols.** It merges top-level definitions *and* import names into `symbols` (resolve-faithful, intentional). A snapshot consumer cannot distinguish project-*defined* from import-*introduced* without re-running `resolve()`. → either add a `definitions` key (top-level-defined only) to `to_dict()`, or document `symbols` as the superset in the snapshot contract.
[...]

PR #1535, opened 2026-05-30
feat(ci): doc-drift gate + ADR-080 — fail CI on dead doc references

### LINKED ISSUE #1531: feat(ci): doc-drift gate — fail CI when docs cite a symbol/path that no longer resolves
**Keystone** of #1530. Generalize the existing 3-symbol docs-drift guard at `.github/workflows/ci.yml:92` and reuse the `dev-core:recheck` symbol-missing mechanism.

Scan `docs/architecture/` + the CLAUDE.md network for `src/…` paths and `CamelCase` symbols; **fail CI** listing each offending `doc:line → claim → code reality`.

This child also authors **ADR-080** (the decision record) — both land in one PR.

**Acceptance:**
- a known dead ref (e.g. `CliNatsDriver`, deleted #1281) makes the gate fail
- live refs pass
- runs in CI on PRs touching docs or src

---
**Re-scoped — ADR-081 (2026-05-30):** the resolver is now the shared **CodeInventory** semantic oracle (`tools/code_inventory.py` — stdlib `ast` + `acl-matrix.json` + roxabi-contracts), NOT textual regex+grep. Replaces `_classify_token` / `_resolve_*` / `_SKIP_SYMBOLS`. The review blockers (verified against the tree) — `module.ClassName` mis-resolution, substring-grep false negatives, NATS-subject-as-dead-module — are fixed by construction. Baseline regenerated + falsified (121→~85 keys; live symbols/subjects dropped out). ADR-081 refines ADR-080.

MickaelV0:
## Doc-drift gate (keystone of #1530)

Implements **ADR-080** — the bridge of the 2-tier architecture-knowledge SSoT: a CI gate that turns *silent* doc-rot into a *loud* failure.

### What
- `tools/check_doc_drift.py` — scans `docs/architecture/` (excl. the immutable ADR archive), `docs/ARCHITECTURE.md`, `docs/standards/`, and the CLAUDE.md network for backticked `src/…` paths, dotted modules, and `CamelCase` symbols; fails CI when a reference no longer resolves in code.
- Exemptions preserve good patterns: historical annotations (e.g. `deleted #1281`), `<!-- drift-ignore -->`, and a **baseline burn-down list** (`tools/doc_drift_baseline.txt`).
- Wired into `.github/workflows/ci.yml` (generalizes the prior 3-symbol guard).
- ADR-080 + `meta.json` (Foundational section).
- 8 tests: dead / live-path / live-symbol / historical / baselined / drift-ignore / update-baseline / ADR-excluded.

### Baseline = burn-down, not amnesty
The repo currently has **121** pre-existing dead refs in living docs + the CLAUDE.md network (e.g. `CliNatsDriver` deleted #1281 still shown live in `docs/ARCHITECTURE.md`, ADR-058 ghost symbols, dead `messaging/` paths). They are baselined so CI is green **now**; the gate blocks any *new* rot. Purging the baseline is a tracked follow-up under #1530.

The ADR archive (`docs/architecture/adr/`) is intentionally **not** scanned — ADRs are immutable decision records citing as-of-writing symbols.

Closes #1531 · epic #1530 · ADR-080


MickaelV0:
## Code Review

**Verdict: 🔴 Request changes** — 2 *verified* correctness bugs in the resolver. The gate is otherwise well-architected: ships green (ratcheted), lands a strong ADR-080, clean separation of concerns, and catches the canonical case (`CliNatsDriver`). But the symbol/module resolver has two convergent false-result bugs (each flagged independently by backend-dev **and** architect, then verified against the live tree). Both are cheap fixes, and bug **A is already polluting the baseline #1536 will burn down**.

Branch `feat/1531-doc-drift-gate` · 7 files · base `staging` · 5 fresh reviewers (security · architect · tester · devops · backend).

---

### 🔴 Blockers

**issue: `_resolve_module` mis-resolves `module.ClassName` → false positives + 6 LIVE symbols wrongly baselined as dead**
`tools/check_doc_drift.py:147-158` — backend-dev + architect (verified)
Root cause: `Path(*token.split("."))` maps `lyra.outbound.emitter.OutboundEmitter` → `src/lyra/outbound/emitter/OutboundEmitter` (never exists). The trailing CamelCase segment is a class *in* a module and must be stripped, then resolved as a symbol.
Verified: baseline marks `lyra.outbound.emitter.OutboundEmitter`, `lyra.llm.llm_client.LlmClient`, `…OutboundFormatter`, `…OutboundErrorHandler`, `…ThrottleCapability`, `…TurnWriter` as dead — yet `class OutboundEmitter` lives in `src/lyra/outbound/emitter.py`, `class LlmClient` in `src/lyra/llm/llm_client.py`. ⇒ those baseline entries are false **and** any new `module.ClassName` ref in a living doc fails CI spuriously. Corrupts the #1536 burn-down (these aren't rot).
Class: [generator-drift]
Fix: (1, recommended) if `parts[-1][0].isupper()` → resolve module prefix `Path(*parts[:-1])` for the file, then `_resolve_symbol(root, parts[-1])`; (2) strip the class suffix, resolve the module file only. Then regenerate the baseline (the 6 entries drop).
Confidence: 97%

**issue: `_resolve_symbol` substring grep (no `-w`) → deleted symbol silently resolves via a longer name**
`tools/check_doc_drift.py:166-176` — backend-dev + architect (verified)
Root cause: `grep -rqF -- "class WorkerPool"` is fixed-string but still substring → matches `class WorkerPoolClient:`. A deleted `WorkerPool` (with `WorkerPoolClient` present) resolves as live → false negative; the gate's core guarantee fails for prefix-colliding names.
Verified: no `class WorkerPool` exists, yet `grep -rqF "class WorkerPool" src/ packages/` matches (`src/lyra/transport/worker_pool_client.py:63: class WorkerPoolClient`).
Class: [generator-drift]
Fix: add `-w` (whole-word) to all three patterns (`grep -rqFw`), or post-filter in Python with `\bclass {re.escape(token)}\b`.
Confidence: 95%

---

### 🟡 Warnings / suggestions (non-blocking)

- **issue: line-level historical exemption masks co-located dead refs** `:230-234` — `is_historical` is per-line, so one `deleted`/`legacy`/`no longer` exempts EVERY backtick on the line (incl. a real dead ref beside it). Flagged by 4 reviewers. Fix: per-token/proximity scope. The matching test T4 can't distinguish line- vs token-scope → tautological. Conf 80%.
- **suggestion(non-blocking): `--update-baseline` can silently grow the ratchet** `:303-305` — writes `new + baselined`, exits 0; a dev who adds rot then regenerates resets the gate green. Add a stderr WARNING when new violations are absorbed + document the workflow in `tools/CLAUDE.md`. Conf 84%.
- **suggestion(blocking): exit-code contract half-implemented** `:280` — docstring + `tools/CLAUDE.md` promise `exit 2 = script broke`, but `main()` returns only 0/1; an internal error exits 1 (ambiguous with violations). Sharper: `_resolve_symbol` `except OSError: return False` turns a missing-`grep`/env failure into a flood of false violations instead of exit 2. Fix: wrap `main()` → `return 2` on unexpected exception; re-raise the OSError path. Class: [missing-error-handling]. Conf 85%.
- **suggestion(non-blocking): path-traversal gate-bypass via `..`** `:143` — `_SRC_PATH_RE` allows `.`/`/`, so `` `src/lyra/../../README.md` `` resolves true (read-only existence; effect = bypass only). Fix: `.resolve()` + `is_relative_to(root)`. Class: [path-traversal]. Conf 70%.
- **suggestion(non-blocking): package-relative `src/…` paths in `packages/*/CLAUDE.md` → false positives** — tracked in #1536; add an in-code comment referencing #1536 + annotate the 2 baseline entries so they're distinguishable from real rot. Conf 90%.
- **nitpick: CI step name misleading** `.github/workflows/ci.yml:~89` — the new gate runs under the legacy "removed symbols" step name; failures surface mislabeled. Split into its own step ("Check doc drift — backtick refs"). Conf 90%.
- **suggestion(non-blocking): `upper>=2` skips single-prefix classes** (`Hub`, `Router`, `Store`) `:131-133` — valid drift targets silently uncovered. Accept the tradeoff but document it above `_CAMEL_RE`. Conf 75%.

### 🧪 Test coverage gaps (tester)
8 tests are well-structured (tmp_path isolation, AAA, public-CLI boundary), but key branches are untested — several would let a regression through:
- **CLAUDE.md-network scanning** (headline feature) — no test that a dead ref in `src/lyra/*/CLAUDE.md` exits 1.
- `_resolve_module` live/dead — untested (directly relates to blocker A).
- `_SKIP_SYMBOLS` (`KeyError`, `GitHub`) not-flagged — untested.
- `_classify_token` rejections (len>120, whitespace, <2 uppercase) — untested.
- exit code 2 — untestable as written (see blocker-adjacent suggestion).
- dead `src/` path (T2 covers only live).
- T7 doesn't assert baselined entries are merged into the rewrite.

### ✅ Praise
- Subprocess hygiene textbook: `-F` + `--` + list-form + `capture_output` + CamelCase-only tokens = no injection (security-auditor, 98%).
- ADR-080 genuinely strong: 3 real alternatives, honest trade-offs, audit numbers as evidence, formalizes the retrieval ladder (architect).
- ADR-archive exclusion (`adr_dir in f.parents`) correct; clean separation classify/resolve/scan/baseline/CLI (architect).
- CI exit propagation correct under `set -euo pipefail`; `uv run` guarantees the venv (devops, 97%).
- `_collect_claude_mds` dot-prefix filter excludes `.claude/worktrees/*`; baseline output stable/sorted/idempotent (backend, 99%).

---
*Reviewers: security-auditor · architect · tester · devops · backend-dev (fresh, no implementation context). Blockers A & B verified against the live tree by the orchestrator.*

MickaelV0:
## Code Review — CodeInventory oracle (commit 05262efd)

**Verdict: 🔴 Request changes** — 1 *verified* false-negative blocker. The oracle is otherwise excellent (path-traversal fixed & confirmed, subprocess gone, pure-stdlib, pyright-clean project-wide, 0.195s, NATS wildcards correct). 5 fresh reviewers (backend · architect · tester · security · devops). **This re-review validated its own purpose: the oracle fixed the false-*positive* class but introduced a false-*negative* class — caught before #1532/#1533 build on it.**

---

### 🔴 Blocker (verified against the tree)

**issue: unscoped suffix check → `module.<common-name>` resolves live (silent false negative)**
`tools/code_inventory.py:513-521` — backend-dev (95%), verified
Root cause: `_resolve_symbol_at_split` tests `suffix in self.symbols` against the GLOBAL symbol set (all imported names from all modules), not scoped to the prefix module. So `lyra.core.errors` → `exists=True` because `errors` (also `asyncio`, `annotations` from `from __future__`, `logging`, `os`…) is imported somewhere — even though `src/lyra/core/errors.py` doesn't exist and `errors` isn't re-exported in `lyra/core/__init__.py`.
Verified: `resolve("lyra.core.errors")`→exists=True; `resolve("lyra.llm.annotations")`→exists=True; `resolve("lyra.core.asyncio")`→exists=True. And `lyra.core.errors` **dropped from the baseline** (was genuinely-dead in the 121) → a dead ref now passes silently — the exact rot the gate exists to stop.
Class: [parallel-path-drift]
Fix: scope the suffix to the prefix module — `if suffix in self.symbols and prefix_module in self.symbols[suffix]` (the dict already maps name→{modules}). Keeps `…emitter.OutboundEmitter` live (defined in that module) while `lyra.core.errors` correctly fails. Then **regenerate + re-falsify** the baseline (the `module.<common-name>` dead refs must re-appear; subjects stay dropped).

---

### 🟡 Should-fix (cheap, principled — convergent)
- **`_PROJECT_PREFIXES` hardcoded** `:57` (backend 88%) — derive from scanned `packages/` (matches "derive, don't hardcode"); a new package → dead refs silently `kind=unknown`.
- **`$JS`/`$KV` dead subjects → `kind=unknown` (skipped)** `:568-572` (backend 82%) — drop the `_PROJECT_PREFIXES` guard for `$`-namespaces → return `kind=subject, exists=False`.
- **`_nats_matches` accepts `>` mid-pattern** `:109` (backend 78% + architect 75%, convergent) — one-line guard `if ">" in p and p[-1] != ">": return False`.
- **Empty subjects set is silent** `:357` (architect 78% + tester 85%) — if 0 subjects loaded (no acl-matrix + no contracts), every `lyra.*` ref flags dead; emit a stderr WARNING.
- **Token-length guard removed → O(N²)** `:513` (security 62%) — re-add `len(token) > 200 → kind=unknown` (old gate capped at 120).
- **No serialization API** `:375` (architect 80%) — add `to_dict()` so #1532 (Tier-1 snapshot) consumes the oracle instead of re-deriving (the parallel-path-drift the ADR forbids).

### 🧪 Test gaps (tester) — several would have caught the blocker
- **regression pair**: assert a live `module.ClassName` resolves AND a genuinely-dead `module.member` (e.g. `lyra.core.errors`) resolves `exists=False` — pins the falsification.
- `packages/` dead-module; malformed/missing acl-matrix.json; `_nats_matches` edges; `_EXTERNAL_KNOWN_NAMES` suppression; tighten `test_module_dead_project_prefix` (loose `kind in (module,subject)`).

### ✅ Praise (verified)
- **Path-traversal fixed & confirmed**: `resolve("src/../../../etc/passwd")` → `exists=False` (real output) — `..` early-exit + `relative_to(root)` double-check.
- **Subprocess fully removed** — pure-stdlib, zero injection surface (was `grep`).
- **pyright clean project-wide** (0 errors; `include += tools` blast radius safe); gate **0.195s** for 443 files; deterministic baseline; NATS `*`/`>` semantics correct + tested; clean `resolve()` priority (subject before dead-module → `lyra.clipool.cmd` correct).

### ⚠️ Out of scope (NOT this PR — pre-existing deps, track separately)
Security agent's `uv audit` flagged idna 3.11 ReDoS (CVE-2026-45409), starlette Host-header, urllib3 CVEs — unrelated to the oracle; → separate follow-up, not a blocker here.

---
*5 fresh reviewers, no impl context. Blocker verified against the live tree by the orchestrator.*

MickaelV0:
## Re-review resolved — `2db73d1c`

🔴 **Blocker fixed** (unscoped suffix → `module.<common-name>` false negative): the suffix is now scoped to the defining module (`prefix in symbols[suffix]`). Verified against the tree: `lyra.core.errors` → `exists=False` (re-added to baseline), `lyra.outbound.emitter.OutboundEmitter` + re-exports (`lyra.core.Agent`) stay live, no new false positives.

🟡 **Cluster all addressed**: derive `_PROJECT_PREFIXES` from `packages/` · `$JS/$KV/_inbox` dead subjects → `kind=subject,exists=False` · `_nats_matches` rejects `>` non-terminal · stderr warning on empty subjects · token-length guard (`len>200`) · `CodeInventory.to_dict()` (shared serialization contract for #1532).

🧪 **Tests**: +regression pair (a `module.member` dead test — would have caught the blocker), packages dead-module, malformed/missing acl-matrix, `_nats_matches` edges, external-name suppression, `to_dict` determinism.

✅ 75 tests green · ruff + `pyright` (no-arg, CI-equivalent) clean · gate exit 0 · baseline re-falsified **85→88** (genuine dead refs re-appear; subjects + live classes + pkg paths stay dropped).

⚠️ Out-of-scope dep CVEs (idna/starlette/urllib3, flagged by audit) tracked separately → #1541.

MickaelV0:
## Code Review — round 3 (5 fresh agents: security · architect · tester · backend · devops)

**Verdict: ✅ Approve with comments** — no blocker introduced. Several agents used `issue:` labels; on orchestrator review + independent verification each is a test-completeness gap, diagnostic-cosmetic, defensive guard, or pre-existing — none changes gate correctness on the real tree.

### Independently verified (orchestrator, not agent-trusted)
- **Baseline integrity:** 88 unique keys (`git diff` = +88 entries); the gate's "128 baselined" = violation *instances* (same token on multiple doc lines). No silent growth.
- **`ast.TypeAlias` gap → theoretical:** 0 real module-level `type X =` in `src/lyra` (the one grep hit is docstring prose).
- **conditional-import gap → no doc-risk:** sole `except ImportError` is a 3rd-party dep → `kind=unknown` → skipped.
- **Fail-closed:** `set -euo pipefail` propagates exit 1 (violations) AND exit 2 (script error) as build failures (devops retracted its initial concern after verifying).
- Gate green on real tree: `0 new violations`.

### Warnings (non-blocking, worth a follow-up)
- `thought:` **Dead `lyra.x.y` reports `kind=subject` instead of `kind=module`** — `_resolve_dotted_with_subjects` runs the subject check before the project-prefix module check; mirror of the ADR-081 bug but **`exists=False` is still correct** — only the reason string is off. (architect 90% · backend `break`→`continue` 72%) — `tools/code_inventory.py:671`
- `thought:` **`to_dict()` conflates defined vs imported symbols** — forward-fit note for **#1532**: snapshot consumer can't tell project-defined from import-introduced without re-running `resolve()`. (architect 75%) — `tools/code_inventory.py:500`
- `thought:` **`subjects` set contains wildcards** (`*`, `>`) — forward-fit note for **#1533**: `token in subjects` misses matches; consumer must reuse `_subject_matches_any`. (architect)
- `issue:`→reclassified **missing *named* regression** (`lyra.core.errors` dead vs `lyra.core.Agent` live) — mechanism IS pinned by synthetic T20 (falsifiable); named pair is belt-and-suspenders. (tester 85%)
- `suggestion:` malformed-`acl-matrix.json` `identities` (str instead of dict) → generic `AttributeError` (fail-closed, but add `isinstance` guard for a clear message). (security 88%) — `tools/code_inventory.py:356`
- `suggestion:` `_nats_matches` empty-pattern degenerate match — guarded by `_looks_like_nats_subject` today; add `if not subject or not pattern: return False` defensively. (backend 85%)

### Pre-existing (NOT introduced here)
- `issue:` `find … | xargs grep` in the old guard exits 1 on empty `find` output under `set -e` (false failure). Latent since before this PR. (devops 80%) — `.github/workflows/ci.yml:92`

### Suggestions / cleanup
- Dead specific-`lyra.*` regex subsumed by generic `"*" in token` (architect nitpick 95%); last-dot block `659-665` subsumed by inner loop (backend 80%); document the load-bearing import shim + `tools/` pyright-strict coupling in `tools/CLAUDE.md` (devops); a few untested branches (tester).

### Praise
- Oracle/gate boundary clean; `resolve()` a minimal coherent contract (architect).
- Regexes ReDoS-safe; path-traversal double-guarded (syntactic + semantic); fail-closed correct (security).
- Scoped `prefix in self.symbols[suffix]` fix correctly derived; `to_dict()` fully deterministic (backend).
- Scoped-resolution regression test is *falsifiable* — asserts the intermediate symbol-table state, not just output (tester).
- ADR-080/081 substantive, honest on trade-offs, inventory-free (architect).
- Tests: **205 passed, 1 skipped** (75 in the two new files); ruff + pyright(no-arg, CI-equiv) clean.

---

PR #1549, opened 2026-05-30
test(typing): make typing e2e NATS test deterministic (#1543)

### LINKED ISSUE #1543: test(flaky): test_publisher_to_listener_e2e_discord intermittently fails — fixed-sleep race
## Flake

`tests/typing/test_integration_nats.py::test_publisher_to_listener_e2e_discord` fails intermittently:

```
AssertionError: Expected 'start' to have been called once. Called 0 times.
```

Observed: 1 failure in CI run #26685774261 (PR #1542 — a shell-only diff that cannot affect this test) + reproduced locally 1/4 runs.

## Root cause

The test publishes a NATS message then waits a **fixed** `await asyncio.sleep(0.1)` (line 50) before asserting `mgr.start.assert_called_once()`. Under load the publish → NATS Core → listener-callback round-trip doesn't always finish within 100ms, so `mgr.start` has been called 0 times when the assertion runs. Same pattern at line 55 for `mgr.cancel`.

## Fix direction

Replace the fixed sleeps with a bounded wait-for-condition (retry-until `mgr.start.called` with a generous timeout, or an `asyncio.Event` the mock sets) so the test is deterministic regardless of runner load.

- `tests/typing/test_integration_nats.py:50`
- `tests/typing/test_integration_nats.py:55`

Surfaced as a follow-up during /dev #1534 (PR #1542).

MickaelV0:
## Summary
- `test_publisher_to_listener_e2e_discord` flaked under parallel-runner load (~1–4 in 8). Issue framed it as fixed-sleep latency; the real race is wider: **NATS Core drops messages with no registered subscriber**, and both `subscribe()` and `publish()` only buffer locally — so the SUB or the message could miss the server entirely (`mgr.start` called **0** times even after a generous 5s wait, not just late).
- Made deterministic on three fronts:
  - `await nc_sub.flush()` after `listener.start()` → SUB registered on the server before any publish
  - `await nc_pub.flush()` after each publish → message reaches the server, not just the local write buffer
  - replaced the two fixed `asyncio.sleep(0.1)` with a bounded `_wait_until` poll on `mgr.start.called` / `mgr.cancel.called` for residual callback-dispatch latency

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1543: test(flaky) fixed-sleep race | OPEN |
| Implementation | 1 commit on `feat/1543-test-publisher-to-listener-e2e-discord-flaky` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (20/20 under `pytest -n2`) | Passed |

## Test Plan
- [x] `pytest tests/typing/test_integration_nats.py::test_publisher_to_listener_e2e_discord` — looped 20× under `-n2`, 0 failures
- [x] Full `tests/typing/` module green
- [x] Production code untouched — test-only determinism fix

Closes #1543

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) via \`/pr\`

### DIFF
--- a/tests/typing/test_integration_nats.py
+++ b/tests/typing/test_integration_nats.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Callable
 from unittest.mock import MagicMock
 
 import pytest
@@ -13,6 +14,22 @@ from lyra.transport.typing_publisher import TypingPublisher
 from lyra.transport.work_scope import WorkScope
 from lyra.typing.listener import TypingListener
 from tests.nats.conftest import requires_nats_server
+
+
+async def _wait_until(predicate: Callable[[], bool], timeout: float = 5.0) -> None:
+    """Poll ``predicate`` until true or ``timeout`` elapses.
+
+    Deterministic replacement for a fixed ``asyncio.sleep`` when waiting on an
+    async round-trip (publish -> NATS Core -> listener callback): returns as
+    soon as the condition holds, and tolerates a slow runner up to ``timeout``.
+    On timeout it simply returns so the caller's assertion produces the failure.
+    """
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + timeout
+    while not predicate():
+        if loop.time() >= deadline:
+            return
+        await asyncio.sleep(0.01)
 
 
 @requires_nats_server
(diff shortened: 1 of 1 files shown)


MickaelV0:
## Code Review

Fresh-agent review (tester + security-auditor) of the single-file diff.

**Verdict: ✅ Approve with comments** — no blockers; 1 non-blocking, low-confidence suggestion.

**suggestion(non-blocking):** `_wait_until` returns silently on timeout — couples correctness to the caller always following with an assertion.
- `tests/typing/test_integration_nats.py:19-32`
- *Root cause:* returns `None` on both success and timeout. Intentional (docstring documents it) and both current call sites correctly follow with `assert_called_once[_with]`, so the real failure message is produced. Forward-looking risk only: a future caller reusing the helper without an immediate assertion would get a silent pass on timeout.
- *Options:* (1) accept as-is — docstring is explicit, `_` prefix signals private, both call sites correct; (2) raise `AssertionError("timed out after {timeout}s")` for a clearer signal at the cost of changing the contract.
- Confidence: ~55%

### Confirmed sound
- `_wait_until` returning-on-timeout is correct for this pattern — the immediately-following `assert_called_once[_with]` yields a useful failure message.
- `loop.time()` (monotonic) is the right clock; 5.0s/0.01s well-calibrated (local round-trip <50ms).
- `nc_sub.flush()` (SUB registered before publish) + `nc_pub.flush()` (message reaches server) eliminate the NATS Core drop race; poll handles residual callback-dispatch latency.
- A genuinely broken listener still fails (after ~5s) — no regression masking. No test tautology.

Security: no findings (test-only change).

---
🤖 Review via `/code-review`

PR #1630, opened 2026-06-01
feat(typing): T2 — Pool + adapters integration, close #1374 + audio bug class (#1377)

### LINKED ISSUE #1374: fix(discord): typing indicator orpheline en #links — start_typing(channel.id) ≠ cancel_typing(thread.id) après auto-thread
## Bug

Sur Discord, quand un utilisateur poste un message dans un **parent channel** qui déclenche l'auto-thread (`_discord_pre_session_hook` crée un thread T):

1. `discord_inbound.py:244` démarre typing keyed sur `message.channel.id` (parent channel) — convention documentée par le commentaire 241-242
2. `pre_session_hook` crée le thread T et **mute `platform_meta.thread_id = T`**
3. Agent tourne (~19s typique)
4. `discord_outbound.send()` ligne 141 calcule `send_to_id = thread_id if thread_id is not None else channel_id` → utilise **T**
5. Ligne 176 `adapter._cancel_typing(T)` → `TypingTaskManager.cancel(T)` → `self._tasks.pop(T, None)` retourne `None` silencieusement
6. La tâche `_discord_typing_worker(#parent_channel_id)` créée à l'étape 1 **n'est jamais annulée**
7. Elle boucle `await channel.typing(); await asyncio.sleep(9)` jusqu'au restart du process

## Symptôme observé

Production M₁ `lyra-discord` (PID 816490), 2026-05-26 ~10:37→12:13: l'indicateur typing reste actif en permanence dans le channel `#links` après un drop d'URL (`https://github.com/pguso/ai-agents-from-scratch`) suivi d'un run agent `lyra_default` complet et idle. Restart manuel requis pour stopper.

## Evidence

- `src/lyra/adapters/discord/discord_inbound.py:241-244` — typing keyed sur `message.channel.id` (parent) avant pre_session_hook
- `src/lyra/adapters/discord/discord_inbound.py:162-209` — `_discord_pre_session_hook` mute `platform_meta.thread_id`
- `src/lyra/adapters/discord/discord_outbound.py:140-141,176` — outbound utilise `thread_id` pour cancel
- `src/lyra/adapters/shared/_shared.py:189-193` — `TypingTaskManager.cancel(k)` no-op silencieux si clé absente (aucun warning log)

## Fix proposé

**Option recommandée:** keying cohérent partout sur `channel.id` (parent).
- `discord_outbound.py:141` séparer: `send_to_id` (= thread_id si présent, pour `messageable.send()`) reste comme est
- Ajouter `typing_scope_id = channel_id` (toujours parent) et l'utiliser pour `_cancel_typing` (et `_start_typing` dans le path intermediate)
- Threader `typing_scope_id` dans `_make_emitter`, `_build_streaming_callbacks`, `OutboundEmitter.typing_scope_id`

**Alternative low-effort:** cancel les DEUX (channel + thread) en outbound — masque le bug sans corriger l'asymétrie.

**Hardening orthogonal:** `TypingTaskManager.cancel()` devrait `log.debug` si clé absente, pour visibilité future.

## Tests

- Repro: poster une URL nue dans un parent channel #links (sans thread existant), observer typing persistant après "agent idle"
- Test unit: `test_discord_outbound.py` — mocker auto-thread + vérifier `TypingTaskManager._tasks` est vide après `send()`

### LINKED ISSUE #1377: feat(typing): T2 — Pool + adapters integration, close #1374 + audio bug class
**Parent epic:** #1375
**Blocked by:** #1376 (T1 — Core)
**Closes:** #1374 (by construction)

## Scope

Migrate all existing imperative `_start_typing/_cancel_typing` call sites to the publish/subscribe flow established in T1. Single atomic cutover, feature-flagged.

### Pool integration (hub)

- `src/lyra/core/pool/pool_processor_exec.py` — wrap `process_one()` in `async with typing_publisher.scope(work_scope):` context manager. Auto-start on entry, auto-end on exit (including exceptions, asyncio.CancelledError, asyncio.TimeoutError).
- Construct `WorkScope(platform, bot_id, scope_id, trace_id)` from `InboundMessage` at pool entry.
- Verify behavior under all turn outcomes: success, timeout (line 60), generic exception (line 82), cancellation (line 79), backend dead → reset.

### Adapter call site removal — 8 sites

| File | Line(s) | Current code | Replacement |
|---|---|---|---|
| `discord_inbound.py` | 244 | `adapter._start_typing(send_to_id)` | Remove. Pool publishes started. |
| `discord_inbound.py` | 287 | `on_drop=lambda: adapter._cancel_typing(send_to_id)` | Replace with `typing_publisher.publish_ended(scope)` |
| `discord_outbound.py` | 174,176 | `_start/_cancel_typing(send_to_id)` end-of-send | Remove. Pool publishes ended. |
| `discord_outbound.py` | 376-377 | callbacks `start_typing/cancel_typing` | Re-wire via publisher in `OutboundEmitter._handle_typing_tail` |
| `discord_audio.py` | 250 | `adapter._start_typing(message.channel.id)` | Audio path also wraps in `typing_publisher.scope()` — **fixes #1374 bug class in audio flow** |
| `discord_audio.py` | 256 | `on_drop=lambda: adapter._cancel_typing(message.channel.id)` | Replace with `publish_ended` |
| `telegram_inbound.py` | 48, 176 | `adapter._start_typing(chat_id)` | Remove. Pool publishes. |
| `telegram_outbound.py` | 175, 383 | `_cancel_typing(chat_id)` | Remove. Pool/emitter publishes. |

`OutboundEmitter._handle_typing_tail` — refactor to call `typing_publisher.publish_started/ended` based on `outbound.intermediate`. Single source.

### Cleanup

- Delete `_start_typing/_cancel_typing` shims from `OutboundAdapterBase`, `DiscordAdapter`, `TelegramAdapter` after all call sites removed. (Or keep as `@deprecated` for 1 release cycle; PR comment.)
- `TypingTaskManager` stays — it's still the engine, just now invoked exclusively by `TypingListener` (T1).

### Feature flag cutover

- Land code behind `LYRA_TYPING_ENABLED=false` (default off).
- Verify in CI integration tests with flag on.
- Cutover sequence: set `LYRA_TYPING_ENABLED=true` in `~/.lyra/env/hub.env` + adapter envs → restart M₁ services in order (nats, hub, then adapters).
- 24h soak period. If clean, follow-up cleanup PR removes the flag + the deprecated shims.

## Acceptance criteria

- **AC1 — #1374 repro fixed**: drop a bare URL in a Discord parent channel (`#links`) → auto-thread created → agent runs → after `agent idle` log, `TypingTaskManager._tasks` for the bot is empty. Verify via integration test (or post-deploy smoke).
- **AC2 — Audio path fixed**: same repro for an audio message that triggers auto-thread. No orphan task.
- **AC3 — Telegram parity**: regression test — typing still works on Telegram regular reply, voice note, /cmd.
- **AC4 — No double typing**: with flag ON, only the publisher path fires `channel.typing()`. With flag OFF, only the legacy path. Never both.
- **AC5 — Exception safety**: agent timeout / backend dead → `ended` published in finally → no orphan task.
- **AC6 — Grep gate**: `grep -rn '_start_typing\|_cancel_typing' src/lyra/adapters/` returns only the `OutboundAdapterBase` abstract method definitions (or zero after shim removal).
- **AC7 — Prod smoke**: post-deploy, monitor `lyra-discord` journalctl for 24h for any new "typing" warnings or orphan task patterns.

## Migration discipline

- **Single PR**, all 8 sites + Pool wrap. No partial state where some sites use publisher and others use direct calls.
- Old typing primitives (`adapter._start_typing`) remain as thin shims that call `publisher.publish_started/ended` so any missed site degrades gracefully (no orphan, just duplicated event filtered by ref-count).
- Shims removed in a separate cleanup commit after 1-week soak.

## Risk

- **Pool DI timing**: `WorkScope` construction in pool requires access to `bot_id` and `trace_id` from the `InboundMessage` — verify both are populated by the time `process_one` runs (they should be, per inbound pipeline contract).
- **Discord MRO**: `OutboundAdapterBase` mod doesn't add `__init__` (per `adapters/CLAUDE.md` Discord MRO constraint).
- **Re-entrancy on retries**: if agent reset spawns a 2nd `process_one` for the same message, ref-count must not double-emit. Verify pool path.

MickaelV0:
## Summary

Implements T2 of the pub/sub typing migration: pool wrapping + adapter conditional shims + feature flag + CI matrix.

## What changed

- **Pool** (`pool_processor_exec.py`): Wraps `process_one()` in `async with pool.typing_publisher.scope(work_scope):`
- **Pool wiring** (`pool_manager.py`): Direct attribute assignment `new_pool.typing_publisher = self._hub._typing_publisher` (mirrors `pairing_manager` pattern)
- **Outbound** (`_base_outbound.py`, `emitter.py`): `configure_typing_publisher()` + `send_streaming()` injection of `typing_publisher` and `_work_scope`
- **Typing tail** (`_placeholder_lifecycle.py`): `_handle_typing_tail` calls `typing_publisher.publish_started/ended` when `LYRA_TYPING_ENABLED=true`
- **Discord** (`adapter.py`): Conditional shims in `_start_typing` / `_cancel_typing` — pub/sub when flag=true, legacy ThrottleCapability when false
- **Telegram** (`telegram.py`): Same shim pattern as Discord
- **TypingPublisher** (`typing_publisher.py`): `is_typing_enabled()` (default `true`), `@asynccontextmanager scope()` for exception-safe typing lifecycle
- **CI** (`.github/workflows/ci.yml`): Matrix strategy with `LYRA_TYPING_ENABLED=["true", "false"]`
- **Tests**: T10-T14 covering scope(), pool wrap, typing tail, Discord shims, Telegram shims
- **Exemptions**: `discord/adapter.py` + `telegram/telegram.py` over 300 lines (typing shim methods)

## Feature flag

`LYRA_TYPING_ENABLED` (default `"true"`):
- `true` → pub/sub path active, legacy adapter path silent
- `false` → legacy adapter path active, pub/sub path silent
- Never both paths simultaneously

## Verification

- [x] All success criteria from spec (SC1–SC13) satisfied
- [x] CI passes with `LYRA_TYPING_ENABLED=true` and `false`
- [x] Pre-commit hook passes (file length exemptions added)
- [x] pyright ✓ | ruff ✓ | pytest ✓

## Related

- Closes #1377
- Closes #1374
- Closes audio bug class (orphan typing tasks eliminated by pool-scoped lifecycle)

---


MickaelV0:
## Code Review

**Rationale:** 11 blocking findings (≥1 issue-class). Highest confidence: 95% (CI matrix name drift). Key concerns: feature flag default silently flipped, unvalidated `scope_id` parsing, orphan task exceptions from fire-and-forget, and test tautologies.

---

### Spec Compliance

All 13 success criteria from `artifacts/specs/1377-feat-typing-t2-pool-adapters-integration-spec.mdx` are met in the diff.

| SC | Status | Evidence |
|---|---|---|
| SC1 | ✓ | `pool_processor_exec.py:68` wraps `process_one()` in `async with pool.typing_publisher.scope(work_scope):` |
| SC2 | ✓ | `scope_id` parsed from `str` to `int`, `trace_id` from `TraceContext.get_trace_id()` |
| SC3 | ✓ | `TypingPublisher.scope()` added with `@asynccontextmanager` |
| SC4 | ✓ | `discord/adapter.py` conditional shim in `_start_typing` / `_cancel_typing` |
| SC5 | ✓ | `discord/adapter.py` conditional shim in `_start_typing` / `_cancel_typing` |
| SC6 | ✓ | `discord/adapter.py` conditional shim in `_start_typing` / `_cancel_typing` |
| SC7 | ✓ | `discord/adapter.py` conditional shim in `_cancel_typing_for` |
| SC8 | ✓ | `telegram/telegram.py` conditional shim in `_start_typing` / `_cancel_typing` |
| SC9 | ✓ | `telegram/telegram.py` conditional shim in `_start_typing` / `_cancel_typing` |
| SC10 | ✓ | `_placeholder_lifecycle.py:143` calls `typing_publisher.publish_started/ended` |
| SC11 | ✓ | `is_typing_enabled()` controls path (default `"true"`) |
| SC12 | ✓ | CI matrix `typing_enabled: ["true", "false"]` added |
| SC13 | ✓ | AC1–AC7 covered by test suite (T10–T14) |

---

### Findings (Blocking — 11)

**1. Matrix job name drift breaks required checks** (95%)
  `.github/workflows/ci.yml:135`
  -- devops
  Root cause: Adding `strategy.matrix` to the `integration` job changes the reported GitHub status-check names from `integration` to `integration (true)` / `integration (false)`, which breaks branch-protection rules or downstream `needs:` dependencies that expect the old unqualified name.
  Class: [candidate/ci-status-check-drift]
  Raw callsites: [{file: ".github/workflows/ci.yml", line: 135}]
  Solutions:
    1. Add a final `integration` report job that `needs: [integration-matrix]` to keep the old check name stable (recommended)
    2. Revert the matrix and run both env variants sequentially within a single job

**2. Feature flag `LYRA_TYPING_ENABLED` default changed silently from `"false"` to `"true"`** (92%)
  `src/lyra/transport/typing_publisher.py:19`
  -- backend-dev, security-auditor
  Root cause: `is_typing_enabled()` defaults to `"true"` when `LYRA_TYPING_ENABLED` is unset. The `TypingPublisher.__init__` previously defaulted to `"false"`. Existing deployments without the env var will now enable pub/sub typing, generating unexpected NATS traffic.
  Class: [candidate/config-drift, candidate/env-default-drift]
  Raw callsites: [{file: "src/lyra/transport/typing_publisher.py", line: 19}, {file: "src/lyra/transport/typing_publisher.py", line: 25}]
  Solutions:
    1. Revert the default to `"false"` to preserve backward compatibility (recommended)
    2. If intentional, document as a breaking change in release notes and update `docs/CONFIGURATION.md`

**3. Fire-and-forget `asyncio.create_task` for typing publisher calls loses exceptions** (85%)
  `src/lyra/adapters/discord/adapter.py:162`, `src/lyra/adapters/discord/adapter.py:194`, `src/lyra/adapters/telegram/telegram.py:212`, `src/lyra/adapters/telegram/telegram.py:226`
  -- backend-dev, security-auditor
  Root cause: `_start_typing` and `_cancel_typing` are sync methods that spawn background tasks without storing the returned Task or adding a done callback. Any NATS or network failure becomes an unretrieved task exception, invisible to monitoring and retry logic.
  Raw callsites: [{file: "src/lyra/adapters/discord/adapter.py", line: 162}, {file: "src/lyra/adapters/discord/adapter.py", line: 194}, {file: "src/lyra/adapters/telegram/telegram.py", line: 212}, {file: "src/lyra/adapters/telegram/telegram.py", line: 226}]
  Solutions:
    1. Attach a done callback to log/retrieve exceptions: `task.add_done_callback(lambda t: t.exception() and log.warning(...))` (recommended)
    2. Refactor `_start_typing`/`_cancel_typing` to be async and await the publish call

**4. `test_timeout_path_scope_entered_and_exited` does not verify the timeout actually fires** (85%)
  `tests/core/test_pool_processor_exec.py:75`
  -- tester
  Root cause: The test only asserts typing-publisher scope enter/exit, not that the timeout mechanism triggers. If `_turn_timeout` is ignored, `_slow_process` sleeps 1.0 s and the test still passes.
  Class: [test-tautology]
  Raw callsites: [{file: "tests/core/test_pool_processor_exec.py", line: 75}]
  Solutions:
    1. Wrap the call in `asyncio.wait_for(guarded_process_one(...), timeout=0.5)` so the test fails if the internal timeout is ignored (recommended)
    2. Assert that the call returns within a bounded duration (e.g. `time.monotonic()` delta < 0.1 s)

**5. No test verifies `guarded_process_one` when `pool.typing_publisher` is `None`** (80%)
  `tests/core/test_pool_processor_exec.py:38`
  -- tester
  Root cause: All four new tests assign a `TypingPublisher` instance. If the implementation dereferences `pool.typing_publisher` without a `None` guard, the existing tests will not catch the `AttributeError`.
  Class: [candidate/missing-negative-path]
  Raw callsites: [{file: "tests/core/test_pool_processor_exec.py", line: 38}]
  Solutions:
    1. Add a test with `pool.typing_publisher = None` that asserts `process_one` is still invoked and no typing scope calls are made (recommended)

**6. `scope_id` string parsed with `split(":")[-1]` and cast to `int` without validation** (80%)
  `src/lyra/adapters/shared/_base_outbound.py:75`, `src/lyra/core/pool/pool_processor_exec.py:51`
  -- backend-dev, security-auditor
  Root cause: Assumes `scope_id` always contains a colon and the trailing segment is a valid integer; malformed input raises `ValueError` and crashes the turn (or is masked as a generic error in the pool).
  Raw callsites: [{file: "src/lyra/adapters/shared/_base_outbound.py", line: 75}, {file: "src/lyra/core/pool/pool_processor_exec.py", line: 51}]
  Solutions:
    1. Wrap in try/except with a fallback or log warning, e.g. `try: scope_id = int(msg.scope_id.rsplit(":", 1)[-1]) except ValueError: scope_id = msg.scope_id` (recommended)
    2. Introduce a `parse_scope_id` utility with regex validation

**7. `test_typing_disabled_falls_back_to_legacy_methods` asserts `publish_started` is not called but does not assert `publish_ended` is not called** (80%)
  `tests/outbound/test_typing_tail.py:102`
  -- tester
  Root cause: The test only checks one side of the publisher. If the implementation incorrectly calls `publish_ended` (or both), the test still passes.
  Class: [test-tautology]
  Raw callsites: [{file: "tests/outbound/test_typing_tail.py", line: 102}]
  Solutions:
    1. Add `tp.publish_ended.assert_not_awaited()` to the existing test (recommended)

**8. Typing path gating is inconsistent between pool execution and adapter/placeholder lifecycle** (75%)
  `src/lyra/core/pool/pool_processor_exec.py:68`, `src/lyra/adapters/discord/adapter.py:153`, `src/lyra/adapters/telegram/telegram.py:203`, `src/lyra/outbound/_placeholder_lifecycle.py:139`
  -- backend-dev
  Root cause: `pool_processor_exec.py` gates only on `pool.typing_publisher is not None`, while adapters and `_placeholder_lifecycle.py` also check `is_typing_enabled()`. If the env var is False but a publisher instance exists, the pool wraps the turn in typing while the adapter/placeholder falls back to native typing.
  Class: [parallel-path-drift]
  Raw callsites: [{file: "src/lyra/core/pool/pool_processor_exec.py", line: 68}, {file: "src/lyra/adapters/discord/adapter.py", line: 153}, {file: "src/lyra/adapters/telegram/telegram.py", line: 203}, {file: "src/lyra/outbound/_placeholder_lifecycle.py", line: 139}]
  Solutions:
    1. Unify on a single source of truth: remove `is_typing_enabled()` from call sites and let the publisher (or hub wiring) decide by setting `None` when disabled (recommended)
    2. Add `is_typing_enabled()` check to `pool_processor_exec.py` to match the adapter pattern

**9. Fallback trace ID is regenerated independently at each typing boundary, breaking continuity** (75%)
  `src/lyra/adapters/discord/adapter.py:160`, `src/lyra/adapters/discord/adapter.py:192`, `src/lyra/adapters/telegram/telegram.py:210`, `src/lyra/adapters/telegram/telegram.py:224`, `src/lyra/adapters/shared/_base_outbound.py:80`, `src/lyra/core/pool/pool_processor_exec.py:52`
  -- backend-dev
  Root cause: `TraceContext.get_trace_id() or uuid.uuid4().hex` is evaluated at every `WorkScope` creation site. If the trace context is missing, `_start_typing` and `_cancel_typing` in the same turn get different UUIDs, as do the emitter and pool scopes.
  Class: [candidate/trace-id-drift]
  Raw callsites: [{file: "src/lyra/adapters/discord/adapter.py", line: 160}, {file: "src/lyra/adapters/discord/adapter.py", line: 192}, {file: "src/lyra/adapters/telegram/telegram.py", line: 210}, {file: "src/lyra/adapters/telegram/telegram.py", line: 224}, {file: "src/lyra/adapters/shared/_base_outbound.py", line: 80}, {file: "src/lyra/core/pool/pool_processor_exec.py", line: 52}]
  Solutions:
    1. Mint the fallback trace ID once per turn/scope and pass it through to all typing boundaries (recommended)
    2. Ensure `TraceContext` is always populated before entering the adapter layer

**10. `test_typing_disabled_falls_back_to_legacy_methods` only tests `intermediate=True`** (75%)
  `tests/outbound/test_typing_tail.py:102`
  -- tester
  Root cause: The `is_typing_enabled=False` branch is only exercised for intermediate messages. If the implementation mishandles the final-message branch when typing is disabled, the test suite misses it.
  Class: [candidate/missing-negative-path]
  Raw callsites: [{file: "tests/outbound/test_typing_tail.py", line: 102}]
  Solutions:
    1. Add `test_typing_disabled_final_falls_back_to_cancel_typing` with `intermediate=False`, `is_typing_enabled=False`, and `typing_publisher=None` (recommended)

**11. `test_scope_calls_publish_started_on_entry` and `test_scope_calls_publish_ended_on_*` only assert one side of the context manager** (75%)
  `tests/transport/test_typing_publisher.py:92`, `tests/transport/test_typing_publisher.py:102`, `tests/transport/test_typing_publisher.py:113`, `tests/transport/test_typing_publisher.py:125`
  -- tester
  Root cause: Each test asserts a single method (`publish_started` OR `publish_ended`) but never asserts the complementary method is NOT called. If `scope()` incorrectly calls `publish_ended` on `__aenter__`, the entry test still passes.
  Class: [test-tautology]
  Raw callsites: [{file: "tests/transport/test_typing_publisher.py", line: 92}, {file: "tests/transport/test_typing_publisher.py", line: 102}, {file: "tests/transport/test_typing_publisher.py", line: 113}, {file: "tests/transport/test_typing_publisher.py", line: 125}]
  Solutions:
    1. Add `assert_not_awaited()` for the complementary method in each test (recommended)
    2. Consolidate into a single test that asserts ordered `mock_calls == [call.publish_started(scope), call.publish_ended(scope)]`

---

### Findings (Non-blocking — 7)

**12. Add `fail-fast: false` so both typing variants run to completion** (85%)
  `.github/workflows/ci.yml:135`
  -- devops
  Root cause: GitHub Actions matrix defaults to `fail-fast: true`, which cancels the remaining matrix job as soon as one variant fails.
  Class: [candidate/ci-matrix-fail-fast]
  Raw callsites: [{file: ".github/workflows/ci.yml", line: 135}]
  Solutions:
    1. Add `fail-fast: false` under the `strategy:` block (recommended)

**13. Divergent async handling of typing publisher between placeholder lifecycle and adapters** (70%)
  `src/lyra/outbound/_placeholder_lifecycle.py:143`, `src/lyra/adapters/discord/adapter.py:162`, `src/lyra/adapters/telegram/telegram.py:212`
  -- backend-dev
  Root cause: `_placeholder_lifecycle.py` awaits `publisher.publish_started`, creating backpressure; adapters fire-and-forget via `asyncio.create_task`. The two paths behave differently under load or failure.
  Class: [parallel-path-drift]
  Raw callsites: [{file: "src/lyra/outbound/_placeholder_lifecycle.py", line: 143}, {file: "src/lyra/adapters/discord/adapter.py", line: 162}, {file: "src/lyra/adapters/telegram/telegram.py", line: 212}]
  Solutions:
    1. Background both paths with a shared task queue/executor that handles errors (recommended)
    2. Make adapter `_start_typing` async and await, aligning with the placeholder path

**14. WorkScope construction is duplicated across four call sites** (70%)
  `src/lyra/adapters/discord/adapter.py:156`, `src/lyra/adapters/telegram/telegram.py:206`, `src/lyra/adapters/shared/_base_outbound.py:76`, `src/lyra/core/pool/pool_processor_exec.py:53`
  -- backend-dev
  Root cause: No factory helper exists to build a `WorkScope` from adapter/message context; each site manually passes `platform`, `bot_id`, `scope_id`, `trace_id`.
  Class: [candidate/duplication]
  Raw callsites: [{file: "src/lyra/adapters/discord/adapter.py", line: 156}, {file: "src/lyra/adapters/telegram/telegram.py", line: 206}, {file: "src/lyra/adapters/shared/_base_outbound.py", line: 76}, {file: "src/lyra/core/pool/pool_processor_exec.py", line: 53}]
  Solutions:
    1. Add `WorkScope.from_adapter(adapter, scope_id)` or `WorkScope.from_message(msg, trace_id)` classmethod to centralize construction (recommended)
    2. Extract a shared helper in `lyra.transport.work_scope`

**15. External mutation of private `emitter._work_scope` from `OutboundAdapterBase`** (65%)
  `src/lyra/adapters/shared/_base_outbound.py:76`
  -- backend-dev
  Root cause: `send_streaming` sets a private attribute on `OutboundEmitter` from outside the class instead of using a public interface.
  Class: [candidate/encapsulation-break]
  Raw callsites: [{file: "src/lyra/adapters/shared/_base_outbound.py", line: 76}]
  Solutions:
    1. Add a public setter `emitter.set_work_scope(work_scope)` or pass it via `emitter.run(events, work_scope=...)` (recommended)
    2. Keep as-is and document the cross-package coupling

**16. `test_dc_start_typing_enabled_delegates_to_publisher` asserts `mock_create_task.assert_called_once()` without verifying the coroutine argument** (65%)
  `tests/adapters/discord/test_typing_shims.py:35`, `tests/adapters/telegram/test_typing_shims.py:35`
  -- tester
  Root cause: If the implementation passes a dummy coroutine to `asyncio.create_task` instead of the publisher's coroutine, the test still passes.
  Class: [test-tautology]
  Raw callsites: [{file: "tests/adapters/discord/test_typing_shims.py", line: 35}, {file: "tests/adapters/telegram/test_typing_shims.py", line: 35}]
  Solutions:
    1. Capture `mock_create_task.call_args[0]` and assert it is the coroutine returned by `mock_publisher.publish_started(...)` (recommended)

**17. Verify `LYRA_TYPING_ENABLED` is parsed as a string comparison in application code** (60%)
  `.github/workflows/ci.yml:150`
  -- devops
  Root cause: The matrix passes string literals `"true"` and `"false"`; if application code uses Python truthiness (`if os.getenv("LYRA_TYPING_ENABLED"):`), the `"false"` variant is truthy, making both matrix jobs identical and the matrix a tautology.
  Class: [candidate/env-bool-string, test-tautology]
  Raw callsites: [{file: ".github/workflows/ci.yml", line: 150}]
  Solutions:
    1. Verify/adjust application code to parse the flag with `.lower() in ("true", "1", "yes")` (recommended)
  Note: Application code does correctly use `.lower() == "true"`, so this is an advisory verification only.

**18. `test_dc_start_typing_disabled_calls_legacy_path` asserts `callable(mock_start.call_args.args[1])` without verifying the correct typing coroutine** (60%)
  `tests/adapters/discord/test_typing_shims.py:116`, `tests/adapters/telegram/test_typing_shims.py:116`
  -- tester
  Root cause: The test checks that the second argument is callable, but any callable would pass. If the shim passes `lambda: None` instead of the actual typing coroutine, the test passes.
  Class: [test-tautology]
  Raw callsites: [{file: "tests/adapters/discord/test_typing_shims.py", line: 116}, {file: "tests/adapters/telegram/test_typing_shims.py", line: 116}]
  Solutions:
    1. Assert the callable is an `asyncio.coroutine` or verify it returns the expected coroutine (recommended)

---

### Summary

| Dimension | Blocking | Non-blocking |
|---|---|---|
| Correctness | 4 | 1 |
| Security | 2 | 0 |
| Tests | 5 | 4 |
| CI/Infra | 1 | 2 |
| **Total** | **11** | **7** |

### Next Step

`/fix` — apply accepted findings, then re-review.

MickaelV0:
## Review Fixes Applied

Fixes for the 11 blocking findings from code review have been applied:

1. **Typing default reverted** ():  default changed from  back to  — safer gradual rollout
2. **CI matrix protection** (): Added  to matrix strategy +  job to preserve old check name for branch protection
3. **Scope ID safety** (, ): Added  around  with fallback to 
4. **Task exception handling** (, ): Added  to  calls to log exceptions from typing publisher tasks
5. **Test coverage** (): Added  fixture so scope tests exercise the typing path
6. **Test assertions** (, ): Added  for complementary methods
7. **Pyright fix** (): Removed  attribute access (internal CPython detail)
8. **File exemptions** (): Updated caps for  (332),  (333),  (305)

**Quality gates:** ruff ✓ | pyright ✓ | pytest 5362 passed ✓

**Commit:**

MickaelV0:
## Code Review — Second Pass (#1630)

**Verdict: Request changes** — 16 blocking issues found across correctness, tests, and architecture.

---

### 🚫 Blockers (sorted by confidence)

**issue(blocking):** `is_typing_enabled()` defaults to `"false"` — spec SC11 explicitly requires `"true"`
  `src/lyra/transport/typing_publisher.py:18`
  -- backend-dev + tester
  Root cause: Spec SC11 says "default `true` — pub/sub active; `false` — legacy path". Code uses `os.getenv("LYRA_TYPING_ENABLED", "false")`, making legacy path the default. Also, `TypingListener` inlines the same env check at `listener.py:99` instead of calling `is_typing_enabled()`, so a default correction would leave the listener default-off.
  Class: [spec-violation, parallel-path-drift]
  Raw callsites: [{"file": "src/lyra/transport/typing_publisher.py", "line": 18}, {"file": "src/lyra/typing/listener.py", "line": 99}]
  Solutions:
    1. Change default to `"true"`: `os.getenv("LYRA_TYPING_ENABLED", "true")` (recommended)
    2. Replace `TypingListener` inline env check with `is_typing_enabled()` call
    3. If default `"false"` is intentional for safety, amend spec SC11 to match code
  Confidence: 98%

**issue(blocking):** `typing_publisher_shim` done callback crashes on cancelled tasks
  `src/lyra/typing/listener.py:72`
  -- security-auditor + tester
  Root cause: `asyncio.Task.exception()` raises `CancelledError` when called on a cancelled task. The lambda `done_callback` does not guard with `t.cancelled()` before invoking `t.exception()`, causing unhandled exceptions during graceful shutdown.
  Raw callsites: [{"file": "src/lyra/typing/listener.py", "line": 72}]
  Solutions:
    1. Add guard: `if t.cancelled(): return; exc = t.exception(); if exc: log.warning(...)` (recommended)
    2. Replace with named callback using `try/except CancelledError` around `t.exception()` (alternative)
  Confidence: 95%

**issue(blocking):** Identical `_start_typing` / `_cancel_typing` shim blocks duplicated in 2 sibling adapter dirs
  `src/lyra/adapters/discord/adapter.py:154`, `src/lyra/adapters/discord/adapter.py:165`
  `src/lyra/adapters/telegram/telegram.py:204`, `src/lyra/adapters/telegram/telegram.py:215`
  -- axial-adr-review
  Root cause: pub/sub vs legacy branching logic is copy-pasted in `discord/` and `telegram/` instead of living in `adapters/shared/` or a stage primitive. `adapters/CLAUDE.md` requires `_start_typing()` to be a single-line `self._typing.start(...)` call with no extra branching.
  Class: [target-axis-trap]
  Raw callsites: [{"file": "src/lyra/adapters/discord/adapter.py", "line": 154}, {"file": "src/lyra/adapters/discord/adapter.py", "line": 165}, {"file": "src/lyra/adapters/telegram/telegram.py", "line": 204}, {"file": "src/lyra/adapters/telegram/telegram.py", "line": 215}]
  Solutions:
    1. Make `_start_typing` / `_cancel_typing` concrete on `OutboundAdapterBase` with the conditional pub/sub fallback; adapters only provide `_factory_builder` and a `platform` property
    2. Add a `TypingAdapterMixin` in `adapters/shared/` that absorbs the `typing_publisher_shim` + `is_typing_enabled()` branching
  Confidence: 95%

**issue(blocking):** Near-identical test suites duplicated in sibling adapter test dirs
  `tests/adapters/discord/test_typing_shims.py:1`
  `tests/adapters/telegram/test_typing_shims.py:1`
  -- axial-adr-review
  Root cause: Test layer mirrors the N×M production trap; 170+ line suites differ only by platform string and adapter class.
  Class: [target-axis-trap]
  Raw callsites: [{"file": "tests/adapters/discord/test_typing_shims.py", "line": 1}, {"file": "tests/adapters/telegram/test_typing_shims.py", "line": 1}]
  Solutions:
    1. Parameterize tests in a single `tests/adapters/test_typing_shims.py` using pytest fixtures for platform / adapter class / scope_id
    2. Test `typing_publisher_shim` and `OutboundAdapterBase` in isolation; keep one thin integration smoke per adapter
  Confidence: 95%

**test-tautology:** Discord pub/sub path tests do not assert legacy `TypingTaskManager.start` / `cancel` is skipped
  `tests/adapters/discord/test_typing_shims.py:41`, `tests/adapters/discord/test_typing_shims.py:68`
  -- tester
  Root cause: Tests verify `mock_publisher.publish_started/publish_ended` was called, but never mock `adapter._typing.start` or `adapter._typing.cancel` to assert they were NOT called. If the shim accidentally returned `False` or the code fell through to the legacy path, the test would still pass.
  Class: [test-tautology]
  Raw callsites: [{"file": "tests/adapters/discord/test_typing_shims.py", "line": 41}, {"file": "tests/adapters/discord/test_typing_shims.py", "line": 68}]
  Solutions:
    1. Add `with patch.object(adapter._typing, "start") as mock_start:` and `mock_start.assert_not_called()` in the pub/sub path tests
  Confidence: 95%

**test-tautology:** Telegram pub/sub path tests do not assert legacy `TypingTaskManager.start` / `cancel` is skipped
  `tests/adapters/telegram/test_typing_shims.py:39`, `tests/adapters/telegram/test_typing_shims.py:66`
  -- tester
  Root cause: Same pattern as Discord — tests verify the publisher was called but do not guard against the legacy path also firing.
  Class: [test-tautology]
  Raw callsites: [{"file": "tests/adapters/telegram/test_typing_shims.py", "line": 39}, {"file": "tests/adapters/telegram/test_typing_shims.py", "line": 66}]
  Solutions:
    1. Add `patch.object(adapter._typing, "start")` / `patch.object(adapter._typing, "cancel")` and `assert_not_called()` in the pub/sub path tests
  Confidence: 95%

**missing-input-validation:** `test_pool_processor_exec.py` has no test for `LYRA_TYPING_ENABLED=false` with `typing_publisher` present
  `tests/core/test_pool_processor_exec.py:95`
  -- tester
  Root cause: The `pool` fixture seeds `LYRA_TYPING_ENABLED=true`. `test_no_publisher_path_processes_normally` sets `pool.typing_publisher = None` but keeps the flag=true. The branch `is_typing_enabled() and pool.typing_publisher is not None` has no negative test for `is_typing_enabled()` being false when a publisher is present.
  Raw callsites: [{"file": "tests/core/test_pool_processor_exec.py", "line": 19}, {"file": "tests/core/test_pool_processor_exec.py", "line": 95}]
  Solutions:
    1. Add a test that sets `monkeypatch.setenv("LYRA_TYPING_ENABLED", "false")` with `pool.typing_publisher` present and asserts `process_one` runs without `publish_started` / `publish_ended` being called
  Confidence: 90%

**missing-input-validation:** `guarded_process_one` malformed `scope_id` fallback to `0` is not exercised
  `tests/core/test_pool_processor_exec.py:42`
  -- tester
  Root cause: The `msg` fixture always uses `scope_id="chat:42"` which parses cleanly. The `ValueError` branch (`scope_id = 0`) for malformed IDs like `"not_a_number"` is never hit in tests.
  Raw callsites: [{"file": "tests/core/test_pool_processor_exec.py", "line": 42}, {"file": "src/lyra/core/pool/pool_processor_exec.py", "line": 54}]
  Solutions:
    1. Add a test with `make_inbound_message(scope_id="not_a_number")` and assert `guarded_process_one` still succeeds with `scope_id=0` in the constructed `WorkScope`
  Confidence: 90%

**suggestion(blocking):** `_handle_typing_tail` in outbound stage replicates the same pub/sub vs legacy branching
  `src/lyra/outbound/_placeholder_lifecycle.py:137`, `src/lyra/outbound/_placeholder_lifecycle.py:148`
  -- axial-adr-review
  Root cause: Cross-stage typing feature has branching at two boundaries (outbound emitter + adapter) instead of one. `OutboundEmitter` now holds both a `ThrottleCapability` (legacy) and a `TypingPublisher` (pub/sub).
  Class: [target-axis-trap]
  Raw callsites: [{"file": "src/lyra/outbound/_placeholder_lifecycle.py", "line": 137}, {"file": "src/lyra/outbound/_placeholder_lifecycle.py", "line": 148}]
  Solutions:
    1. Unify behind a `TypingCapability` Protocol in `outbound/` with two implementations (legacy ThrottleCapability, pub/sub TypingPublisher); inject the single capability into `OutboundEmitter`
    2. Move the pub/sub dispatch into `emitter._start_typing()` / `emitter._cancel_typing()` so `_placeholder_lifecycle` calls the same method regardless of flag state
  Confidence: 90%

**suggestion(blocking):** `TypingPublisher.scope()` should wrap `publish_started` inside the `try` block to ensure `publish_ended` runs on setup failure
  `src/lyra/transport/typing_publisher.py:56`
  -- backend-dev
  Root cause: `@asynccontextmanager` does not invoke `__aexit__` if `__aenter__` (the code before `yield`) raises. Currently `await self.publish_started` is outside the `try` block. If `publish_started` raises (e.g., `asyncio.CancelledError` during `await self._publish`), the refcount is already incremented but `publish_ended` is never called, leaking the refcount.
  Raw callsites: [{"file": "src/lyra/transport/typing_publisher.py", "line": 56}]
  Solutions:
    1. Move `publish_started` into the `try` block:
       ```python
       try:
           await self.publish_started(work_scope)
           yield
       finally:
           await self.publish_ended(work_scope)
       ```
    2. Add a test that mocks `publish_started` to raise and asserts `publish_ended` is still called
  Confidence: 85%

**test-tautology:** `test_timeout_path_scope_entered_and_exited` wraps `guarded_process_one` in redundant `asyncio.wait_for`
  `tests/core/test_pool_processor_exec.py:91`
  -- tester
  Root cause: `guarded_process_one` catches `asyncio.TimeoutError` internally and returns normally after dispatching a timeout reply. The outer `await asyncio.wait_for(guarded_process_one(...), timeout=0.5)` adds no value and is confusing.
  Class: [test-tautology]
  Raw callsites: [{"file": "tests/core/test_pool_processor_exec.py", "line": 91}]
  Solutions:
    1. Remove the outer `asyncio.wait_for` and just `await guarded_process_one(msg, agent, pool)`
  Confidence: 85%

**missing-error-handling:** Overlapping `TypingPublisher.scope()` calls for the same `WorkScope` are not tested
  `tests/transport/test_typing_publisher.py:97`
  -- tester
  Root cause: `TypingPublisher` maintains a refcount. Two concurrent `scope()` entries for the same key should result in `publish_started` firing once (refcount=1), and the first `publish_ended` should not fire (refcount=1), only the second should (refcount=0). No test verifies this interaction.
  Raw callsites: [{"file": "tests/transport/test_typing_publisher.py", "line": 97}]
  Solutions:
    1. Add a test that nests two `async with pub.scope(scope):` blocks and asserts `publish_started` was called once and `publish_ended` was called once after both exits
  Confidence: 85%

**test-tautology:** `test_scope_calls_publish_ended_on_exception_exit` and `test_scope_calls_publish_ended_on_cancelled_error` do not verify `publish_started` was called
  `tests/transport/test_typing_publisher.py:120`, `tests/transport/test_typing_publisher.py:132`
  -- tester
  Root cause: These tests only assert `publish_ended` was called after the exception. If `scope()` had a bug that skipped `publish_started` and only called `publish_ended` in `finally`, the test would still pass.
  Class: [test-tautology]
  Raw callsites: [{"file": "tests/transport/test_typing_publisher.py", "line": 120}, {"file": "tests/transport/test_typing_publisher.py", "line": 132}]
  Solutions:
    1. Add `pub.publish_started.assert_awaited_once_with(scope)` after the `async with` block in both tests
  Confidence: 85%

**missing-input-validation:** `OutboundAdapterBase.send_streaming` malformed `scope_id` fallback to `_sid=0` is untested
  `src/lyra/adapters/shared/_base_outbound.py:78`
  -- tester
  Root cause: `send_streaming` catches `ValueError` from `int(original_msg.scope_id.rsplit(":", 1)[-1])` and falls back to `_sid=0`. No test exercises this path.
  Raw callsites: [{"file": "src/lyra/adapters/shared/_base_outbound.py", "line": 78}]
  Solutions:
    1. Add a test that passes `scope_id="invalid"` to `send_streaming` (or to `_handle_typing_tail` via the emitter) and asserts `_work_scope.scope_id == 0`
  Confidence: 80%

**missing-error-handling:** `TypingPublisher.scope()` with `enabled=False` is not tested
  `tests/transport/test_typing_publisher.py:97`
  -- tester
  Root cause: All `scope()` tests construct `TypingPublisher(..., enabled=True)`. The `publish_started`/`publish_ended` methods are no-ops when `enabled=False`, but the context manager contract is untested for this branch.
  Raw callsites: [{"file": "tests/transport/test_typing_publisher.py", "line": 97}]
  Solutions:
    1. Add a test `test_scope_noop_when_enabled_false` that uses `enabled=False` and asserts `publish_started`/`publish_ended` are not called
  Confidence: 80%

---

### ⚠️ Non-blocking

**nitpick:** `OutboundEmitter` has two `TYPE_CHECKING` blocks with duplicate imports
  `src/lyra/outbound/emitter.py:44`, `src/lyra/outbound/emitter.py:20`
  -- backend-dev
  Root cause: The diff adds a second `if TYPE_CHECKING:` block after `from lyra.transport._result import Err` that re-imports `TypingPublisher` and `WorkScope`, which are already imported in the first `TYPE_CHECKING` block.
  Solutions:
    1. Remove the second `TYPE_CHECKING` block (lines 44-45)
    2. Consolidate all `TYPE_CHECKING` imports at the top of the file
  Confidence: 100%

---

### ✅ Positives

**praise:** Pool stage correctly composes transport primitive without per-adapter duplication
  `src/lyra/core/pool/pool_processor_exec.py:75`
  -- axial-adr-review
  Root cause: `pool.typing_publisher.scope(work_scope)` wraps `process_one` in one stage-aligned module; zero per-adapter logic. Keep this pattern as the reference for future cross-stage features.
  Confidence: 95%

**note:** `configure_typing_publisher` is MRO-safe — `OutboundAdapterBase` has no `__init__` (confirmed). `configure_typing_publisher` is a post-construction setter that mirrors the existing `configure_tool_display` pattern.

**note:** `TraceContext.get_trace_id() or uuid4().hex` is consistent and safe across all call sites.

**note:** `_handle_typing_tail` does NOT use the async context manager — it calls `typing_publisher.publish_started/ended` directly (single event calls), which is correct for the emitter path.

**note:** No import-layer violation in `lyra.typing.listener` — it imports only `lyra.transport.*` modules, all below `lyra.typing` in the `.importlinter` layers contract.

---

### Summary

| Dimension | Findings | Blocking |
|-----------|----------|----------|
| Correctness | 2 | 2 |
| Security | 1 | 1 |
| Tests | 9 | 9 |
| Architecture | 3 | 3 |
| Readability | 1 | 0 |
| **Total** | **16** | **15** |

**Verdict: Request changes** — 15 blocking issues. Top priority: fix the `is_typing_enabled()` default (spec-violation, 98%) and the `CancelledError` crash in `typing_publisher_shim` (95%). The adapter shim duplication is a `target-axis-trap` that should be resolved before merge to avoid N×M drift.

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 15 finding(s)
**Applied via 1b1:** 0 finding(s)
**Deferred:** 0 finding(s)
**Skipped:** 0 finding(s)
**Failed:** 0 finding(s)

### Auto-Applied

- [applied] `src/lyra/transport/typing_publisher.py` — default `LYRA_TYPING_ENABLED` to `true` (SC11)
- [applied] `src/lyra/transport/typing_publisher.py` — move `publish_started` inside `try` block in `scope()` context manager
- [applied] `src/lyra/typing/listener.py` — guard done_callback with `t.cancelled()` before calling `t.exception()`
- [applied] `src/lyra/typing/listener.py` — replace inline `os.getenv` with `is_typing_enabled()` call
- [applied] `src/lyra/outbound/emitter.py` — remove duplicate `TYPE_CHECKING` block
- [applied] `src/lyra/outbound/emitter.py` — centralize pub/sub vs legacy dispatch in `_start_typing()` / `_cancel_typing()`
- [applied] `src/lyra/outbound/_placeholder_lifecycle.py` — delegate `_handle_typing_tail` to `emitter._start_typing()` / `_cancel_typing()`
- [applied] `tests/adapters/discord/test_typing_shims.py` — add `assert_not_called` for legacy path in pub/sub tests
- [applied] `tests/adapters/telegram/test_typing_shims.py` — add `assert_not_called` for legacy path in pub/sub tests
- [applied] `tests/adapters/test_base_outbound.py` — add `test_send_streaming_malformed_scope_id_fallback`
- [applied] `tests/core/test_pool_processor_exec.py` — add `test_disabled_flag_with_publisher_present`
- [applied] `tests/core/test_pool_processor_exec.py` — add `test_malformed_scope_id_fallback`
- [applied] `tests/transport/test_typing_publisher.py` — add `test_scope_overlapping_calls`
- [applied] `tests/transport/test_typing_publisher.py` — strengthen exception/cancelled tests with `assert_awaited_once_with`
- [applied] `tests/transport/test_typing_publisher.py` — add `test_scope_noop_when_enabled_false`

### Quality Gate

- `ruff check` on modified files: passed
- `pyright` on modified files: 0 errors
- `pytest` on modified test files: 56 passed, 0 failed

---
*Pushed in `66f7290e` — `fix(typing): auto-apply review findings from #1630`*

MickaelV0:
## Code Review — PR #1630 (re-review after fix)

### Verdict: 🔴 Request changes

7 blockers found across architecture, tests, and backend correctness. Spec compliance (SC1–SC13) is clean.

---

### 🔴 Blockers (sorted by confidence)

**issue:** `_handle_typing_tail` re-implements pub/sub typing logic already owned by `OutboundEmitter._start_typing` / `_cancel_typing`
  `src/lyra/outbound/_placeholder_lifecycle.py:137`
  -- architect (95%)
  Root cause: `_handle_typing_tail` explicitly checks `is_typing_enabled + typing_publisher + _work_scope` before calling `publish_started/publish_ended`, but `OutboundEmitter._start_typing()` and `_cancel_typing()` already contain the exact same conditional logic. `_send_placeholder` in the same file correctly delegates to `emitter._cancel_typing()`. This inconsistency creates a maintenance hazard: any change to the pub/sub gating condition must be applied in two places.
  Solutions:
    1. Simplify `_handle_typing_tail` to delegate to `emitter._start_typing()` and `emitter._cancel_typing()`, removing the explicit pub/sub checks and the `is_typing_enabled` import.
    2. Extract the gating condition into a shared helper (e.g. `_should_use_pub_sub(emitter)`) so both sites share the same predicate.

**issue:** Unified-mode typing listeners are started but never stored for teardown
  `src/lyra/bootstrap/factory/wiring_helpers.py:196`
  -- architect (95%)
  Root cause: `_wire_adapters` creates `TypingListener` instances per adapter and calls `await listener.start()`, but the listeners are local variables and are not stored in `WiredAdapters` or returned to the unified lifecycle. `run_lifecycle` in `bootstrap_lifecycle.py` has no reference to stop them during teardown. Standalone mode correctly stores listeners and calls `tl.stop()` in teardown.
  Solutions:
    1. Add `tg_typing_listeners: list[TypingListener]` and `dc_typing_listeners: list[TypingListener]` to `WiredAdapters`, populate them in `_wire_adapters`, and stop them in `run_lifecycle` (mirroring the standalone pattern).

**issue:** `typing_publisher is not None` guard in `_handle_typing_tail` and `_start_typing`/`_cancel_typing` lacks a negative test that FAILS when the guard is deleted
  `tests/outbound/test_typing_tail.py:78`, `tests/outbound/test_emitter_typing.py:80`
  -- tester (95%)
  Root cause: The fallback tests (`test_intermediate_no_typing_publisher_falls_back_to_start_typing`, `test_legacy_fallback_when_typing_publisher_none`) also set `work_scope=None`, so the remaining `and` expression (`is_typing_enabled and work_scope is not None`) stays False even when the `typing_publisher` check is removed. The test passes regardless of the guard.
  Solutions:
    1. Add a test with `typing_publisher=None, work_scope=<set>` (e.g. `WorkScope(...)`). Deleting the guard makes the condition `is_typing_enabled and work_scope is not None` evaluate to True, causing `AttributeError` on `None.publish_started()` — the test would fail.

**issue:** `is_typing_enabled` guard in `_handle_typing_tail` — `test_typing_disabled_final_no_publisher_falls_back_to_cancel_typing` passes even when the guard is deleted
  `tests/outbound/test_typing_tail.py:129`
  -- tester (95%)
  Root cause: `typing_publisher=None` and `work_scope=None` collude to keep the remaining `and` expression False, so the code still falls back to `_cancel_typing()` regardless of `is_typing_enabled`. The test is only a negative test for the fallback path, not for the disabled flag.
  Solutions:
    1. Add a test with `is_typing_enabled=False, typing_publisher=<set>, work_scope=<set>, intermediate=False`. Deleting the guard makes the condition `typing_publisher is not None and work_scope is not None` evaluate to True, causing `publish_ended` to be called instead of `_cancel_typing()` — the test would fail.

**issue:** `missing-error-handling` — `_base_outbound.py` silently swallows `scope_id` parse failure without logging
  `src/lyra/adapters/shared/_base_outbound.py:79`
  -- backend-dev (85%)
  Root cause: `int(original_msg.scope_id.rsplit(":", 1)[-1])` raises `ValueError` on empty or non-numeric `scope_id`; fallback `_sid = 0` is silent. `pool_processor_exec.py:55-58` logs the same failure with `log.warning`, so the omission is inconsistent.
  Solutions:
    1. Add `log.warning("malformed scope_id %r", original_msg.scope_id)` before `_sid = 0`.

**issue:** `missing-error-handling` — `TypingPublisher` refcount diverges from wire state when `_publish` fails
  `src/lyra/transport/typing_publisher.py:33`
  -- backend-dev (75%)
  Root cause: `publish_started` increments `_refcount` before calling `_publish`. If `_publish` fails (exception swallowed per AC5), refcount is 1 but no `started` event was sent. `publish_ended` (called from `scope()` finally) then decrements to 0 and emits `ended` without a matching `started`. Listener receives `ended` → tries to cancel typing that was never started.
  Solutions:
    1. Only commit refcount after `_publish` succeeds, or roll back on failure.

**issue:** `missing-error-handling` — `TypingPublisher._refcount` is not protected against concurrent task access
  `src/lyra/transport/typing_publisher.py:26`
  -- backend-dev (70%)
  Root cause: `typing_publisher_shim` creates fire-and-forget tasks that call `publish_started`/`publish_ended`. The read-modify-write on `_refcount` is not atomic across tasks. A `_cancel_typing` task can interleave between a `_start_typing` task's refcount update and its `_publish` await, producing `ended-for-absent-scope` followed by a dangling `started` that never gets cancelled.
  Solutions:
    1. Guard all `_refcount` accesses with an `asyncio.Lock`.

---

### 🟡 Warnings / Nits

**nit:** Unawaited coroutine objects in legacy-path tests trigger `RuntimeWarning`.
  `tests/adapters/discord/test_typing_shims.py:139`, `tests/adapters/telegram/test_typing_shims.py:137`
  -- tester (90%)
  Solutions:
    1. Use `asyncio.iscoroutinefunction(coro_factory)` instead of `inspect.iscoroutine(coro_factory())`.

**nit:** `test_scope_calls_publish_ended_on_*` exit tests do not assert `publish_started` was called.
  `tests/transport/test_typing_publisher.py:108`
  -- tester (85%)
  Solutions:
    1. Add `pub.publish_started.assert_awaited_once_with(scope)` to each exit test.

**question:** `test_pool_processor_exec.py` `test_disabled_flag_with_publisher_present` relies on `monkeypatch.setenv` rather than patching `is_typing_enabled` directly.
  `tests/core/test_pool_processor_exec.py:96`
  -- tester (75%)
  Solutions:
    1. Use `patch("lyra.core.pool.pool_processor_exec.is_typing_enabled", return_value=False)` for robustness.

---

### 🟢 Praise

- **Spec compliance** — All SC1–SC13 pass (product).
- **Edge coverage** — `test_pool_processor_exec.py` covers success, exception, cancelled, timeout, disabled, malformed scope_id, and no publisher (tester).
- **Real wire behavior** — `test_scope_overlapping_calls` tests actual refcount behavior without mocking (tester).
- **Defensive construction** — `WorkScope` validates `platform`/`bot_id`/`trace_id` via regex before NATS subject construction (backend).

---

### 🔒 Secret Scan

No secrets detected in diff.

MickaelV0:
## Code Review — PR #1630 (re-review after fix)

### Verdict: 🔴 Request changes

Prior pass (7 blockers + 3 nits): **all resolved** except N1 (partial). This pass surfaced 1 correctness asymmetry, 1 state-split, 1 reliability gap, and 2 test issues — all cheap, focused fixes. Remaining items are advisory/follow-up.

---

### ✅ Prior findings — resolved

| Prior | Status | Evidence |
|---|---|---|
| B1 `_handle_typing_tail` dup gating | ✅ RESOLVED | delegates to emitter `_start/_cancel_typing` (`_placeholder_lifecycle.py:136-145`) |
| B2 unified listeners not torn down | ✅ RESOLVED | `WiredAdapters.{tg,dc}_typing_listeners` + `tl.stop()` in `run_lifecycle` (`bootstrap_lifecycle.py:99-102`) |
| B5 silent scope_id parse | ✅ RESOLVED | `log.warning` before fallback (`_base_outbound.py:84`) |
| B6 refcount commit-before-publish | ✅ RESOLVED (started side) | `if ok: self._refcount[key]=new` (`typing_publisher.py:40-42`) — but see 🔴 #1 (ended side) |
| B7 refcount not atomic | ✅ RESOLVED | `asyncio.Lock` both methods (`typing_publisher.py:28,33,48`) |
| B3/B4 guard-kill tests | ✅ RESOLVED | 4 negative tests, **proven falsifiable** (guard deleted → test fails) |
| N2 assert started on scope exit | ✅ RESOLVED | `publish_started.assert_awaited_once_with` |
| N3 patch is_typing_enabled | ✅ RESOLVED | `patch("...pool_processor_exec.is_typing_enabled")` |
| N1 unawaited coroutine | 🟡 PARTIAL | legacy-path fixed; pub/sub-path tests still leak → see 🔴 #4 |

---

### 🔴 New blockers (sorted by confidence)

**issue:** `publish_ended` commits refcount decrement + key deletion *before* `_publish`, then discards the bool return
  `src/lyra/transport/typing_publisher.py:54-58`
  -- backend-dev (92%) — missing-error-handling
  Root cause: B6's fix made `publish_started` commit-after-success, but `publish_ended` still decrements/deletes the key unconditionally then `await self._publish(...)` with the `bool` silently dropped. On ended-publish failure (or CancelledError), the key is gone but no `ended` reached the wire → listener stuck in "started" with no recovery path. Asymmetric with the started side.
  Fix: capture `ok`; on `!ok` keep the key deleted (rollback would re-arm a stale refcount) but `log.warning("typing_publisher: ended publish failed, state may diverge")` for observability.

**issue:** Duplicate `TypingPublisher` — `_build_hub` sets one, `_build_hub_and_wire` overwrites with a second independent instance
  `src/lyra/bootstrap/factory/hub/hub_core.py:75-76`, `src/lyra/bootstrap/factory/hub/hub_assembly.py:100-101`
  -- architect (90%)
  Root cause: `hub_core._build_hub` creates `TypingPublisher(deps.inbound_bus._nc)` + `set_typing_publisher`; `hub_assembly._build_hub_and_wire` then creates a *second* `TypingPublisher(nc)` and overwrites it. Two objects with separate `_refcount` dicts + separate `asyncio.Lock`s; the first (and any state on it) is silently discarded. Dead construction in hub_core + a silent split between the standalone path (hub_core only) and unified path (double-set).
  Fix: remove the construction from `hub_core._build_hub` (leave it to callers), or pass the publisher via `BuildHubDeps` so it's single-sourced.

**issue:** `WorkScope(...)` construction is outside the `try/except ValueError` — malformed `platform`/`bot_id` aborts outbound delivery
  `src/lyra/adapters/shared/_base_outbound.py:86-91`
  -- security-auditor (70%) — missing-input-validation
  Root cause: the `try` wraps only `int(scope_id...)`; the following `WorkScope(...)` runs `_TOKEN_RE` validation in `__post_init__` and raises `ValueError` on non-conformant `platform`/`bot_id`. That propagates unhandled through `send_streaming`, aborting the whole reply (not just typing). Latent DoS via config drift (today platform/bot_id come from operator config).
  Fix: extend the `try` to cover `WorkScope(...)`; on failure set `emitter.typing_publisher = None` + `log.warning` (degrade typing, keep delivery).

---

### 🟡 Test issues

**nit:** N1 only partially fixed — pub/sub-path shim tests leak a coroutine (RuntimeWarning persists)
  `tests/adapters/telegram/test_typing_shims.py:64`, `tests/adapters/discord/test_typing_shims.py:63`
  -- tester (95%)
  Root cause: `test_{tg,dc}_{start,cancel}_typing_enabled_delegates_to_publisher` extract the AsyncMock coroutine via `create_task.call_args.args[0]`, `assert asyncio.iscoroutine(coro)`, but never `coro.close()` (the legacy-path tests got the fix, these didn't). This is the source of the 4 `coroutine was never awaited` warnings.
  Fix: add `coro.close()` after the `iscoroutine` assert in all four pub/sub-path tests.

**suggestion(blocking):** `test_wire_adapters` is a tautology for the new listener wiring
  `tests/bootstrap/test_wiring_helpers.py:796`
  -- tester (88%) — test-tautology
  Root cause: `fake_nc.subscribe = AsyncMock()` lets the test pass, but it never asserts `result.tg_typing_listeners` / `dc_typing_listeners` — the entire `TypingListener` wiring loop (`wiring_helpers.py:197-221`) can be deleted and the test still passes.
  Fix: assert `len(result.tg_typing_listeners) == 1`, `len(result.dc_typing_listeners) == 1`, and `fake_nc.subscribe.assert_awaited()`.

---

### 🟢 Advisory / follow-up (non-blocking)

- **thought (architect 85%, target-axis-trap):** gating predicate copy-pasted across `telegram.py` + `discord.py` `_start/_cancel_typing` (4 blocks). Bounded at 2 platforms; lift to `OutboundAdapterBase` before a 3rd (clipool) appears.
- **suggestion (backend 85%):** `self._lock` held across `await self._publish(...)` serializes concurrent scopes (different keys) through NATS I/O. Narrow the lock to the refcount RMW; publish outside.
- **suggestion (architect 75%):** teardown stops listeners *before* dc-adapters `close()` — a typing event during the close window is dropped (cosmetic; indicator lingers). Consider draining subscribers last.
- **nit (backend 78%):** `scope()` missing `-> AsyncIterator[None]` annotation.
- **issue(non-blocking) (security 65%, data-exposure):** `trace_id` serialized in `TypingEvent` wire payload. Non-secret, ACL-gated subscribers — low impact; exclude from wire or document.
- **thought (backend 70%):** fire-and-forget shim tasks make `started`/`ended` ordering non-deterministic on rapid start→cancel (pre-existing; `scope()` path is immune).

---

### 🟢 Praise

- Guard-kill negative tests (B3/B4) are well-structured and **proven non-tautological** (each fails when its guard is deleted).
- `test_publish_error_swallow` / `test_publish_ended_error_swallow` assert exact refcount state, not just "no raise" — genuine negative gates.
- NATS typing ACL grants present + correctly scoped (`acl-matrix.json` hub `publish lyra.typing.>`, adapters `subscribe lyra.typing.{platform}.>` — cross-tenant spoof blocked).
- `WorkScope.__post_init__` enforces `^[A-Za-z0-9_-]{1,48}$` before subject construction — no NATS subject injection.

### 🔒 Secret scan
No secrets detected in diff.

MickaelV0:
## Review Fixes Applied

Re-review findings addressed in `54f8eb76`. All 5 accepted findings applied; 6 advisory items left as follow-up.

**Applied:** 5 · **Skipped:** 0 · **Deferred:** 0 · **Failed:** 0

### Applied
- ✅ **issue (92%)** `publish_ended` now captures `_publish` bool and `log.warning`s on failure (key stays deleted, no stale re-arm) — symmetric with `publish_started` — `transport/typing_publisher.py`
- ✅ **issue (90%)** single-source `TypingPublisher` — removed the duplicate construction. _Correction to the proposed fix:_ `hub_core._build_hub` is called by **both** `unified.py` (directly) and `hub_assembly._build_hub_and_wire` (standalone), so deleting hub_core would break unified. Instead removed `hub_assembly`'s overwrite (lines 100-101 + unused import); **hub_core is now the sole construction site for both paths** — `bootstrap/factory/hub/hub_assembly.py`
- ✅ **issue (70%)** `WorkScope(...)` moved inside the `try/except ValueError`; on failure → `typing_publisher = None` + warn, outbound delivery proceeds — `adapters/shared/_base_outbound.py` (+ `test_base_outbound.py` contract test updated)
- ✅ **nit (95%)** `coro.close()` added to the 4 pub/sub-path shim tests — RuntimeWarnings eliminated (`never awaited` count now 0) — `tests/adapters/{telegram,discord}/test_typing_shims.py`
- ✅ **suggestion (88%)** `test_wire_adapters` now asserts `len(tg/dc_typing_listeners) == 1` + `subscribe.assert_awaited()` — listener-wiring loop is covered — `tests/bootstrap/test_wiring_helpers.py`

### Verification
- Targeted suites: **76 passed** · `never awaited`: **0**
- Bootstrap regression (`tests/bootstrap`): **209 passed**
- ruff check + format: clean · pyright (touched src): **0 errors**

### Not addressed (advisory — follow-up)
target-axis-trap adapter dup · lock-held-across-publish perf · listener teardown ordering · `scope()` return annotation · `trace_id` on wire · shim start/cancel ordering. None blocking; suitable for a follow-up issue.

Issue #1670, opened 2026-06-02
NATS subjects lyra.* → factory.* — wire contract (roxabi-contracts) + streams/KV/ACL


MickaelV0:
Core of the NATS leg of the lyra→factory migration. **Breaking wire-contract change** —
must cut over in lockstep with every NATS peer (see satellite sub-issues). Deliberately
**excluded** from the package/CLI rename PR #1669 (that PR kept all `lyra.*` subjects intact
and disambiguated the project namespace `factory.` from the subject namespace `lyra.`).

## Why separate
NATS subjects are exact-match strings. A renamed publisher (`factory.llm.generate.request`)
will never reach a subscriber still bound to `lyra.llm.generate.request`. So this cannot land
incrementally — all publishers/subscribers + JetStream stream/consumer/KV names + ACL grants
flip together in one cutover window.

### 1. roxabi-contracts (wire SSoT) — `packages/roxabi-contracts/`
- **132** `lyra.*` subject literals across `src/roxabi_contracts/*/subjects.py`
  (`event`, `metric`, `gh`, `image`, `jobs`, `llm`, `turns`, `voice`, `verify`, `outbound`).
  Rename literal types + f-string builders `lyra.*` → `factory.*`.
- **Breaking** → bump `0.6.0` → `0.7.0`; publish; satellites re-pin (sub-issues).

### 2. JetStream streams / KV (wire identifiers) — currently kept as `LYRA_*`
- Streams `LYRA_OUTBOUND_AUDIO`, `LYRA_AUDIT`, `LYRA_TURNS`; KV bucket `LYRA_STATE`.
- Rename → `FACTORY_*`. Requires stream/consumer re-creation or migration at cutover
  (JetStream stream names are immutable; plan create-new + drain or recreate-on-empty).

### 3. ACL / NATS server config — `deploy/nats/`
- `acl-matrix.json`: all `lyra.*` subjects → `factory.*`; `owner: "lyra"` → `"factory"`.
- Regen `auth.conf` (`make nats-regen-specs`), ACL spec sentinel blocks, fixtures
  (`tests/scripts/fixtures/v3-current.json`). Leave frozen `v3-pre-grant-group.json`.

### 4. Code-side residue
- `src/factory/nats/nats_bus.py` `subject_prefix = "lyra.inbound"` → `factory.inbound`.
- `scripts/subject_literals_allowlist.txt` — drop the `lyra.inbound` allowlist entry once renamed.
[...]

Issue #13, opened 2026-03-02
Machine 2 + Ollama + Qwen 2.5 14B setup


MickaelV0:
## Context
Machine 2 (9800X3D + RTX 5070Ti 16GB). Heavy LLM on demand.

## Deliverables
- Ollama installed on Machine 2
- Qwen 2.5 14B Q6_K loaded (~11GB VRAM)
- FastAPI `/llm` API exposed on local network (OpenAI-compatible)
- Connection test from Machine 1 via httpx

## Acceptance criteria
- >20 tok/s GPU inference
- Correct responses in French
- API accessible from Machine 1

Issue #12, opened 2026-03-02
YouTube: create channel + 3 first manual videos


MickaelV0:
## Context
YouTube algorithm indexes slowly. No audience = no revenue.
Niche: French-speaking dev/AI.

## Deliverables
- YouTube channel created (branding, description, banner)
- 3 first videos published (manual workflow: script → voice → video)
- TTS pipeline validated with voicecli

## Acceptance criteria
- 3 videos published and indexed
- Manual workflow documented (base for future automation)

Issue #11, opened 2026-03-02
MedTech validation: 10-20 posts on medical dictation + reports


MickaelV0:
## Context
MedTech cardio niche. Direct contact: partner is a cardiologist.
Pain point #1: consultation reports = 2-3h/day of data entry.

## Strategy
- Posts on concrete pain points (time lost on data entry, repetitive letters)
- Observe signals (engagement, DMs, comments)
- Constraint: avoid anything related to diagnosis (European MDR)

## Prerequisites
- Interview cardiologist typical day (to be scheduled)

## Acceptance criteria
- 10-20 posts published
- Signals measured (same criteria as LegalTech)

PR #1722, opened 2026-06-03
fix(deploy): resolve 8 cutover-exposed deploy/seed code defects (#1718)

### LINKED ISSUE #1718: #1670 follow-up: genuine CODE defects exposed by factory.* cutover
**Origin:** postmortem of the #1670 `factory.*` cutover (2026-06-03). Tracker: #1710. Epic: #1671.
**Source:** `artifacts/postmortems/2026-06-03-1670-prod-cutover-blockers.md` (full root-cause analysis).

These are the **genuine CODE defects** the cutover *exposed* — rename-independent, will bite any future operator regardless of the `lyra→factory` migration. Bundled into one hardening issue.

### #9 — `install.sh` secret/identity list drifts from Makefile + ACL SSoT
`deploy/install.sh` `SEEDS=(…)` (7 nats secrets, **no gh-helper**) vs `Makefile` `nats-secrets-install` (8, incl. `factory-nats-gh-helper`). Neither is derived from `deploy/nats/acl-matrix.json` (declared SSoT). Adding identity #1082 updated matrix + Makefile but not `install.sh`.
**Fix:** derive `SEEDS` from `acl-matrix.json` in both (or one calls the other) + drift gate asserting `install.sh secrets == acl-matrix identities`.

### #11 — `_BotSeedEntry` breaking-validates existing `config.toml`
`src/factory/agent_cmd/bots/init.py` declares `owner_users: list[str]` + `ConfigDict(extra="forbid")`, but live `config.toml` carries `owner_users = [7377831990]` (**ints**) + a `default` key → `factory bot init` fails validation on all bots. Non-fatal at runtime (hub reads `config.db`, not `config.toml`) but breaks the seed path.
**Fix:** coerce `owner_users` int→str (or `list[int|str]`), don't `forbid` keys already back-compatted, or ship a `config.toml` migration.

### #12 — `make quadlet-install` is destructive-then-fallible (non-atomic)
Target runs `rm -f $QUADLET_DIR/factory*` **first**, then a fallible `uv run factory bot init` → a `bot init` failure (here #11) aborts mid-sequence, leaving units deleted-but-not-recopied (half-installed deployment). Can brick a live host on any transient failure.
**Fix:** render into temp dir + atomic swap, OR move `bot init` before the `rm`, OR make `bot init` non-fatal when `BotStore` already populated.

### #7 — `genkeys` root-guard vs rootless seed ownership
`factory-acl genkeys` (full) requires root, but `--add-identity`/`--regen-authconf` run rootless and seeds are `mickael:mickael 0600`. Running as root would **mis-own** the seeds. Guard predates the `systemd --user` rootless model.
**Fix:** drop/relax the root guard for the rootless seed dir, or document `--add-identity` as the sanctioned single-identity path.

### #10 — split-brain nkeys data-dir resolution (NOT fully traced)
`factory-acl` read/wrote `~/.lyra/nkeys` while `factory_data_dir()` returns `~/.roxabi/factory` (verified). Explicit `ROXABI_FACTORY_DIR` export was ignored. `cli_ops.py` sets `_DEFAULT_SEEDS_DIR = factory_data_dir()/"nkeys"` yet `genkeys` used `~/.lyra` → a second divergent resolution exists.
**Action:** trace the `genkeys` seeds-dir resolution; unify on `factory_data_dir()`; honor `ROXABI_FACTORY_DIR` everywhere.

### #13 — `install.sh` blob symlink over existing dir (minor)
`install.sh` reports `~/.roxabi/factory/blobstore → /data/factory/blobs` but doesn't replace a pre-existing directory → blobs misplaced, manual fixup needed. Assumes clean state.

### #14 — cross-namespace bridge symlinks must be relative
`git_ownership_probe` crash-looped clipool: absolute bridge symlink `~/projects/roxabi-factory → /home/mickael/projects/lyra` dangles inside the container mount namespace. Only a **relative** symlink resolves in both host + bind-mounted container ns.
**Fix:** codify "cross-namespace bridge symlinks MUST be relative" in the deploy standard, OR make `git_ownership_probe` realpath leniently + emit a remediation hint instead of a bare crash. (Real dir rename already done on M₁; this guards future operators.)

MickaelV0:
## Summary
- Bundle of 8 rename-independent deploy/seed defects exposed by the #1670 `factory.*` prod cutover. Spine (#9): `deploy/quadlet.toml` `required_secrets` becomes the enforced secrets SSoT (per `container-deployment-standard.md` S2), derived into `install.sh` via a committed generated manifest and guarded by a new drift gate cross-checking unit `Secret=` directives + acl-matrix (`owner==factory` filter). Six bounded fixes alongside.
- Design driven interactively with architect/devops/axial agents → Shape C (quadlet.toml = declared SSoT, units + acl-matrix = read-only cross-check), avoiding the target-axis-trap of re-rooting the ADR-064 ACL pipeline.

### Defects resolved
| # | Fix |
|---|-----|
| #9  | quadlet.toml `required_secrets` completed for all 8 components; `deploy/secrets-policy.toml` + `tools/emit_secrets_manifest.py` → `deploy/generated/secrets-manifest.sh`; `install.sh` sources it (no hand-maintained `SEEDS`, zero runtime python dep); installs `factory-nats-gh-helper` (the original miss) |
| #9  | new `tools/check_secrets_drift.sh` gate (registered in `.claude/stack.yml`, pre-push): unit `Secret=` ⊆ `required_secrets`, manifest parity, `factory-nats-*` == acl-matrix factory/container identities; optional secrets SKIP; **filters `owner==factory`** (no voicecli/external false-positive) |
| #13 | `install.sh` blob symlink guards a pre-existing non-symlink dir (hint or safe replace) |
| #7  | `scripts/_modes.py`: rootless seed-gen by default; `/etc/nats/nkeys` dual-write opt-in (`FACTORY_ACL_WRITE_ETC_NATS`) — host `nats.service` retired, path confirmed vestigial |
| #10 | `_seeds_dir()` → `factory_data_dir()/nkeys`, honors `ROXABI_FACTORY_DIR`, keeps `SEEDS_DIR`; ends the `~/.lyra` split-brain |
| #11 | `agent_cmd/bots/init.py`: `owner_users`/`trusted_users` accept `list[int\|str]`, field validator coerces → `str`; `_merge_bots` coerces too |
| #12 | `Makefile quadlet-install`: `factory bot init` runs **before** the destructive `rm -f` (no half-installed state on seed failure) |
| #14 | `git_ownership_probe.py`: dangling/absolute bridge symlink → remediation hint (must be relative) instead of generic crash; standard codified (`container-deployment-standard.md` S22) |

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1718: #1670 follow-up CODE defects exposed by factory.* cutover | OPEN |
| Analysis | [analysis](artifacts/analyses/1718-cutover-code-defects-analysis.mdx) | Present |
| Spec | [spec](artifacts/specs/1718-cutover-code-defects-spec.mdx) | Present |
| Implementation | 1 commit on `feat/1718-cutover-code-defects` (F-full, 3 waves / 9 agents) | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (69 new/extended) | Passed |

## Test Plan
- [ ] `bash tools/check_secrets_drift.sh` exits 0 against the tree; flips to non-zero if a unit `Secret=` lacks a `required_secrets` entry
- [ ] `bash deploy/install.sh --dry-run` (with seeds present) lists all 7 `factory-nats-*` incl `factory-nats-gh-helper`
- [ ] `uv run factory bot init` succeeds against `config.toml` with int `owner_users`
- [ ] drift gate does NOT false-positive on `voicecli-nats-*` / external acl-matrix identities

> Note: `~/projects/docs/container-deployment-standard.md` S22 (cross-namespace bridge symlinks MUST be relative) is a shared ecosystem doc outside this repo — edited separately, not in this PR's diff (intentional per analysis).

Fixes #1718

---


MickaelV0:
## Code Review

6 fresh agents (security, devops, backend, tester, architect, product-lead). Spec compliance: **all 8 defect acceptance criteria met** (verified per-criterion). Verdict gated on 3 confirmed blockers below — each independently verified against the tree, not just agent-reported.

### 🔴 Blockers (Request changes)

**1. Drift gate not wired to CI** — `tools/check_secrets_drift.sh`
Spec #9 requires "pre-push **+ CI** on `deploy/`". `.claude/stack.yml` registers it (pre-push), but `.github/workflows/ci.yml` has no step — it runs `check_doc_drift` + `check_architecture_snapshot` but not this gate. A PR touching only `deploy/` secrets could merge without the gate firing in CI.
→ Add a `check_secrets_drift` step to `ci.yml` (alongside the existing gate steps). _(pl-review 95%, arch-review 95%)_

**2. `_require_root()` error names a nonexistent flag** — `scripts/_modes.py:111-112`
Error message: `"error: --write-etc-nats path requires root (sudo factory-acl genkeys --write-etc-nats ...)"`. There is no `--write-etc-nats` flag — the opt-in is the `FACTORY_ACL_WRITE_ETC_NATS=1` env var (confirmed: `gen_nkeys.py` defines no such arg). An operator hitting this would try a flag that errors with `unrecognized arguments`.
→ Rewrite the message to reference `FACTORY_ACL_WRITE_ETC_NATS=1`; also fix `--regenerate`/`--show`/`--fix-perms` help text that still says "root required" unconditionally. _(pl-review 98%)_

**3. `converge.sh` bypasses the manifest path** — `deploy/converge.sh:46`
The automated converge loop calls `make -C ... quadlet-secrets-install` (the hand-listed Makefile target, line 208), **not** the manifest-driven `install.sh`. That target is missing the secrets this PR adds (`factory-nats-blobstore`, `factory_blobstore_token`, and the turn-writer seed) → on every converge cycle those Podman secrets are silently not (re)created. This is the same drift class that caused #1670's half-installed prod. `install.sh --secrets-only` already exists (line 211).
→ Replace the `converge.sh:46` call with `bash "${FACTORY_DIR}/deploy/install.sh" --secrets-only`. _(devops-review 98%)_

### 🟡 Warnings (advisory — fold into the fix pass)

- **Tautological gate test** — `tests/tools/test_emit_secrets_manifest.py` `test_gate_does_not_fail_with_external_identities_present`. Check (c) iterates `factory-nats-*`-prefixed quadlet secrets and looks them up in the ACL set; extra `voicecli-nats-*` entries are never visited, so deleting the `owner==factory` filter would **not** fail the gate → the test can't catch filter removal. Make it a real mutation test (inject a `factory-nats-*`-named non-factory identity). _(test-review 92%)_
- **Stale false-defect doc + dead stub** — `tests/tools/test_install_dry_run.py`. Docstring describes a dry-run `blobstore.tok` bug that the generated-policy dry-run skip already prevents; the `_create_stub_seeds` token stub is dead. Remove + correct docstring. _(test-review 90%)_
- **parse-rule duplication (generator-drift)** — `check_secrets_drift.sh` re-implements `emit_secrets_manifest.py`'s `Secret=` parse rule in bash; guarded only by a comment. Delegate bash→`python3` or add a round-trip contract test. _(devops, arch)_
- **`grep -oP` (PCRE) portability** — `check_secrets_drift.sh:169`. Fine on GNU/Ubuntu CI, breaks on BusyBox/Alpine. Prefer a `python3` parse (already a dependency). _(devops)_
- **`--fix-perms` misses `seeds_dir/auth.conf`** — `scripts/_modes.py` rootless path globs `*.seed` only; the operator-owned `auth.conf` mode is never corrected. _(sec, be — 70%)_
- **install.sh hardening** — missing-manifest `source` aborts with no remediation hint; `secret_name`/`_src` unquoted inside `bash -c` (repo-controlled input, defense-in-depth); dry-run non-empty-dir blob branch falls through to `run ln -sf` (misleading preview). _(sec, devops)_
- **emit_secrets_manifest.py** — `collect_unit_secrets_index()` is dead code; `p["source"]`/`p["policy"]` unguarded (KeyError on malformed policy). _(be)_
- **optional-secret path split** — `secrets-policy.toml` stores `source="n/a"` for `factory-gh-pem`/`factory-claude-oauth` while `install.sh` hardcodes their paths in a `case` (parallel-path). _(arch)_

### ✅ Praise
- All 8 acceptance criteria met: quadlet.toml complete (8/8), manifest derived + committed, `factory-nats-gh-helper` installed, gate's 3 cross-checks + optional-SKIP, rootless genkeys + opt-in `/etc/nats`, `_seeds_dir()` unified (no `.lyra`), int `owner_users` coercion, Makefile atomic, probe relative-symlink hint, S22 standard codified.
- CURRENT.generated.md regenerated cleanly; acl-matrix annotated (no identity change → no ADR-064 fan-out).

---
**Verdict: 🔴 Request changes** — 3 blockers (CI wiring, nonexistent-flag error msg, converge.sh stale path). All low-effort, no design change. Advisory items are good hygiene for the same pass.

MickaelV0:
## Review Fixes Applied

Commit `1dcdb269` (+ prior `6cc859f5`, `da86840a`). All 3 blockers resolved; advisory cleanups folded in; 1 deferred, 1 skipped.

### Blockers — fixed
- **CI wiring** — `check_secrets_drift` step added to `.github/workflows/ci.yml` (was pre-push only).
- **Nonexistent flag** — `_modes.py`/`gen_nkeys.py` error msg + `--help` now reference `FACTORY_ACL_WRITE_ETC_NATS=1` (no phantom `--write-etc-nats`).
- **converge path** — `converge.sh` calls `install.sh --secrets-only` (manifest-derived), no longer the stale `make quadlet-secrets-install` → the new secrets now install on every converge.

### Drift gate hardened (beyond the original findings)
- **check (c) now bidirectional** — forward (quadlet `factory-nats-*` ⊆ acl-matrix factory+container) **+ reverse** (every factory+container identity's `factory-nats-*` secret ∈ quadlet). The `owner==factory` filter — previously **inert** under forward-only — is now load-bearing, matching the spec's "≠" intent. _(surfaced by tester-fix; the original "negative test" couldn't catch filter removal)_
- Gate delegates `Secret=` parsing to `emit_secrets_manifest.py --list-unit-secrets` (one parse rule) and drops `grep -oP` (PCRE) → portability.

### Advisory — applied
- Optional-secret sources moved into `secrets-policy.toml` (concrete `gh-app.pem`/`claude-oauth.tok`); install.sh resolves uniformly via `SECRET_SOURCES` (hardcoded `case` removed).
- `emit`: dead `collect_unit_secrets_index` removed; `.get()` guards on policy entries.
- `_mode_fix_perms` corrects `seeds_dir/auth.conf`; `_mode_regenerate` backups reordered so the destructive wipe runs last.
- `install.sh`: missing-manifest preflight (`exit 2` + hint); dry-run blob symlink no longer prints a misleading `ln -sf`; `bash -c` quoted.
- Tests: real mutation tests for gate (c) reverse + load-bearing filter; removed dead `blobstore.tok` stub + false-defect docstring; `--list-unit-secrets` coverage.

### Deferred → #1724
- Manifest supply-chain integrity (checksum/strict-parse before `source`) — repo-controlled input, single-operator threat model, hardening not a live vuln.

### Skipped
- `ROXABI_FACTORY_DIR` env path-validation — single-operator threat model, out of scope.

**Verification:** ruff ✅ · pyright 0 errors ✅ · drift gate exit 0 (bidirectional) ✅ · 523 passed / 1 skipped ✅ · arch-snapshot ✅

Issue #1740, opened 2026-06-04
deploy: install the two #1729 factory auto-deploy timers on M₁ (quadlet-sync + post-autoupdate)


MickaelV0:
## Problem

M₁'s hands-off deploy is only **partially** live. Of the three auto-deploy timers from #1729 (PR #1732), only `podman-auto-update.timer` is now running (re-enabled + de-poisoned 2026-06-04). The two **factory-specific** timers were never installed on M₁:

| Timer | M₁ state | Role |
|---|---|---|
| `podman-auto-update.timer` | ✅ enabled (fixed 2026-06-04) | image pulls (GHCR digest → restart) |
| `factory-quadlet-sync.timer` | ❌ **not-found** | pull `origin/staging` → `make quadlet-install` on unit/template change |
| `factory-post-autoupdate.timer` | ❌ **not-found** | on image-digest change → full `make converge` (auth.conf regen + secret refresh + ordered restart) |

Consequence: **code-driven** unit/template changes and **post-pull convergence** (auth.conf regen, secret rotation, ordered restart) are NOT automated. Only raw image swaps happen automatically; anything needing a converge is still manual.

## Root cause

The lyra→factory cutover (#1670) installed `factory-*` units via `make quadlet-install`, which does NOT install these timers. Only `deploy/install.sh` does (`make quadlet-sync-install` + the post-autoupdate unit). `install.sh` has not been run end-to-end on M₁ since the rename.

## Acceptance

- [ ] `factory-quadlet-sync.timer` + `factory-post-autoupdate.timer` installed + enabled + active on M₁
- [ ] `OnFailure=factory-deploy-failure.service` wired (journal alert on converge failure)
- [ ] Dry-run / forced trigger proves each fires and converges cleanly (no hung `restart --wait`, stamp written)
- [ ] Verify `install.sh` is idempotent against the current live M₁ state before running (it has not run since cutover — audit for drift first)

## Notes

- Best done as its own `/dev` cycle, not ad-hoc — `install.sh` §10 also re-enables `podman-auto-update.timer` (already done manually) and touches secrets; needs a careful pre-flight against live M₁.
- Audit trail + the autopull half of this fix: memory `project-autodeploy-timers-not-installed`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

MickaelV0:
## Deployed — both #1729 timers now live on M₁ ✅

[...]

Issue #1729, opened 2026-06-04
M₁ auto-deploy is dead: factory-quadlet-sync.timer not-found + podman-auto-update.timer disabled


MickaelV0:
**Origin:** surfaced during the #1060/#1725/#1727 prod deploys (2026-06-04).

## Problem
`docs/ops/container-publishing.md` claims (since #929) M₁ auto-pulls GHCR images via `podman-auto-update.timer` (every 5 min) and that `factory-quadlet-sync.timer` pulls `origin/staging` — *"No manual intervention is needed after a staging merge."* On M₁ (`roxabituwer`) right now:
- `systemctl --user is-enabled podman-auto-update.timer` → **disabled**
- `systemctl --user is-enabled factory-quadlet-sync.timer` → **not-found** (unit absent)

So nothing auto-deploys. M₁ was running pre-#1060 code (image + git checkout from 2026-06-03) until manually pulled/restarted. Every staging merge silently drifts prod from the repo (image **and** the git checkout that renders auth.conf/quadlets), and ACL/auth.conf changes never reach NATS without a manual `factory-acl genkeys --regen-authconf` + secret refresh + `systemctl --user restart factory-nats`.

## Impact
- Doc↔reality drift: the documented "no manual intervention" deploy path is non-functional.
- Silent prod staleness; manual deploys required for every change (image + config).
- Repo↔prod auth.conf drift accumulates (ACL grants linger past their revert).

## Fix (investigate)
- Why is `factory-quadlet-sync.timer` missing — was it never installed, removed, or renamed? Restore/install it (it should `git pull` + `make quadlet-install`).
- Why is `podman-auto-update.timer` disabled — enable it, or update the doc if auto-update was intentionally retired. Confirm `AutoUpdate=registry` labels are present on the `:staging`/`:staging-svc` containers.
- Reconcile `docs/ops/container-publishing.md` with whatever the intended mechanism actually is.

PR #1747, opened 2026-06-04
fix(deploy): drop systemctl enable on generated blobstore unit (#1746)

### LINKED ISSUE #1746: fix(deploy): install.sh aborts at `systemctl enable` on generated quadlet unit — never reaches §9/§10 (timers)
## Problem

`deploy/install.sh:306` runs:

```bash
run systemctl --user enable factory-blobstore.service
```

Quadlet `.container` units are **generator-managed** — they live in `/run/user/<uid>/systemd/generator/` (`systemctl is-enabled` → `generated`) and are enabled via their `[Install] WantedBy=default.target` at `daemon-reload`. `systemctl enable` on such a unit fails:

```
Failed to enable unit: Unit /run/user/1000/systemd/generator/factory-blobstore.service is transient or generated
```

Because install.sh runs under `set -euo pipefail`, this **aborts the script before**:
- §9 `make quadlet-sync-install` (installs the `factory-quadlet-sync` + `factory-post-autoupdate` timers — #1729)
- §10 `systemctl --user enable --now podman-auto-update.timer`

## Impact

A full `bash deploy/install.sh` run has **never completed past §8 on M₁** since `factory-blobstore` was added to the script. This is the root cause of #1740 (the #1729 auto-deploy timers were never installed). Confirmed live on M₁ 2026-06-04 — worked around by running `make quadlet-sync-install` + the auto-update enable manually.

## Why CI/pytest missed it

`tests/tools/test_install_dry_run.py` exercises `install.sh --dry-run`, where the `run` wrapper only **echoes** commands (`[dry-run] systemctl …`). The real `systemctl enable` failure only fires on a live box → classic deploy-only defect (cf. #1721 "CI-green ≠ deploy-ready").

## Fix

Remove `install.sh:306-307` (the `enable` + its `[ok]` echo). `factory-blobstore.container` already carries `WantedBy=default.target`, so it auto-enables on the existing daemon-reload — identical to every other quadlet unit in the bundle (hub, telegram, discord, …), none of which are explicitly `systemctl enable`d.

## Acceptance

- [ ] `install.sh:306-307` removed
- [ ] A full `bash deploy/install.sh` run on M₁ completes through §10 (reaches `make quadlet-sync-install` + auto-update enable) without aborting
- [ ] `test_install_dry_run.py` updated/extended if it asserted the removed enable line

MickaelV0:
## Summary
- Remove the `systemctl --user enable factory-blobstore.service` call from `deploy/install.sh` (§8). Quadlet units are **generator-managed** and auto-enable via their `[Install] WantedBy=default.target` at `daemon-reload`; `systemctl enable` on such a unit fails `Unit … is transient or generated`, and under `set -euo pipefail` that **aborted install.sh before §9** (`make quadlet-sync-install`) **and §10** (auto-update enable) — the root cause of #1740 (the #1729 timers were never installed on M₁).
- `factory-blobstore.container` already carries `WantedBy=default.target`, so it auto-enables exactly like every other quadlet unit in the bundle (hub, telegram, discord, nats…), none of which are explicitly enabled.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1746: install.sh aborts at `systemctl enable` on generated quadlet unit | OPEN |
| Implementation | 1 commit on `feat/1746-install-enable-generated-unit` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (1 new regression test) | Passed |

## Test Plan
- [ ] `uv run pytest tests/tools/test_install_dry_run.py -q` → all pass (11)
- [ ] New `test_blobstore_service_not_explicitly_enabled` fails if the enable line is re-added (mutation-proven)
- [ ] On a live M₁ box: a full `bash deploy/install.sh` run now reaches §9/§10 without aborting (validated manually 2026-06-04 during the #1740 deploy)

> Note: the dry-run pytest cannot reproduce the live failure (the `run` wrapper only echoes commands), so the new test guards against the broken line being re-added rather than the runtime abort itself.

Fixes #1746

---

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -302,7 +302,6 @@ log "Reloading systemd user daemon ..."
 run systemctl --user daemon-reload
 echo "  [ok]   daemon-reload"
 
-
 # ── 8. Seed BotStore from config.toml (idempotent) ─────────────────────────
 # Required since #1416: Authenticator reads from BotStore, not config.toml.
 # Skipping this causes a hub crash-loop on first boot.
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -302,9 +302,6 @@ log "Reloading systemd user daemon ..."
 run systemctl --user daemon-reload
 echo "  [ok]   daemon-reload"
 
-log "Enabling factory-blobstore.service ..."
-run systemctl --user enable factory-blobstore.service
-echo "  [ok]   factory-blobstore.service enabled"
 
 # ── 8. Seed BotStore from config.toml (idempotent) ─────────────────────────
 # Required since #1416: Authenticator reads from BotStore, not config.toml.
--- a/tests/tools/test_install_dry_run.py
+++ b/tests/tools/test_install_dry_run.py
@@ -302,14 +302,21 @@ class TestBlobstoreServiceNotExplicitlyEnabled:
         #   run systemctl --user enable factory-blobstore.service
         # Under --dry-run `run` emits:
         #   [dry-run] systemctl --user enable factory-blobstore.service
-        # If this assertion fires, the line was re-added and must be removed.
-        bad = "systemctl --user enable factory-blobstore.service"
-        assert bad not in combined, (
-            "install.sh must NOT call `systemctl --user enable"
-            " factory-blobstore.service`.\n"
+        # Match any line that both enables AND names the blobstore unit, so the
+        # guard also catches re-introductions that append `--now` (the sibling
+        # §10 `enable --now podman-auto-update.timer` makes that variant likely).
+        offenders = [
+            line
+            for line in combined.splitlines()
+            if "enable" in line and "factory-blobstore.service" in line
+        ]
+        assert not offenders, (
+            "install.sh must NOT `systemctl --user enable` factory-blobstore.service"
+            " (any variant, incl. `--now`).\n"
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Code Review

Fresh-agent review (devops · tester · security-auditor) on the final diff.

**Verdict: Approve** — 1 review finding applied before merge.

### devops
- ✅ `factory-blobstore.container` carries `[Install] WantedBy=default.target` (like all other `factory-*` units) → removing the explicit enable does **not** leave it un-enabled.
- ✅ Removed block was a pure side-effect log/run/echo — no variable or ordering dependency.
- ✅ §8/§9/§10 flow intact.

### security-auditor
- ✅ No silent-disable, no exposure change, no secret/token touched, no injection surface. Clean.

### tester
- ✅ Not a tautology — the guard bites on the exact historical regression (mutation-proven).
- `issue` (88%): exact-string match missed the `enable --now factory-blobstore.service` variant → **fixed** (`ed7dd98b`): now matches any dry-run line containing both `enable` and `factory-blobstore.service`; mutation-proven against the `--now` form.
- ✅ `returncode == 0` precondition is meaningful (prevents vacuous pass on empty output); helpers reused consistently.

All CI green; lint + 11 tests pass locally.

Issue #1800, opened 2026-06-08
feat(obs): live jobs dashboard — v1 kv.watch(registry) → v2 stream factory.job.>


MickaelV0:
**Live jobs dashboard** — v1 `kv.watch(registry)` → v2 stream `factory.job.>`.

## Decision (SSoT)

`artifacts/analyses/job-model-concept-analysis.md` §15 (dashboard substrate — no separate obs plane).

- **v1** — the active-jobs registry KV is JetStream-backed → `kv.watch` = real-time feed. Needs ephemeral-consumer ACLs (vs `kv.get` zero-ACL via #1572 STREAM.MSG.GET).
- **v2** — subscribe the `factory.job.>` stream directly: the job subtree **IS** the trace — every sub-job exchange is already on `factory.job.<id>.*`, so there is no separate observability plane to build.

#1622 (obs epic — OTel/Langfuse on WorkEnvelope) is **CLOSED** → this dashboard is autonomous, not blocked on it.

## Cross-refs

- **Blocked-by:** C (registry — the v1 watch target).
- **Relates:** #1572 (zero-ACL STREAM.MSG.GET — contrast with watch ACLs).

---

## Review refinement (2026-06-08 — product F-07 / architect F6)

AC (blocking before M₁ ship): `acl-matrix.json` MUST grant the dashboard consumer identity the ephemeral-consumer ACL for the `factory-active-jobs` KV (`$JS.API.CONSUMER.CREATE.KV_factory-active-jobs`) — verified in the unit ACL fixture. `kv.watch` needs ephemeral consumers (vs `kv.get` zero-ACL, #1572); ACL misses surface only on M₁ deploy, not CI (cold-boot outage precedent).

MickaelV0:
❄ Gelé post-slice (re-tri 06-10, review §4) — blocked-by #1796 (gelé).

MickaelV0:
Superseded — on reprend proprement via control-plane (#1760) + web smoke adapter (phase 1). Pas d'enfants ouverts. Fermeture pour reset scope.

MickaelV0:
**Deferred 2026-08-01:** v2 stream work tracked as sibling **#2314** under #1792 (v1 remains closed; do not reopen this issue for v2).

Issue #1799, opened 2026-06-08
feat(hub): Shape D steer end-to-end on factory.job.<id>.steer


MickaelV0:
**Shape D steer end-to-end** on `factory.job.<job_id>.steer`.

## Decision (SSoT)

`artifacts/analyses/job-model-concept-analysis.md` §13 (Shape D) + §15.

Close the steerable-stateful loop:

```
inbound steer message
  → concurrency_router (D) resolves the open job via the registry (C)
  → publish on factory.job.<job_id>.steer
  → worker consumes mid-run (job_id durable for the whole run)
```

steer targets `job_id` **natively** — no separate steer_target field — because job_id is durable for the entire run (per ADR-084 amendment, issue B).

## Cross-refs

- **Blocked-by:** C (registry — resolve open job + steer_subject), D (router — fresh|steer|queue decision).
- **Relates:** #1044 (Shape B worker fleet consumes steer), #1490 (lyra-harness stateless worker).

MickaelV0:
❄ Gelé post-slice (re-tri 06-10, review §4) — en outre la forme dépend du verdict #1807 : oh-my-pi fournit `steer()` nativement. Priorité P2→P3.

MickaelV0:
Evidence from the #2142 trace (2026-07-02) relevant to this steer e2e work — **job_id mismatch on the chat path**: registry entries carry a hub-minted uuid4 (`pool_processor.py:54`) while the wire `JobEnvelope.job_id` is minted separately by `llm/drivers/omp_rpc.py:167` → the stored `steer_subject = factory.job.<registry_id>.steer` has no subscriber even for registered jobs (the omp bridge subscribes only `factory.job.<envelope_id>.steer` mid-run). The "job_id durable for the whole run, minted once at dispatch" invariant (ADR-084) is the fix for both. #2142 (now blocking this) keys launch-origin entries by the envelope job_id.

MickaelV0:
**Status 2026-08-01:** blocked-by #2142 + #2147 + #1797. PARTIAL — omp/dashboard steer publish exist; e2e inbound missing.

Issue #1798, opened 2026-06-08
feat(worker): STT/LLM/TTS/image → sub-jobs on factory.job.<id>.* (pub/sub, observable, steerable)


MickaelV0:
**STT / LLM / TTS / image → sub-jobs** on `factory.job.<job_id>.*` (pub/sub, observable, steerable).

## Decision (SSoT)

`artifacts/analyses/job-model-concept-analysis.md` §15 (sub-jobs; what-is-not-pub/sub).

Pipeline stages that cross a **process boundary** become sub-jobs (`parent_job_id` / `composite_depth`), published pub/sub on `factory.job.<job_id>.*`, tracked in the registry, and steerable.

Example (voice→voice): `Telegram voice → Hub (+blob) → STT → transcript → cli_pool → TTS (blob) → Telegram` = a chain of sub-jobs under one root job.

- **Ports survive** — `LlmProvider` / `TtsProtocol` / `STTProtocol` / `AuditSink` are unchanged; only the **transport adapter** migrates from request-reply to sub-job pub/sub.
- **Boundary rule:** cross-process pipeline stage = sub-job; intra-worker hop = opaque (¬job).

## Cross-refs

- **Blocked-by:** C (registry tracks the sub-jobs), E (pub/sub result taxonomy).
- **Relates:** #1048 (NATS-LLM proxy for composite jobs), #1052 (vault.add_from_url composite).

---

## Review refinement (2026-06-08 — axial F2, BLOCKING)

The job-pub/sub transport adapter MUST be a **single shared primitive** (e.g. `factory.transport` job_rpc / sub_job_transport), composed by each port's wiring. Do NOT implement per-port NATS subscribe/publish logic (one LlmJobAdapter + TtsJobAdapter + SttJobAdapter + ImageJobAdapter each rolling their own = the N×M target-axis-trap, ADR-073).
- Ports (`LlmProvider`/`TtsProtocol`/`STTProtocol`/`AuditSink`) stay unchanged; only the transport layer is unified.
- Pattern mirror: `OutboundAdapterBase` — mechanism defined once on the base; per-platform overrides only the formatter.
- Name the shared primitive BEFORE implementation begins (avoids the 3-strikes extraction retrofit).
- **Boundary precision (product F-06):** `cli_pool` internal LLM multi-turn = intra-worker opaque (¬sub-job) even when `cli_pool` is a separate NATS process; only the Hub→cli_pool invocation boundary is a sub-job.

[...]

Issue #1797, opened 2026-06-08
feat(inbound): concurrency_router — single shared stage (fresh|steer|queue)


MickaelV0:
**concurrency_router — single shared inbound stage module** deciding `fresh | steer | queue`.

## Decision (SSoT)

`artifacts/analyses/job-model-concept-analysis.md` §15.3 / §15.6.

- Inbound decision stage = hub rule engine reads the registry index (`pool_id → job_id`) → emits `{ fresh | steer | queue }`.
- **concurrency_mode partition:** `steer | queue` are pool-scoped (resolved here against the singleton index); `parallel` is NOT pool-scoped — it is a worker fan-out capability (keyed job_id only), not an inbound routing decision.
- MUST be **ONE shared stage module** — not per-platform. Axial N×M: a per-adapter copy would drift (N platforms × M concurrency concerns). Same rationale as `make_typing_factory` (ADR-073 target-axis-trap).

## Cross-refs

- **Blocked-by:** C (reads the active-jobs registry).
- **Relates:** #475 (unified channel rule engine — host for the routing rules).
- **Blocks:** G (steer routing).

---

## Review refinement (2026-06-08 — axial F1 / F5)

- Name the concrete module: a single shared inbound stage under `factory/inbound/` (concurrency_router). Adapters bind it via `functools.partial` — NO per-adapter concurrency_router function (mirror `pre_route_hook` / `pre_session_hook`).
- Add an `.importlinter` guardrail: the router must not be importable from `factory.adapters` (analogous to the `inbound-no-adapters` contract, #1287). Placing it under `factory.inbound` enforces this for free.
- Module decomposition follows the **stage axis** (lifecycle), NOT the executor-shape axis (A/B/C/D). Executor shape is a runtime classification injected via strategy/protocol, not a code-organization boundary (§14).

MickaelV0:
❄ Gelé post-slice (re-tri 06-10, review §4) — blocked-by #1796 (gelé). La slice est cron-triggered, pas de routage inbound concurrent nécessaire. Priorité P2→P3.

MickaelV0:
**Status 2026-08-01:** blocked-by #2147 + #2130. DESIGN — concurrency_router not in tree yet.

PR #1863, opened 2026-06-12
feat: factory-active-jobs NATS-KV registry (#1796)

### LINKED ISSUE #1796: feat(infra): active-jobs registry — factory-active-jobs NATS-KV (singleton index, hub-writer, TTL)
**Cross-process active-jobs registry** — `factory-active-jobs` NATS-KV.

## Decision (SSoT)

`artifacts/analyses/job-model-concept-analysis.md` §15.3 / §15.5 + §14.

Today's busy/steerable state lives in-hub-memory (`pool._current_task`, `TypingPublisher._refcount`) — blind to jobs running out-of-process (Shape B/D workers). Lift it into a shared registry.

## Schema

- **key** = `job_id`
- **value** = `{ pool_id, status: open|closing, started_at, steer_subject, concurrency_mode, worker_loc? }`
- **index** `pool_id → job_id` — **SINGLETON** (the pool serialises turns; parallelism lives in the job-tree via `parent_job_id`, not in the pool index).
- **writer = HUB-only** (heartbeat-message pattern, precedent `factory.clipool.heartbeat` / ADR-075).

## Lifecycle

- **CLOSE** = terminal `JobResult` (authoritative, event-driven).
- **heartbeat** = TTL-refresh fallback.
- **reap** = TTL lapse.
- job_id=run makes `lifespan(entry) = lifespan(run) =` the busy/steerable window.

## concurrency_mode partition

`steer | queue` = pool-scoped (singleton index). `parallel` = NOT pool-scoped (worker fan-out, keyed job_id only).

## Cross-refs

- **Blocked-by:** B (job_id=run ratified), #1778 (in-process JobContext carrier this mirrors).
- **Relates:** #1624 (worker-driven typing reads the same active-job state).
- **Blocks:** D, F, G, H.

---

## Review refinement (2026-06-08 — architect F3/F4, dependency F-03)

- **Dependency (F-03 resolved → Option 3):** blocked-by #1794 (job_id=run ratified) + #1619 (WorkEnvelope = the job_id source). **Relates #1778** — the registry is the cross-process *mirror* of the in-process JobContext, but is constructible from the envelope and does NOT wait on the carrier refactor.
- **Heartbeat producer (per shape):** Shape D (remote stateful worker) → the worker self-heartbeats (hub holds no in-process ref); Shapes A/B/C → the hub drives TTL refresh from `Pool._current_task`. Specify the producer explicitly — otherwise crashes leave stale `open` entries until TTL lapse.
- **`parallel` mode (F4):** the registry replaces `TypingPublisher._refcount` only for `steer`/`queue` (pool-scoped singleton). `parallel` allows multiple open jobs on the same `(platform, bot_id, scope_id)`; the registry is keyed `job_id`, the refcount keyed `(platform, bot_id, scope_id)`. Preserve ref-counting semantics for `parallel` (or keep `TypingPublisher` alongside for that mode).

MickaelV0:
## Summary

Adds the **`factory-active-jobs` NATS-KV registry** — the single-writer liveness substrate that enforces *one active job per pool* and exposes the live job set for the dispatch-side concurrency gate (epic #1792). Foundation child; blocks #1797 / #1798 / #1799 / #1800.

Closes #1796.

## What's implemented

| Slice | Module | Behavior |
|-------|--------|----------|
| 1 | `core/ports/active_jobs.py` | `ActiveJobsPort` Protocol + `ActiveJobEntry` (frozen) + `RegistryConflictError` |
| 1 | `infrastructure/stores/active_jobs_kv.py` | `KvActiveJobsStore` over `factory-active-jobs` bucket (TTL 120s); **create-only CAS** on the pool-index key → second opener of a busy pool is rejected, same `job_id` re-open is idempotent; `direct=False` pinned |
| 2 | `infrastructure/stores/active_jobs_refresher.py` | `RegistryCoordinator` — in-memory job map + periodic `refresh_all()` (pool-source liveness, 30s) + `on_heartbeat(loc)` (heartbeat-source) |
| 2 | `transport/worker_pool_client.py` | optional `on_heartbeat` callback (registry tap-in) |
| 2 | `bootstrap/standalone/hub_standalone.py` | store `connect()` + coordinator `start()`/`stop()` wired into hub boot |

- ACL: hub granted publish `$KV.factory-active-jobs.>`. All four generated ACL artifacts regenerated + `v3-pre-grant-group.json` hand-mirrored.
- Docs: messaging KV-plane catalogue row + job-model.md status (`Done (substrate); live open/close → #1797`).

## Deferred (by design)

- **Live `open()`/`close()` call-sites** — the registry exposes them and the refresher runs (pool-source), but wiring them into the live pipeline is **#1797** (concurrency_router, the first consumer — polled, no clean run-completion hook). Substrate + tests land here.
- **Heartbeat pass-through** — `WorkerPoolClient.on_heartbeat` exists and is tested, but is not yet connected to the coordinator (`worker_loc` stays dormant until the Shape-D workers populate it).
- **`auth.conf` regen** — CI-only (NKey seeds).

## Verification

- 60 stores + refresher tests green; 116 transport tests green; full suite 2014 pass (2 pre-existing `test_agent_refiner_cli` typer failures on staging, unrelated).
- Gates green: `str_exc_bus_bound`, `doc_drift`, `secrets_drift`, `folder_size` (15/15), `import-linter` (11/0). `typing_publisher.py` untouched.
- nats-py KV signatures verified at source: `KeyValueConfig.direct`, `kv.create()` → `KeyWrongLastSequenceError`, `kv.update(last=)`.

## Axial

Touches `core/ports/` + `infrastructure/` → carries `dev-core:axial-adr-review`. Core port stays infra-free (port-injection); no infrastructure types leak into the domain object.

### DIFF
--- a/artifacts/frames/1796-active-jobs-registry-frame.mdx
+++ b/artifacts/frames/1796-active-jobs-registry-frame.mdx
@@ -1,0 +1,59 @@
+---
+title: "feat(infra): active-jobs registry — factory-active-jobs NATS-KV (singleton index, hub-writer, TTL)"
+issue: 1796
+status: approved
+tier: F-lite
+date: 2026-06-11
+---
+
+## Problem
+
+The hub's busy/steerable state lives in-process only: `pool._current_task` (`src/factory/core/pool/pool.py:102` — is a run active on this conversation?) and `TypingPublisher._refcount` (`src/factory/transport/typing_publisher.py:27`, keyed `tuple[str, str, int]`). The moment a job runs out-of-process (Shape B workers #1044, Shape D stateful runs #1792), the hub is blind: it cannot answer "is pool X busy?", cannot route an inbound message to `steer | queue | fresh` (#1797), and a dashboard cannot show live jobs (#1800 v1).
+
+#1796 lifts that state into a shared **`factory-active-jobs`** NATS-KV registry — keyed `job_id`, with a `pool_id → job_id` singleton secondary index — so cross-process job state survives and is observable. It is the foundation child of epic #1792: #1797 (router), #1798 (sub-jobs), #1799 (steer), #1800 (dashboard) all read it.
+
+**Drift since issue creation (2026-06-08):** the cited SSoT `artifacts/analyses/job-model-concept-analysis.md` was graduated and archived with a SUPERSEDED banner (`a5a20301`). The live SSoT is **`docs/architecture/job-model.md`** (§Active-jobs registry, §Concurrency, §Observability) — schema, writer rule, liveness model, and mode partition are all ratified there. No design re-litigation needed; the frame anchors on the graduated doc.
+
+## Who
+
+- **Primary:** hub (sole writer — opens/closes/refreshes entries per run) and the #1797 concurrency_router (first reader: `fresh | steer | queue` decision at ingress).
+- **Secondary:** #1800 dashboard v1 (`kv.watch` on the bucket), #1624 worker-driven typing (reads the same active-job state), #1799 steer (resolves `steer_subject` from the entry), ops (live-jobs visibility on M₁).
+
+## Constraints
+
+- **Ratified schema (job-model.md, not re-decidable here):** key = `job_id`; value `{pool_id, status: open|closing, started_at, steer_subject, concurrency_mode, worker_loc?}`; secondary index `pool_id → job_id` **singleton**, enforced for `steer`/`queue` modes only; `parallel` mode is NOT pool-scoped (keyed `job_id` only, no index entry).
+- **Writer = HUB only.** Workers never write the registry. Shape D liveness uses the **heartbeat-message pattern** (precedent `factory.clipool.heartbeat`, `clipool_worker.py` `_HEARTBEAT_SUBJECT` / **ADR-049** heartbeat contract — note: the issue body mis-cites ADR-075, which is the TurnWriter subscriber, not heartbeats): the worker publishes a heartbeat message; the hub consumes it and performs the TTL-refreshing KV re-put. Shapes A/B/C: the hub drives TTL refresh from `pool._current_task`. The spec must fix the heartbeat subjects and refresh cadence explicitly — unresolved producers leave stale `open` entries until TTL lapse (2026-06-08 review refinement).
+- **Lifecycle (ratified):** CLOSE = terminal `JobResult` (authoritative, event-driven delete); heartbeat = TTL refresh fallback; reap = TTL lapse safety net for crash-before-terminal.
+- **KV key charset (project memory, #1721 outage):** `pool_id` contains colons (`telegram:lyra:…`) which nats-py `VALID_KEY_RE` rejects — silent total-outage class. All KV stores MUST key through `src/factory/infrastructure/stores/_kv_keys.py`. The `pool_id → job_id` index keys MUST be sanitized through the same helper.
+- **Provisioning pattern (paved):** hub-only create-or-open **before `announce_hub_ready()`** — the model is `factory-state` in `bootstrap/wiring/kv_watch_channels.py` (hub-only, pre-ready). (`blobstore/serve.py` also does create-or-open but as sidecar best-effort — NOT the model here.) New bucket = new ACL surface: `$KV.factory-active-jobs.>` publish for hub → **ACL fan-out** (regen 4 artifacts + hand-mirror `tests/scripts/fixtures/v3-pre-grant-group.json`; equivalence checks CI-only).
+- **Bucket config pinned day-1:** spec must pin `allow_direct` on `KeyValueConfig` at creation (changing it later = bucket recreation). The choice shapes the #1800 read path and reader ACLs — document the rationale in the spec.
+- **Reader ACL (#1797/#1800):** readers inside the hub process need no new grant. Out-of-hub readers SHOULD use `kv.get` (zero-ACL via #1572 `STREAM.MSG.GET`) rather than `kv.watch` (needs ephemeral-consumer grants); if `kv.watch` is chosen for any reader, it is a separate ACL grant with its own regen pass (#1800's problem, but the spec must not foreclose it).
+- **TTL semantics:** TTL is bucket-level (`max_age`) in NATS KV; per-key refresh = re-put. Heartbeat cadence must be < bucket TTL with margin. Both values = **named constants** (the store module lives under `infrastructure/`, but anything landing under `src/factory/core/` falls under the `hardcoded_constants` gate — named + imported, never inlined).
+- **`parallel` mode preservation (review refinement F4):** the registry replaces `_refcount` only for `steer`/`queue`. For `parallel`, multiple jobs share one `(platform, bot, scope)` — coexistence strategy is an **open spec decision** (options: keep `TypingPublisher._refcount` alongside, registry-side counting); ref-counting semantics must survive either way.
+- **Layering + axial:** store module belongs in `src/factory/infrastructure/stores/` (sibling of `message_index_kv.py`; test anchor `tests/infrastructure/stores/test_message_index_kv.py`, AsyncMock pattern). `folder_size` gate: stores/ = 13 `.py` files today, `QG_FOLDER_MAX=15` (qg.conf) → 14 after addition, no exemption needed. Import-linter layers gate applies — the pool/hub side consumes the store via an injected port/protocol, never `import factory.infrastructure` from `core/`. A PR touching top-level `core/` + `infrastructure/` auto-triggers the `dev-core:axial-adr-review` label (CLAUDE.md) — expected, don't strip it.
+
+## Out of Scope
+
+- The `fresh | steer | queue` routing decision itself — #1797 (concurrency_router, single shared inbound stage).
+- Sub-jobs pub/sub migration (#1798), steer end-to-end (#1799), dashboard (#1800 — including its `kv.watch` ephemeral-consumer ACLs; the registry only has to be watchable).
+- Worker-driven typing rewiring (#1624) — reads this registry later; no typing code changes here.
+- JOBS dispatch stream (#1203) — different tier (at-least-once dispatch); zero file overlap expected (parallel track).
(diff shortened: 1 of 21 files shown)


MickaelV0:
## Code Review — PR #1863 (#1796 factory-active-jobs KV registry)

7 fresh domain agents (security · architect · backend · tester · devops · product-lead · axial). Single-chunk. The load-bearing findings were re-verified at source by the lead (noted inline). Verdict at bottom.

---

### 🔴 Blockers

**issue(blocking): unguarded refresh loop — a single transient NATS error silently and permanently kills liveness** · C 93% *(backend 95%, lead-verified at source)*
`infrastructure/stores/active_jobs_refresher.py:124` (`_loop`) + `:80` (`refresh_all`) + `infrastructure/stores/active_jobs_kv.py:254` (`refresh`)
`_loop` → `refresh_all` → `port.refresh(jid)` is a fully unguarded chain, and `refresh()`'s job-key `await kv.put(_job_key(...))` at L254 is **outside** the `except nats.errors.Error` block (only the *idx* re-put at L261 is guarded). A transient disconnect on that put propagates up, `_loop`'s `asyncio.Task` dies, `self._running` stays `True` (so nothing restarts it), and **all** tracked jobs stop being refreshed → they TTL-expire at 120 s → singleton index keys vanish → pools are wrongly freed while their jobs are still running. One NATS blip permanently disables the singleton guarantee this PR exists to enforce — with no log, no restart, no metric.
- Class: `missing-error-handling`
- Fix (recommended): wrap the body of `_loop` in `try/except Exception: log.error(..., exc_info=True); continue` so a transient failure costs one cycle, not the whole loop. Also fold the L254 job-key put into the existing `except nats.errors.Error` block in `refresh()`. (Per-item guard in `refresh_all` is a reasonable third layer.)

**issue(blocking): tautological put assertion — the job-key payload is never actually verified** · C 98% *(tester)*
`tests/infrastructure/stores/test_active_jobs_kv.py:73`
`kv.put.assert_awaited_once_with(_job_key("job-1"), kv.put.await_args.args[1])` reads the expected value *from the actual call* — it compares the call to itself. Serializing garbage bytes (or breaking `_entry_to_bytes`) would still pass. The write half of `open()` has no real coverage.
- Class: `test-tautology`
- Fix: build the expected bytes independently (`_entry_to_bytes(_entry())`) and assert against that, or `json.loads` the captured arg and assert field-by-field.

**issue(blocking): SC-5-mandated faked-TTL reap test is missing** · C 95% *(tester + product-lead)*
`tests/infrastructure/stores/test_active_jobs_kv.py`
Spec SC-5 explicitly requires a "faked-`ttl` reap test proves crash entries expire." No test simulates post-TTL expiry. `test_boot_ttl_passed_to_create_key_value` only checks the constant is forwarded — it does not prove expiry/handling. Given that the blocker above is *about* TTL-expiry behaviour, this coverage gap is doubly relevant.
- Fix: in `TestRefresh`, set `kv.get.side_effect = KeyNotFoundError` (post-expiry) and assert `refresh()` no-ops gracefully (it already returns on `KeyNotFoundError` at L246) — and add a test that a steer-mode entry whose idx is gone is handled.

---

### 🟠 Warnings

- **issue: `kv_safe_part` is non-injective → singleton-index key collisions** · C 82% *(security, lead-verified)* — `_kv_keys.py:16` does `_UNSAFE_RE.sub("_", value)`, so `"telegram:lyra:1"` and a literal `"telegram_lyra_1"` map to the **same** KV key. For the singleton index this means two distinct pools can contend on one idx key (false conflict, or one stealing the other's slot). **Pre-existing shared helper** (also used by `message_index_kv`), *not introduced by this PR* — but #1796's singleton *correctness* now depends on it being injective. → recommend a follow-up issue (injective encoding: `base64.urlsafe_b64encode` or a hash suffix) rather than blocking this PR. Class `missing-input-validation`.
- **issue: hub cold-restart leaves orphan KV entries unrefreshed until TTL** · C 90% *(architect)* — `_jobs` starts empty on boot; a crash+restart within the 120 s window leaves prior KV entries live but never refreshed by the new hub → phantom "open" jobs for ≤120 s. Semantically intended (crash ⇒ closed, TTL = safety net) but the invariant is undocumented. → add one sentence to the `RegistryCoordinator` docstring + `job-model.md` §Liveness.

---

### 🟡 Suggestions / Nitpicks

- **suggestion(non-blocking): write job key *after* the CAS, not before** · C 70% *(backend 92%, lead-downgraded)* — `open()` writes `_job_key` (L209) before the singleton `kv.create` (L224); a conflicting opener leaves an orphan job key. **Verified benign:** job keys are only ever read via the index→winner indirection (`get_by_pool`), the index never points to the loser, the coordinator never records a conflicted job, and the orphan TTL-reaps in ≤120 s — so it is unreachable, not a correctness bug. Reversing the order (CAS first) removes the orphan for cleanliness.
- **suggestion(non-blocking): promote `ACTIVE_JOBS_REFRESH` to the port module** · C 85% *(architect + axial, independent)* — `RegistryCoordinator`'s default cadence imports a constant from the concrete KV adapter (`active_jobs_kv`). No layer violation (both infra), but the coordinator's default is implicitly KV-specific. Move the constant to `core/ports/active_jobs.py`.
- **suggestion(non-blocking): log when `winner_id` falls back to `"<unknown>"`** · C 80% *(backend)* — the TOCTOU between CAS-reject and the winner `get` (L226-233) yields an opaque conflict; a DEBUG line distinguishes a genuine conflict from a TTL-expire race.
- **nitpick: add `@pytest.mark.asyncio` (or class-level `pytestmark`) to `test_active_jobs_kv.py`** · C 40% *(tester 90%, lead-verified DOWN)* — the agent flagged these as silent no-ops. **Not a coverage gap:** `pyproject.toml` sets `asyncio_mode = "auto"`, so bare `async def` tests *do* run (the 2014-pass count is valid). Pure style inconsistency with `test_active_jobs_refresher.py`, which decorates explicitly.
- **thought: `RegistryConflictError` embeds `existing_job_id`** · C 70% *(security)* — fine at definition; verify every catch site wraps it in `SanitizedError`/logs only and never forwards `str(exc)` to a bus-bound field. None found in scope.
- **nitpick: stale `"TODO T4"` notes in `v3-pre-grant-group.json`** *(devops)* — notes-field only, not CI-gated (the equivalence gate compares grant *sets*). Fold into a fixture-notes cleanup chore.

---

### 🟢 Praise

- **The singleton CAS conflict paths are genuinely tested, not mock-tautological** (tester, 97%) — `…_different_job_conflicts`, `…_idempotent_reopen`, `…_winner_vanished_still_conflicts` each force `KeyWrongLastSequenceError`; deleting `kv.create` fails all three, and `kv.update.assert_not_awaited()` guards against any revert to the old steal-a-busy-pool path. This is the core invariant and it is well-proven.
- **create-only CAS is correct** (backend, lead-verified) — `kv.create` → `KeyWrongLastSequenceError` → `RegistryConflictError`; same-`job_id` re-open is idempotent (L233); `direct=False` correctly maps to `allow_direct=False`; all keys route through `kv_safe_part`; str_exc_bus_bound clean.
- **Port is a clean driven-port** (architect 100% + axial) — `core/ports/active_jobs.py` has zero infrastructure imports (stdlib + Protocol + frozen dataclass + domain exception); nats-py is `TYPE_CHECKING`-gated in the adapter.
- **ADR-079 lifecycle order respected** (architect 100%) — bucket ensured → `connect()` → coordinator `start()` → `announce_hub_ready()`, fail-fast; `stop()` in shutdown. `refresh_all` snapshots via `list(self._jobs)` (mutation-safe).
- **Forward-compatible substrate** (architect 100%) — `worker_loc`, `_by_loc`, `open`/`close`, dormant `on_heartbeat` all already shaped so #1797 wires call-sites with zero port/coordinator churn.
- **ACL fan-out complete & consistent** (devops + security) — `$KV.factory-active-jobs.>` present in all 4 artifacts; least-privilege; hub's existing `$JS.API.>` covers bucket provisioning; `messaging.md` KV-plane row accurate; no monitor grant.
- **Axial: fully aligned with primary axis `stage`** (axial-adr-review) — +1 stage module composing shared primitives, not an N×M copy; `class.*Client.*Client` 0 hits.
- **Spec: 4/9 met (SC-1/2/3/5), 5 deferred-by-design, 0 silent drops** (product-lead) — all deferrals (live open/close → #1797, heartbeat dormant, auth.conf → CI) genuinely out of this spec's criteria; `kv.create`-only CAS is spec-compatible with the `kv.update(last_revision)` wording (create-over-tombstone is transparent in nats-py).

---

One genuine runtime defect gates merge: the unguarded refresh loop silently and permanently disables the liveness mechanism on any transient NATS error — verified at source, and it defeats the singleton guarantee this PR delivers. Plus a tautological write-assertion and the SC-5-mandated TTL-reap test that's absent. All three are tightly scoped (≤15 lines + 2 tests). The substrate itself — CAS correctness, port hygiene, lifecycle, ACL fan-out, axial alignment — is sound and well-tested. The `kv_safe_part` non-injectivity is a real but pre-existing shared-helper concern → handle via follow-up issue, not here.

MickaelV0:
## Review Fixes Applied — `79414d33`

| Blocker | Resolution |
|---|---|
| 🔴 unguarded refresh loop → liveness silently dies | `_loop` and `refresh_all` now guard every sweep (log+skip, never propagate); `KvActiveJobsStore.refresh` folds the previously-unguarded job-key `put` into the `nats.errors.Error` try block. A transient NATS error costs one cycle, not the whole loop. |
| 🔴 tautological put assertion | `test_open_writes_job_and_idx_steer` asserts the exact payload via independent `_entry_to_bytes(entry)` re-serialization, not `args[1]` vs itself. |
| 🔴 SC-5 faked-TTL reap test absent | `test_open_reclaims_pool_after_ttl_reap` — create(win)→create(conflict)→reap→create(win) lifecycle proves an expired pool slot is reclaimable. |
| + regression guard | `test_refresh_all_one_failure_does_not_abort_sweep` — one job raising must not stop the others. |

Verify: 39/39 store+refresher tests green · ruff clean · str_exc_bus_bound OK · test_sleep OK.

Follow-up (not in this PR): `kv_safe_part` non-injectivity (pre-existing shared helper, also used by `message_index_kv`) → recommend a dedicated issue for an injective KV-key encoding.

PR #1869, opened 2026-06-12
fix(deploy): wire factory-omp install + bake omp binary into image

MickaelV0:
## Summary
- **factory-omp was unrunnable as shipped (#1858):** the unit pointed `Image=` at the `factory-omp-base` scratch carrier (single file, no factory CLI), while `_rpc_bridge.py` expects `/opt/omp/omp` inside the factory image. The agent-runtime stage now bakes the binary via `COPY --from` the pinned carrier (sha256 re-verified at startup), and the unit uses `ghcr.io/roxabi/factory:staging` (clipool pattern).
- **factory-omp was never installed:** `[component.omp]` existed in `deploy/quadlet.toml` but the three hand-enumerated deploy lists didn't know it — added to `make quadlet-install`, the converge.sh structural restart fan-out, and the quadlet-install-verify.sh `UNITS` array (try-restart-or-start covers first activation).
- **New gate `tools/check_quadlet_manifest_install.sh`** (stack.yml + ci.yml step): every manifest container must be wired into all three paths (factory-nats exempt from the restart list — own `_restart_nats` path). Falsified both directions: removing the Makefile line or the converge entry makes the gate fail.

## Deploy note (M₁)
First converge after merge may start factory-omp against the previous `factory:staging` digest (no `/opt/omp/omp` yet) → short crash-loop until `podman-auto-update` pulls the post-merge image at the next 5-min tick. Self-healing; no operator action required.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1867: fix(deploy): wire factory-omp into quadlet-install + converge — PR #1858 deploy gap | OPEN |
| Implementation | 2 commits on `feat/1867-factory-omp-deploy-wiring` | Complete |
| Verification | Lint ✅ ruff format ✅ quadlet-lint ✅ gates ✅ (secrets_drift, volumes_table, arch_snapshot, manifest_install + 2 falsifications) — pyright/pytest n/a (no .py touched) | Passed |

## Test Plan
- [ ] CI green incl. new `Check quadlet manifest install wiring` step
- [ ] Post-merge on M₁: `ls ~/.config/containers/systemd/ | grep factory-omp` after quadlet-sync tick
- [ ] `systemctl --user is-active factory-omp` once auto-update pulls the post-merge `factory:staging`

Fixes #1867

---
Generated with [Claude Code](https://claude.com/claude-code) via `/pr`

Issue #1867, opened 2026-06-12
fix(deploy): wire factory-omp into quadlet-install + converge — PR #1858 deploy gap


MickaelV0:
PR #1858 (#1812) shipped `deploy/quadlet/factory-omp.container` + `[component.omp]` in `deploy/quadlet.toml` + both secrets (present on M₁), but the unit is NEVER installed: converge reports "Converge complete" while `~/.config/containers/systemd/` has no factory-omp.container. Root cause: deploy is not manifest-driven — three hand-enumerated lists were not updated.

Tasks:
- [ ] `Makefile` `quadlet-install`: add `cp deploy/quadlet/factory-omp.container` line
- [ ] `deploy/converge.sh`: add `factory-omp` to the structural-drift client restart list
- [ ] `deploy/quadlet-install-verify.sh`: expected units 8 → 9
- [ ] New gate `tools/check_quadlet_manifest_install.sh`: every `container =` in `deploy/quadlet.toml` must be installed by the Makefile `quadlet-install` recipe (cp or render `--dest`) — prevents recurrence; register in `.claude/stack.yml` quality_gates
- [ ] Post-merge operator step: one-time `systemctl --user daemon-reload && systemctl --user start factory-omp.service` on M₁ (converge only restarts already-active units)

Context: relates to #1812 (closed). #1813 (runtime selection) needs factory-omp actually running on M₁. Discovered during 2026-06-12 M₁ audit.

MickaelV0:
**Scope expansion (discovered during implement):** the unit shipped by #1858 cannot run at all — `Image=ghcr.io/roxabi/factory-omp-base:15.10.8` is the **scratch binary-carrier** ("never runs directly", per `deploy/omp-base/README.md`), yet `Exec=factory adapter omp` needs the factory CLI. The spec (1812-omp-worker-spec.mdx) baked this in. `_rpc_bridge.py` expects the binary at `/opt/omp/omp` **inside the factory image**, which today has no `COPY --from` of the carrier.

Wiring the install without fixing the image would crash-loop the unit and fail every converge on M₁ (5-min alert spam). Added to scope, same PR:

- [ ] `Dockerfile` agent-runtime stage: `COPY --from=ghcr.io/roxabi/factory-omp-base:15.10.8 /opt/omp/omp /opt/omp/omp`
- [ ] `deploy/quadlet/factory-omp.container`: `Image=` → `ghcr.io/roxabi/factory:staging` (clipool pattern)
[...]

Issue #1807, opened 2026-06-09
spike(harness): evaluate Pi / oh-my-pi as lyra-harness runtime (vs custom thin loop)


MickaelV0:
## Goal
De-risk swapping the planned ~200-line **custom thin loop** (#1490 "Runtime" decision + child "Custom loop runtime spike") for an OSS agent runtime. The runtime eval (`artifacts/analyses/agent-runtime-goose-vs-pi.md`) recommends the **Pi-family** (oh-my-pi as runtime + earendil/pi as stable anchor) over **Goose**. This spike validates the integration linchpin before any commit.

## Why now
- 2026-06-15 Anthropic split-billing cliff → need an OSS, provider-agnostic runtime routed via the M1 LiteLLM proxy (llmcli), replacing `claude -p`/clipool.
- Eval verdict: **Pi-family wins.** Goose eliminated — tool-use is buried in `Message.content` (fragile reconstruction of `ToolUseLlmEvent`) and its only programmatic surface is the `_goose/unstable/*` ACP (Python SDK is a ping/pong stub).
- oh-my-pi natively provides what the job model expected to **build**: `steer()` queue (Shape D #1799), `task` sub-agent tool (sub-jobs/composite_depth #1798), MCP + host-tool JSON-Schema (#493 ToolHandler), a **named turnkey LiteLLM provider** (default baseUrl `localhost:4000/v1` = llmcli), and first-class `tool_execution_start/update/end` events (1:1 with our `Text/ToolUse/Result`).
- **Linchpin (verified):** `external_repos/oh-my-pi/python/omp-rpc` = typed Python client (alpha, MIT) wrapping `omp --mode rpc` NDJSON, with host-tool helpers exposing custom tools via JSON Schema → the harness worker stays **Python `NatsAdapterBase`**, no TypeScript worker needed.

## Spike tasks (timebox ~2-3 days)
1. Read `external_repos/oh-my-pi/python/omp-rpc/` API + `docs/rpc.md`; confirm a stable session lifecycle (start / prompt / steer / abort / events / host-tools).
2. Throwaway Python stub driving one `omp --mode rpc` session; run a text turn + a tool turn.
3. Map omp events → `TextLlmEvent` / `ToolUseLlmEvent` / `ResultLlmEvent`; fixture-compare vs the current clipool chunk shape.
4. Bridge one existing `ToolHandler` (#493) via omp host-tool JSON-Schema; round-trip execute.
[...]

Issue #1812, opened 2026-06-10
feat(harness): OmpWorker — omp_rpc NATS runtime backend (omp --mode rpc)


MickaelV0:
**Origin:** #1807 spike (GREEN — 4/4 live probes). Add `omp_rpc` as a harness runtime backend **alongside** clipool/`claude -p` (runtime is pluggable; clipool stays).

Python `NatsAdapterBase` driving `omp --mode rpc` (subprocess-per-`RpcClient`, `executable=` prebuilt binary):
- Dispatch `factory.jobs.omp` → `factory.job.<id>.{progress,result,opened,closed,steer}`.
- Events: `on_message_update`(text_delta) → `TextLlmEvent`; `on_tool_execution_start` → `ToolUseLlmEvent{tool_name,tool_id,input}`; `on_agent_end`+`get_state()` → `ResultLlmEvent{session_id,is_error}` — live-verified 1:1.
- Host-tool JSON-Schema bridge = #493 ToolHandler (sync `execute` → `run_in_executor`).
- `steer()` (between-tool-calls) → bridge NATS `factory.job.<id>.steer` (#1799). `abort()` / `new_session()` lifecycle.

Caveats (from #1807): baseUrl hard-coded :4000 (#1811), needs base image (#1810), `switch_cwd`/resume gaps moot under stateless Shape-B.

Out of scope: runtime selection (separate). Parent #1490; refs #1044 #1792 #1799 #493 #1778.

---

### ACL / contracts fan-out (do FIRST — known red-walk pattern)

- [ ] Add `omp-worker` identity to `deploy/nats/acl-matrix.json`
- [ ] Enumerate ALL new subjects up front in `packages/roxabi-contracts` (jobs subjects): `factory.jobs.omp`, `factory.job.<id>.{progress,result,opened,closed,steer}`
- [ ] Regenerate the 4 derived artifacts: auth.conf · 706 ACL spec · v3-current fixture · `CURRENT.generated.md`
- [ ] Verify `secrets_drift` + `architecture_snapshot` gates locally before PR

### End-to-end SHA-pin guard (complements #1810's image-layer pin)

- [ ] Integration test: OmpWorker refuses to start when the `omp` binary digest ≠ pinned SHA — bump validation must exercise the full worker protocol, not just the image layer

### Security design constraints (from PR #1823 audit, 2026-06-10)

[...]

PR #1872, opened 2026-06-13
fix(omp): runtime closure — workspace mount + omp_rpc image pin

### LINKED ISSUE #1871: fix(omp): runtime closure — workspace probe target + omp_rpc image dep (factory-omp crash-loop)
Post-#1869 deploy verification on M₁ (2026-06-13 ~00:55 CEST): `factory-omp.service` installs and starts correctly (unit ✓, `/opt/omp/omp` baked ✓, secrets ✓, ACL ✓) but **crash-loops** (restart counter >175, exit 1 every ~13s, ~22% CPU continuous).

**Root cause 1 — ownership probe target absent:** `_bootstrap_omp_standalone` (`src/factory/bootstrap/standalone/worker_standalone.py:142`) calls `run_git_ownership_probe()` with no args → default `/home/factory/projects/roxabi-factory` → `sys.exit(1)` ("target directory does not exist"). The clipool unit mounts `%h/projects:/home/factory/projects:z`; `factory-omp.container` mounts only `deploy/omp` config. Options: (a) mount the projects workspace like clipool (probe purpose = verify keep-id uid mapping on the bind-mounted workspace — omp is a coding agent and will need it for jobs), (b) `Environment=FACTORY_OWNERSHIP_PROBE_PATH=` override (defeats the probe), (c) skip probe for omp (loses the uid-mapping fail-fast).

**Root cause 2 — omp_rpc not in the image:** the bootstrap docstring says "omp_rpc is a container image dep (absent from pyproject.toml)" (deliberate — alpha + fork risk, pin by SHA per #1807/#1810), but nothing installs it: the Dockerfile builds from `uv sync --frozen --no-dev` only, and the `factory-omp-base` carrier ships only the binary. `import omp_rpc` (deferred, `_rpc_bridge.py`) will fail even after the probe is fixed. Needs a pinned install step in the agent-runtime Dockerfile stage (source: oh-my-pi `python/omp-rpc` @ pinned SHA — see #1810/#1811 artifacts) or a revisit of the "image dep" decision.

**Interim state:** `systemctl --user stop factory-omp` applied on M₁ (2026-06-13). NOTE: the next **structural** converge restarts it → crash-loop resumes silently (converge does not detect flapping — restart succeeds, unit dies 13s later). CI does not catch either gap (omp tests mock omp_rpc; no boot-in-container test).

Acceptance:
- [ ] `factory-omp.service` active >10 min on M₁, restart counter stable
- [ ] probe passes (or is explicitly scoped out for omp with rationale)
- [ ] `podman exec factory-omp python -c "import omp_rpc"` OK
- [ ] consider a smoke gate: container boot test or import check in CI (gap class: "deps declared as image-deps but never installed")

Relates: #1812 (closed — shipped the code), #1813 (runtime selection — needs this running), #1867/#1869 (deploy wiring — done). Discovered during the 2026-06-12/13 M₁ audit.

MickaelV0:
## Summary

Closes the two root causes behind the `factory-omp` crash-loop (restart >175, exit 1 every ~13s) found in the 2026-06-13 M₁ post-#1869 audit. Deploy wiring (#1867/#1869) was correct; the unit failed at **runtime** for two independent reasons:

- **RC1 — ownership-probe target absent.** `_bootstrap_omp_standalone` calls `run_git_ownership_probe()` with no args → defaults to `/home/factory/projects/roxabi-factory`, which the omp unit never mounted. **Fix:** add `Volume=%h/projects:/home/factory/projects:z` to `factory-omp.container`, mirroring `factory-clipool` exactly. `UserNS=keep-id:uid=1500,gid=1500` (self-declared on the standalone container; clipool inherits the equivalent from `factory-gh.pod`) maps host files to uid 1500 so the probe's ownership check passes.
- **RC2 — `omp_rpc` not in the image.** The deferred `import omp_rpc` (`_rpc_bridge.py`) had no installer: `uv sync --frozen --no-dev` covers pyproject only, and `omp_rpc` is deliberately kept **out** of `uv.lock` (alpha lib, pin-by-SHA — #1807/#1810). **Fix:** in the `agent-runtime` stage, `uv pip install` `omp_rpc` pinned at commit `4b5200a1` (== `v15.10.8`, the same omp version the binary carrier pins → matching RPC protocol), plus a build-time `RUN python -c "import omp_rpc"` smoke that fails the image build if the dep is mis-installed (closes the "declared-but-never-importable" gap class).

### Scope notes

- **e2e validation deferred to llmCLI#114.** A full "omp runs a job end-to-end" check needs a live LLM; M₁'s LiteLLM upstreams are currently dead (llmCLI#114). This PR lands the structural runtime fix so the next converge stops resuming the crash-loop.
- **Image build is post-merge.** `publish.yml` builds the image on **push to `staging`** (and release tags), not on PRs — so the `import omp_rpc` smoke runs on staging-publish after merge, not in this PR's CI. The exact pinned `uv pip install` + import was **pre-validated locally** against the real remote (resolved → built → installed `omp-rpc==0.1.0`, zero third-party deps).
- **Version-bump lockstep:** an omp version bump must update three pins together — `deploy/omp-base/Containerfile` (`OMP_VERSION`/`OMP_SHA256`), `_rpc_bridge.py` (`_PINNED_SHA256`), and the commit pin in `Dockerfile`. Documented inline.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1871: runtime closure (factory-omp crash-loop) | OPEN |
| Implementation | 1 commit on `feat/1871-omp-runtime-closure` | Complete |
| Verification | Gates ✅ (volumes_table, secrets_drift, quadlet_manifest_install) · install smoke ✅ pre-validated · no Python runtime code touched | Passed |

## Test Plan

- [ ] On M₁ after a converge: `systemctl --user status factory-omp` → active, restart counter stable >10 min.
- [ ] `podman exec factory-omp /app/.venv/bin/python -c "import omp_rpc"` → exits 0.
- [ ] `podman exec factory-omp ls /home/factory/projects/roxabi-factory` → probe target present, owned by uid 1500.
- [ ] Staging `publish` job green (the build-time `import omp_rpc` smoke executes there).
- [ ] Full job e2e — tracked under llmCLI#114 once M₁ LiteLLM upstreams recover.

Fixes #1871

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,7 @@ USER root
 # to dial the dispenser Unix socket. Smallest dep that handles UNIX-CONNECT cleanly;
 # BSD nc -U fallback in the shims is for hosts where socat is unavailable.
 RUN apt-get update \
- && apt-get install -y --no-install-recommends socat \
+ && apt-get install -y --no-install-recommends socat git \
  && rm -rf /var/lib/apt/lists/*
 
 # UID 1500 pinned per ADR-053 (Quadlet container UID stability)
@@ -102,6 +102,22 @@ RUN test -x /usr/bin/gh \
  && chmod 0755 /opt/factory-gh/gh \
  || true
 ENV FACTORY_GH_BIN=/opt/factory-gh/gh
+
+# ── omp_rpc Python package (#1871) ─────────────────────────────────────────────
+# omp_rpc is deliberately kept OUT of uv.lock (alpha lib, pin-by-SHA pattern —
+# #1807/#1810). Install at image build time from the pinned commit that matches
+# OMP_VERSION=v15.10.8 / factory-omp-base:15.10.8. The commit SHA is locked here;
+# a version bump must update deploy/omp-base/Containerfile OMP_VERSION+OMP_SHA256,
+# src/factory/adapters/omp/_rpc_bridge.py _PINNED_SHA256, AND this pin — in lockstep.
+# uv is not present in agent-runtime (only in builder); bring the static binary from
+# the official astral-sh image so we can pip-install into /app/.venv without touching
+# the project lockfile.
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /usr/local/bin/uv
+RUN uv pip install --python /app/.venv/bin/python \
+      "git+https://github.com/can1357/oh-my-pi@4b5200a163060c057f1d886df15d7ef083a54a62#subdirectory=python/omp-rpc"
+# Build-time import smoke — fails the image build if omp_rpc is mis-installed or
+# the subdirectory path changes upstream. Closes the "declared but never importable" gap.
+RUN /app/.venv/bin/python -c "import omp_rpc"
 
 WORKDIR /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,9 @@ USER root
 # socat — required by the gh_token shim scripts (git-credential-factory-gh + factory-gh)
 # to dial the dispenser Unix socket. Smallest dep that handles UNIX-CONNECT cleanly;
 # BSD nc -U fallback in the shims is for hosts where socat is unavailable.
+# git — two consumers: (1) runtime — the omp ownership probe shells out to
+# `git rev-parse HEAD` (git_ownership_probe.py); (2) build — the omp_rpc `git+URL`
+# install below. Do NOT drop git while either consumer exists.
 RUN apt-get update \
  && apt-get install -y --no-install-recommends socat git \
  && rm -rf /var/lib/apt/lists/*
--- a/deploy/quadlet/factory-omp.container
+++ b/deploy/quadlet/factory-omp.container
@@ -22,6 +22,10 @@ Environment=LITELLM_API_KEY_FILE=/run/secrets/factory-litellm-key.tok
 Tmpfs=/tmp
 Volume=%h/projects/roxabi-factory/deploy/omp:/home/factory/.config/omp-pi:ro,Z
 Environment=PI_CODING_AGENT_DIR=/home/factory/.config/omp-pi
+# Workspace — mirrors host ~/projects so the git ownership probe finds its target
+# (run_git_ownership_probe defaults to /home/factory/projects/roxabi-factory).
+# Same mount as factory-clipool; UserNS=keep-id:uid=1500,gid=1500 ensures correct ownership.
+Volume=%h/projects:/home/factory/projects:z
 Label=io.containers.autoupdate=registry
 
 [Service]


MickaelV0:
## Code Review — PR #1872 (`fix(omp): runtime closure`)

Two fresh reviewers (security-auditor + devops), single-chunk (2 files). Findings adjudicated against the actual omp worker behavior — the two agents conflicted on the mount flag; resolved below with code evidence.

### Adjudication note — the mount-flag conflict
`omp_worker.py` / `_rpc_bridge.py` never reference `/home/factory/projects`. The **only consumer of the new mount today is the ownership probe** (`git rev-parse HEAD`, `cwd=target`, read-only). omp's job-workspace writing is unwired (#1812/#1813). Therefore:
- `:ro` **is sufficient** for the current probe-only use → least privilege.
- `:Z` (private relabel) was **rejected** — clipool mounts the *same* host `~/projects`; a private relabel would break clipool's shared access. Keep lowercase `:z`.

### ⚠️ Decision (least-privilege) — mount access mode
**`issue` (devops, C88, adjudicated):** the workspace bind is read-write, but the only current reader is the probe. An autonomous LLM coding agent gets write to the *entire* host `~/projects` (all repos) it doesn't yet use.
- Recommended: `Volume=%h/projects:/home/factory/projects:ro,z` — read-only now; widen to writable when #1812/#1813 wires job execution (that PR touches this unit anyway).
- Counter (security-auditor): clipool already grants identical write access → no *net-new* risk; mirroring clipool (writable) is defensible. → **owner decides** (see PR thread).

### Apply before merge
- **`suggestion` (devops, C72):** the new `apt … git` comment cites only the runtime probe; `git` is *also* needed at build for the `git+URL` install. Broaden the comment so a future maintainer doesn't drop `git` assuming a single use. (`git_ownership_probe.py:81` confirms a runtime `git` dependency — so the dep itself is correct.)

### Accepted / no action (with rationale)
- **install runs as `USER root` (security-auditor, C62):** build-time only; lands in `factory:factory` `.venv` → unprivileged at runtime; consistent with the existing root apt/chmod blocks in this stage. Accept.
- **no `--require-hashes` (security-auditor, C65):** not applicable to VCS installs; the immutable commit-SHA pin + the runtime `_PINNED_SHA256` digest gate on the binary are two independent controls. Accept.
- **layer-cache placement (devops, C85):** valid in principle (mirror the pre-COPY omp-binary placement), but the cache value for a tiny zero-dep pure-python git fetch is marginal vs the 175 MB binary, and the `--target`/`PYTHONPATH` split adds complexity. Declined for this PR; optional polish.
- **smoke placement (devops, C50):** current pre-`USER` placement is correct (root reads `factory:factory` `.venv`). No action.

### Follow-ups (out of scope — filing issues)
- **3-way pin lockstep has no CI gate (devops, C65):** omp-base `OMP_VERSION` / Dockerfile commit-SHA / `_rpc_bridge.py _PINNED_SHA256` are coupled only by docs. Matches this repo's gate-everything culture → new `check_omp_pin_lockstep.sh`.
- **`uv:latest` unpinned (devops, C70):** matches the existing builder convention (also `uv:latest`); pin both (or by digest) in the same follow-up.

### Praise
- Layered integrity: build-time `sha256sum -c` (omp-base) + runtime `_verify_digest()` + commit-pinned source = three independent checkpoints.
- `ReadOnly=true` + `DropCapability=all` + `NoNewPrivileges=true` preserved; no hardening directive weakened.

---
**Verdict: Request changes** — 1 least-privilege decision (mount mode, owner call) + 1 comment-accuracy fix. No correctness or exploitable-security blocker; the runtime fix itself is sound and the omp_rpc install was pre-validated against the pinned commit (zero third-party deps).

Issue #1893, opened 2026-06-15
refactor(stores): rename cli_session_id → provider_session_id (backend-agnostic session token)


MickaelV0:
**Origin:** #1813 (deferred from V2 / Slice 2). NOT a drift-undo.

V2 reused the existing `pool_sessions.cli_session_id` column + `TurnStoreSessionMixin.get_cli_session/_set_cli_session` as a generic opaque session-resume token (the omp `.jsonl` session_file path) — no migration was done, to keep the slice tight.

**Scope (cosmetic/structural rename):** rename `cli_session_id` → `provider_session_id` across the column + the ~47 callsites (TurnStore, TurnWriter, turn_store_protocol, session_commands, clipool_worker, cli_pool_session, …) with a DB migration for `pool_sessions`.

**Also evaluate (later):** if #1813 V5 (per-job override) lands a 2nd concurrent backend per agent, the single token column collides on mid-conversation backend switch → per-backend token separation. At the 3rd-backend three-strikes threshold, extract a shared `SessionBridge` primitive (see #1813 spec 'Known constraints').

PR #1894, opened 2026-06-15
feat(omp): OmpPool + native-session memory (#1813 slice V2)

### LINKED ISSUE #1813: feat(harness): runtime selection — clipool/claude-p ⟷ omp_rpc per agent/job
**Origin:** #1807 spike (GREEN) + decision 2026-06-10 — **clipool is NOT retired**; runtime is a per-agent / per-job **choice**.

Config-driven selection of the runtime backend (clipool/`claude -p` ⟷ `omp_rpc`); both coexist. Surfaces:
- Agent/job config (C = `config.db`) carries a `runtime` selector.
- Hub dispatch routes to `factory.jobs.clipool` vs `factory.jobs.omp` accordingly.
- Default + override semantics; parity checks before flipping any default.

Blocked-by OmpWorker (#1812). Parent #1490; ref #1807 result doc.

MickaelV0:
## Summary
Slice **V2** of the runtime-selection epic (#1813).

- **Native-session memory:** omp agents now remember context across turns via a **hub-driven** session token (mirror of clipool's `cli_session_id`). `OmpRpcDriver` gains real `SessionAware` methods (`_turn_store`/`_pending_resume`): `queue_resume` resolves the token, `complete()` injects `provider_session_id` into the `JobEnvelope` and persists the backhauled `session_file`.
- **`OmpPool`** (worker-side, CliPool mirror): warm `RpcClient` per `pool_id`, LRU evict at `OMP_POOL_CAP` (env, default 4), cold-start `switch_session(session_file)` or `new_session()`+`get_state()`. Worker stays a dumb executor.
- **Dispatch unified (Axis A):** standalone `SessionAware` / `WorkspaceAware` protocols replace the four `_cli_pool`/`_cli_nats_driver` branch sites in `SimpleAgent` with `isinstance` resolution — collapses the pre-existing 2-branch duplication, absorbs omp polymorphically, and fixes a latent `reset_backend()` asymmetry (the nats path was silently dropped). omp is excluded from `WorkspaceAware` (no `switch_cwd` faking).
- **Token reuse (Axis B):** omp reuses the existing `pool_sessions.cli_session_id` column — **no migration**. Cosmetic rename deferred to **#1893**.
- **Deploy:** session dir persisted on a named volume `factory-omp-sessions` (clipool-parity → cold-start resume survives a worker restart).

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #1813: runtime selection — clipool ⟷ omp_rpc | OPEN (epic, slice V2) |
| Analysis | [1813-runtime-selection-analysis.mdx](artifacts/analyses/1813-runtime-selection-analysis.mdx) | Present |
| Spec | [1813-runtime-selection-spec.mdx](artifacts/specs/1813-runtime-selection-spec.mdx) | Present |
| Plan | [1813-runtime-selection-v2-plan.mdx](artifacts/plans/1813-runtime-selection-v2-plan.mdx) | Present |
| Implementation | 6 commits on `feat/1813-runtime-selection-v2` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (5861 passed; 9 new test files) | Passed |

## Test Plan
- [ ] omp agent recalls a prior-turn fact via its warm pool client
- [ ] a 2nd conversation (distinct `session_id`) does NOT see the 1st's history
- [ ] clipool + nats backends unchanged (no regression — existing suite green)
- [ ] M₁: `INTEGRATION=1` omp session-resume smoke (manual — needs live binary)

## SC → Test Matrix

| SC | Test(s) | Status |
|----|---------|--------|
| omp remembers context across turns | `test_omp_rpc_session :: queue_resume/complete inject+persist` | ✓ proven |
| no cross-conversation bleed | `test_omp_rpc_session :: test_cross_conversation_isolation` | ✓ proven |
| OmpPool warm/cold-start + LRU | `test_omp_pool` (5 cases) | ✓ proven |
| SessionAware/WorkspaceAware conformance | `test_session_protocols` | ✓ proven |
| reset_backend nats asymmetry fixed | `test_simple_agent :: test_reset_backend_routes_through_nats_driver` | ✓ proven |
| worker stays dumb executor | `test_omp_worker_pool` | ✓ ran green |
| clipool/nats no-regression | `test_simple_agent` (t8n/t9an/t9bn + existing) | ✓ ran green |
| wire-compat (old envelope validates) | `test_omp_payload_v2` | ✓ ran green |
| integration cold-start/resume/evict | `test_omp_session_resume_integration` | ⚠ INTEGRATION=1 (M₁-manual) |
| streaming / concurrency-wallclock / per-job override / parity runbook | — | deferred → V4 / V3 / V5 / V6 |

"✓ proven" = test runs green AND fails when the implementation is stashed (falsification gate executed).

## Notes
- **Multi-slice epic** — this is **slice V2**. V3 (OmpPool concurrency), V4 (streaming), V5 (per-job override), V6 (parity runbook) remain. Uses `Refs` (not `Fixes`) so the epic stays open.

Refs #1813 (slice V2) · follow-up #1893

---

### DIFF
--- a/artifacts/plans/1813-runtime-selection-v2-plan.mdx
+++ b/artifacts/plans/1813-runtime-selection-v2-plan.mdx
@@ -1,0 +1,790 @@
+---
+title: "Plan: runtime selection — OmpPool + native-session memory (slice V2)"
+issue: 1813
+spec: artifacts/specs/1813-runtime-selection-spec.mdx
+complexity: 8/10
+tier: F-full
+slice: V2
+generated: 2026-06-15
+---
+
+## Summary
+
+Slice V2 wires hub-driven native-session memory into the omp runtime by mirroring the exact pattern LlmClient uses for clipool: a `_turn_store`-backed `queue_resume` / `complete` backhaul loop that persists and resolves the `.jsonl` session_file as `provider_session_id` in `pool_sessions`. Axis A (dispatch unification) introduces two standalone `runtime_checkable` Protocols (`SessionAware`, `WorkspaceAware`) in `core/ports/llm.py`, replacing four separate `_cli_pool`/`_cli_nats_driver` branch sites in `simple_agent.py` with two resolved-once attributes. Axis B reuses the existing `cli_session_id` column and `TurnStoreSessionMixin` as-is (no migration), deferring the cosmetic rename to a filed follow-up. The worker gains `OmpPool` (warm `RpcClient` per `pool_id`, LRU-capped, `switch_session`/`new_session`+`get_state` cold-start), wired into `OmpWorker.handle()`; `OmpRpcDriver` gains real SessionAware methods mirroring LlmClient — NOT stubs.
+
+---
+
+## Architecture
+
+### Data-flow: hub-driven session round-trip
+
+```mermaid
+flowchart TD
+    CFG[config.db\nbackend=omp-rpc] --> HUB[hub dispatch\nresolve provider_session_id\nfrom pool_sessions]
+    HUB --> DRV[OmpRpcDriver.complete\nqueue_resume: pop _pending_resume\ninject provider_session_id into payload]
+    DRV -->|JobEnvelope\npool_id + provider_session_id| Q[factory.jobs.omp]
+    Q --> WRK[OmpWorker.handle\nextract pool_id + provider_session_id]
+    WRK --> POOL[OmpPool.acquire\npool_id + session_file]
+    POOL -->|warm hit| WC[warm RpcClient\nin-process session]
+    POOL -->|cold-start| CS{session_file?}
+    CS -->|yes| SW[switch_session\nresume existing context]
+    CS -->|no| NS[new_session\n+ get_state -> session_file]
+    WC --> RES[JobResult.data\nsession_file backhaul]
+    SW --> RES
+    NS --> RES
+    RES --> DRV2[OmpRpcDriver.complete\nread session_file from JobResult.data\npersist via _turn_store.set_cli_session\nkeyed by linked session_id]
+    DRV2 --> STORE[(pool_sessions\ncli_session_id / provider_session_id)]
+    STORE -.next-turn resolve.-> HUB
+```
+
+### File x function map
+
+```mermaid
+flowchart LR
+    subgraph core/ports/llm.py
+        SA[SessionAware Protocol\nlink_lyra_session\nreset\nqueue_resume]
+        WA[WorkspaceAware Protocol\nswitch_cwd]
+        ALL[__all__ export]
+    end
+
+    subgraph core/simple_agent.py
+        SB[_session_backend: SessionAware|None\n_workspace_backend: WorkspaceAware|None\nresolved in __init__ via isinstance]
+        RB[reset_backend\n_maybe_register_reset\n_maybe_register_resume\nlink call in process]
+    end
+
+    subgraph llm/drivers/omp_rpc.py
+        ORD[OmpRpcDriver\n_turn_store injected\n_pending_resume dict\n_lyra_sessions dict\nqueue_resume / reset / link_lyra_session\ncomplete: inject+backhaul session_file]
+    end
+
+    subgraph adapters/omp/omp_pool.py
+        OP[OmpPool\nclients dict pool_id→RpcClient\nacquire / release / aclose\nLRU cap=OMP_POOL_CAP]
+    end
+
+    subgraph adapters/omp/omp_worker.py
+        OW[OmpWorker.handle\nextract pool_id+provider_session_id\nsession_file in JobResult.data]
+    end
+
+    subgraph infrastructure/stores/turn_store_session.py
+        TSS[TurnStoreSessionMixin\nget_cli_session / _set_cli_session\nreused as-is for omp token]
+    end
+
+    subgraph roxabi_contracts / core/llm_types.py
+        JE[JobEnvelope.payload\n+pool_id +provider_session_id optional]
+        JR[JobResult.data\n+session_file optional]
+    end
+
+    subgraph tests
+        T_PROTO[test_llm_protocols.py\nfalsification evidence]
+        T_AGENT[test_simple_agent.py\nparametrized reset x3 backends]
+        T_OMP[test_omp_rpc_driver.py\nrecall + isolation + wire-compat]
+        T_POOL[test_omp_pool.py\nunit]
+        T_WRK[test_omp_worker.py\nunit]
+    end
+
+    SA --> SB
+    WA --> SB
+    SB --> RB
+    ORD -->|implements| SA
+    ORD --> TSS
+    ORD --> JE
+    ORD --> JR
+    OW --> OP
+    OW --> JE
+    OW --> JR
+    T_PROTO --> SA
+    T_PROTO --> WA
+    T_AGENT --> SB
+    T_OMP --> ORD
+    T_POOL --> OP
+    T_WRK --> OW
+```
+
+---
+
+## Agents
+
+| Instance | Type | Tasks | Subjects |
+|----------|------|-------|---------|
+| backend-dev-1 | backend-dev | T1, T2 | protocol, contract |
+| backend-dev-2 | backend-dev | T3, T4 | pool, worker |
+| backend-dev-3 | backend-dev | T5, T6 | bootstrap, resume |
+| backend-dev-4 | backend-dev | T7, T12 | dispatch, followup |
+| devops-1 | devops | T8, T9 | deploy, manifest |
+| tester-1 | tester | T10, T11 | tests, integration |
+
+---
+
+## Wave Structure
+
+| Wave | Trigger | Agents | Tasks |
+|------|---------|--------|-------|
+| W1 | start | backend-dev-1, devops-1 | T1, T2, T8 |
+| W2 | T1+T2 done | backend-dev-2, backend-dev-3 | T3, T4, T5, T6 |
+| W3 | T3+T4+T5+T6 done | backend-dev-4 | T7 |
+| W4 | T8 done | devops-1 | T9 |
+| W5 | T6+T7 done | tester-1 | T10 |
+| W6 | T4+T5+T6+T7+T8+T9 done | tester-1 | T11 |
+| W7 | T7 done | backend-dev-4 | T12 |
+
+### Budget — per task
+
+| Task | Class | Est ops | Split? |
+|------|-------|---------|--------|
+| T1 | RED | 18 | N |
+| T2 | RED | 10 | N |
+| T3 | GREEN | 28 | N |
+| T4 | GREEN | 20 | N |
+| T5 | GREEN | 12 | N |
+| T6 | GREEN | 32 | N |
+| T7 | REFACTOR | 30 | N |
+| T8 | GREEN | 16 | N |
+| T9 | GREEN | 12 | N |
+| T10 | GREEN | 35 | N |
+| T11 | RED-GATE | 15 | N |
+| T12 | FOLLOW-UP | 6 | N |
+
+### Budget — per agent instance
+
+| Instance | Tasks | Σ ops | Subjects | Split? |
+|----------|-------|-------|---------|--------|
+| backend-dev-1 | T1, T2 | 28 | protocol, contract | N |
+| backend-dev-2 | T3, T4 | 48 | pool, worker | N |
+| backend-dev-3 | T5, T6 | 44 | bootstrap, resume | N |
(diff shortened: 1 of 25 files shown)


MickaelV0:
## Code Review — #1813 V2 Slice 2 (OmpPool + Native-Session Resume)

> Verdict: **request-changes** — 5 confirmed blockers. Design is architecturally sound; fixes are surgical and scoped.

---

### Blockers

**B4 — DEBT annotation missing `expires=` token** | confidence 99%
`src/factory/adapters/omp/omp_pool.py:169`
```python
except Exception:  # noqa: BLE001  — DEBT:boundary-broad-catch# pool cleanup
```
Root cause: `check_debt_expiry.sh` gate requires every `DEBT:` annotation to carry an `expires=YYYY-MM-DD` token. This form has no token and will trip the pre-push gate in CI.
Fix: append `expires=YYYY-MM-DD` (choose a concrete date, e.g. `expires=2026-09-01`) to the inline comment.

---

**B1 — `pool_id` never included in `JobEnvelope.payload`** | confidence 97%
`src/factory/llm/drivers/omp_rpc.py:167-173` (worktree V2)
```python
payload: dict[str, Any] = {
    "prompt": text,
    "model_cfg": model_cfg.model_dump(),
    "system_prompt": system_prompt,
}
if pending_resume is not None:
    payload["provider_session_id"] = pending_resume
```
Root cause: `pool_id` is computed and used for `_pending_resume.pop(pool_id, None)` but never written into the payload dict. The worker at `omp_worker.py:141-145` hard-rejects envelopes where `envelope.payload.get("pool_id", "")` is falsy with `ValueError("missing pool_id")`. Every hub→worker dispatch will fail immediately.
Fix: add `payload["pool_id"] = pool_id` inside `complete()` before the publish call.
Note: the existing unit test in `tests/llm/drivers/test_omp_rpc_session.py` asserts `provider_session_id` presence but never asserts `pool_id` presence — that gap allowed this to pass CI. Add the assertion.

---

**B3 — Worker hard-rejects old-format envelopes, contradicting spec and contracts docstring** | confidence 90%
Files: `src/factory/adapters/omp/omp_worker.py:141-145` vs. `packages/roxabi-contracts/src/roxabi_contracts/jobs/models.py:29` vs. `artifacts/plans/1813-runtime-selection-v2-plan.mdx` (T4)
Root cause: The contracts docstring explicitly documents that old `{prompt: str}`-only envelopes are "tolerated, treated as None on old senders." The plan (T4) specifies `pool_id = envelope.payload.get("pool_id", envelope.job_id)` as the correct fallback. The worker instead raises `ValueError("missing pool_id")` and publishes an error result, breaking any pre-V2 sender still in the queue (e.g. JetStream backlog replay — see project memory on contract required-field backward compat).
Fix: replace the hard-reject with the plan-specified fallback: `pool_id = envelope.payload.get("pool_id") or envelope.job_id`.

---

**B2 — `OmpPool._entries` has no `asyncio.Lock`; concurrent cold-starts are possible** | confidence 85%
`src/factory/adapters/omp/omp_pool.py:81,88-132`
```python
self._entries: dict[str, _PoolEntry] = {}   # line 81 — no lock
```
`acquire()` contains `await asyncio.to_thread()` yield points at lines 113 and 119-120 inside the check-then-act pattern. Two concurrent `acquire("same_pool_id")` calls can both observe a cache miss, both enter `_start_client()`, and both launch separate omp processes for the same `pool_id` — leaking one process and potentially corrupting the `session_file` entry.
Fix: add `self._lock = asyncio.Lock()` at construction; wrap the check-start-store block in `async with self._lock:`. A per-key lock (or `asyncio.Event` per in-flight cold-start) is preferable for throughput if pool contention is expected, but a single lock is correct and sufficient for Slice 2 concurrency levels.

---

**B5 — Quadlet volume ordering is advisory, not enforced** | confidence 85%
`deploy/quadlet/factory-omp.container:4`
```ini
After=factory-omp-sessions-volume.service
```
Root cause: `After=` alone creates an ordering dependency but not a binding one. If `factory-omp-sessions-volume.service` fails to activate (e.g. volume creation race on first-boot provisioning), systemd will still start `factory-omp` on schedule. The named volume mount will then fail at Podman level with a runtime error rather than a clean systemd unit failure.
Fix: add `Requires=factory-omp-sessions-volume.service` (or `BindsTo=` if you want factory-omp to stop when the volume unit stops) alongside the existing `After=`. This matches the pattern used by other volume-dependent units in the project.

---

### Warnings

**W1 — `_start_client()` in `OmpPool` bypasses `_verify_digest()`** | confidence 80%
`src/factory/adapters/omp/omp_pool.py:151-163`
Root cause: `RpcBridge.__init__` calls `_verify_digest()` once at bridge construction (documented in `adapters/omp/CLAUDE.md`). `OmpPool._start_client()` constructs a new `RpcClient` without going through `RpcBridge` — the digest gate is bypassed for every pool cold-start. If an attacker can replace `/opt/omp/omp` between bridge construction and a pool cold-start (low probability in the container model, but the gate exists precisely to enforce the pin), the bypass is silent.
Recommendation: extract `_verify_digest()` into a shared module-level call in `omp_pool.py.__init__` or in `_start_client()`, or route pool clients through the bridge constructor. Not a confirmed blocker because the `ReadOnly=true` container model limits the attack surface, but the gap should be documented or closed.

**W2 — Private method access across module boundary (`_set_cli_session`)** | confidence 78%
`src/factory/llm/drivers/omp_rpc.py:209`
```python
await self._turn_store._set_cli_session(  # noqa: SLF001
```
Root cause: `_OmpSessionStore` Protocol defines `_set_cli_session` with a leading underscore, making it nominally private. Accessing it from `omp_rpc.py` via `noqa: SLF001` is a workaround that signals the interface is misnamed. The Protocol should expose this as a public method (e.g. `set_cli_session`) since it is intentionally called cross-module.
Recommendation: rename to `set_cli_session` in the Protocol definition and all implementations; remove the `SLF001` suppression.

**W3 — `set_turn_store` not on `LlmProvider` protocol; wiring uses `# type: ignore`** | confidence 75%
`src/factory/bootstrap/factory/providers.py`
```python
omp_rpc_driver.set_turn_store(omp_turn_store)  # type: ignore[attr-defined]
```
Root cause: `set_turn_store` is a `SessionAware`-only method not declared on the `LlmProvider` base protocol. Pyright correctly rejects this; the suppression masks the gap. If the driver is ever swapped out for a non-`SessionAware` provider, this will silently become a no-op or an `AttributeError` at runtime.
Recommendation: introduce a `SessionAware` protocol check at the call site (`if isinstance(driver, SessionAware): driver.set_turn_store(...)`) or narrow the type at bootstrap to `OmpRpcDriver` explicitly.

**W4 — Integration test is CI-skipped; the only cross-component session-resume coverage is unit-only** | confidence 72%
`tests/integration/test_omp_session_resume_integration.py`
```python
pytestmark = pytest.mark.skipif(not os.environ.get("INTEGRATION"), ...)
```
Root cause: The integration test that exercises session-path persistence end-to-end is gated behind `INTEGRATION=1`. CI does not set this variable, so the resume path has no CI coverage. The existing unit test (`test_omp_rpc_session.py`) patches the turn store and does not exercise the actual pool interaction.
Recommendation: if a full integration test requires omp binary, mark it `@pytest.mark.integration` and gate on the env var — but ensure at minimum one unit-level test exercises `complete()` → payload → worker deserialization round-trip for both `pool_id` and `provider_session_id`.

**W5 — Integration test directly manipulates `driver._pending_resume` and `pool.acquire()`** | confidence 68%
`tests/integration/test_omp_session_resume_integration.py` (`test_switch_session_called_with_persisted_path`)
Root cause: The test manually pops `driver._pending_resume[pool_id]` and calls `pool.acquire()` directly, bypassing `driver.complete()`. This makes the test tautological with respect to the B1 bug (the payload gap goes untested) and couples the test to internal implementation details.
Recommendation: test through `driver.complete()` end-to-end; assert on the published NATS payload rather than private state.

---

### Suggestions

**S1 — Cache `_read_cap()` at construction rather than re-reading on every `_maybe_evict()` call**
`src/factory/adapters/omp/omp_pool.py:38-44`
`OmpPool._read_cap()` reads `OMP_POOL_CAP` from env on every eviction check. The value is a deployment constant; reading it per-call adds unnecessary overhead and makes it harder to reason about cap changes at runtime. Read once in `__init__` and store as `self._cap`.

**S2 — Add `pool_id` assertion to the existing unit test**
`tests/llm/drivers/test_omp_rpc_session.py`
After fixing B1, add `assert published_payload["pool_id"] == expected_pool_id` alongside the existing `provider_session_id` assertion. This prevents the B1 regression class from recurring.

**S3 — Inline import inside test method body**
`tests/integration/test_omp_session_resume_integration.py`
Inline imports inside test method bodies violate PEP 8 and make dependency scanning harder. Move to module-level imports.

**S4 — `omp/CLAUDE.md` still documents `PI_CODING_AGENT_DIR` as "writable tmpfs"**
`src/factory/adapters/omp/CLAUDE.md:36-38` vs. `deploy/quadlet/factory-omp.container:35`
The quadlet now uses a named persistent volume (`factory-omp-sessions.volume`) for `/home/factory/.config/omp-pi`, not a tmpfs. The CLAUDE.md invariant block still says "writable tmpfs (`mode=1777`)" which is now outdated and misleading. Update to reflect the named volume.

---

### Praise

The subscribe-before-publish pattern in `omp_rpc.py` (subscribing to `factory.job.<id>.result` before publishing the `JobEnvelope`) correctly eliminates the reply-race that existed in V1. This is the right fix and should be preserved.

`_pending_resume.pop(pool_id, None)` before publish (rather than after) is correct: it prevents double-consume on retry without introducing a state leak if the publish fails.

The `_result_sent` double-publish guard in `RpcBridge` (reset at the start of each `run()`, checked in both success and error paths) is a clean idempotency primitive that makes sequential jobs on one bridge safe without shared-state coordination.

ADR-073 SanitizedError discipline is correctly applied throughout: `_classify_exception` uses `type(exc).__name__` only; `str(exc)` / `repr(exc)` are absent from all bus-bound fields in the new code.

---

### Summary

This PR introduces a solid architectural foundation for pool-scoped session memory (subscribe-before-publish, _result_sent guard, ADR-073 discipline are all correct) but ships with a wire-breaking payload gap (B1) that will cause 100% job failure on first deploy. B2 (no lock on the pool dict) and B3 (hard-reject vs. documented fallback) compound the risk. B4 and B5 are gate and ops correctness issues. All five blockers have surgical, scoped fixes; the design itself does not require rework.

**Verdict: request-changes** — 5 confirmed blockers must be resolved before merge.

MickaelV0:
## Review Fixes Applied

All 5 confirmed blockers + 2 warnings resolved in `e1b00fc7`. Full suite (5861) + pyright (CI scope, incl. tests) + integration (`INTEGRATION=1`) + pre-push gates all green.

### Applied
- **B1** [blocker] `OmpRpcDriver.complete()` now writes `pool_id` into the `JobEnvelope.payload` — the worker routes on it. **Test added** asserting `pool_id` rides the wire (the gap that let this pass CI).
- **B3** [blocker] `OmpWorker.handle()` falls back to `job_id` when `pool_id` is absent (wire-compat with pre-V2 senders / JetStream replay) instead of hard-rejecting. Test `test_handle_rejects_missing_pool_id` → `test_handle_falls_back_to_job_id_when_pool_id_absent`.
- **B2** [blocker] `OmpPool` serializes cold-starts under an `asyncio.Lock` with double-checked locking — concurrent `acquire()` for one `pool_id` no longer launches duplicate omp processes.
- **B5** [blocker] `factory-omp.container` adds `Requires=factory-omp-sessions-volume.service` (binding, not just `After=`).
- **W1** `OmpPool._start_client` now enforces the omp binary digest pin (reuses `_rpc_bridge._verify_digest` — single source, no duplicated sha256).
- **W3** `providers.py` narrows `set_turn_store` via `isinstance(OmpRpcDriver)` (removed `# type: ignore`).
- Bonus: `reset()` also clears `_lyra_sessions[pool_id]` (stale-mapping leak); tidied a DEBT comment.

### Adjudicated — not applied
- **B4** — false positive. `check_debt_expiry.sh` uses a 6-month grace window, not a required `expires=` token; the gate passes. Only the (now-fixed) glued comment was real.
- **W2** (rename `_set_cli_session` → public) — deferred to **#1893**: renaming the established `TurnStore` token method spans ~47 callsites incl. the live clipool path; that is the tracked token-rename, out of scope for this slice.
- **W4** (integration CI-skipped) — by design (`INTEGRATION=1`, M₁-manual); unit coverage closes the cross-component gap (B1/B3 wire tests).

Issue #1906, opened 2026-06-16
chore(deploy): declare dynamic per-bot token class in secrets-policy.toml (Control 11 + drift gates false-WARN on live bot tokens)


MickaelV0:
**Origin:** #1901 Control 11 surfaced `factory-bot-{telegram,discord}-{lyra,aryl}` as "orphan" Podman secrets on M₁. They are **not stale** — they are the live Telegram/Discord **bot tokens**, mounted by the deployed adapters (`factory-telegram.container:42-43`, `factory-discord.container:42-43`, `target=bot_token-<bot>`). The flag is a **policy-coverage gap**: `secrets-policy.toml` enumerates only fixed infra secrets, but per-bot tokens are a *dynamic family*.

## Why naive declaration breaks
Per-bot secrets are rendered at install from `config.toml [[auth.<platform>_bots]]` by `tools/render_quadlet.py` (`{{bot_secrets}}` → `Secret=factory-bot-{platform}-{bot_id}`, plus a `-webhook` variant at render_quadlet.py:53). They are **absent from `quadlet.toml` required_secrets** (e.g. telegram = `["factory-nats-telegram", "factory_blobstore_token"]` only).

→ Adding exact `[secret.factory-bot-…]` entries would **fail `check_secrets_drift` check (d)** (the `required_secrets ↔ [secret.*]` bidirectional assertion: policy entry with no matching required_secret).

## Proposed design — a dynamic *class*, not exact entries
Introduce a class section, e.g.:
```toml
[secret-class.bot-token]
pattern  = "factory-bot-{platform}-{bot_id}"   # + optional "-webhook" suffix
policy   = "bot-provisioned"
source   = "config.toml [[auth.<platform>_bots]] — created by `factory bot init`"
```
Then teach the three consumers:
1. **check_secrets_drift check (d):** members matching a `[secret-class.*]` pattern are **exempt** from the required_secrets bidirectional check (they're dynamic, ¬static required).
2. **check_secrets_source.sh:** class members have no canonical seed file → treat like `source=n/a`/provisioned (¬FAIL).
3. **baseline check.sh Control 11:** a live `factory-bot-*` secret is recognized via the class instead of WARN-orphan.

## High-value extension (recommended)
Control 11 should cross-check each live `factory-bot-<platform>-<bot>` secret against the **configured** bots in `~/.roxabi/factory/config.toml [[auth.<platform>_bots]]`:
- bot IS configured → ✅ (active token)
[...]

Issue #1905, opened 2026-06-16
chore(deploy): --frozen on host-side uv invocations (anti-recurrence: uv.lock churn jammed M₁ ff-only pull)


MickaelV0:
**Origin:** deploy incident while shipping #1900 wave 1 (mitigated by #1903 lock-fix + #1899 dirty-guard — but the churn *vector* remains).

## Problem
Host-side `uv run` invocations in the quadlet render / ACL-spec targets re-resolve the environment with no `--frozen`/`--locked`. If the committed `uv.lock` drifts from `pyproject.toml` (as it did: roxabi-contracts 0.7.0 vs installed 0.11.0), `uv run` **silently rewrites `uv.lock`** in the working tree → the factory-quadlet-sync timer's ff-only pull then hits the #1899 dirty-checkout guard and refuses to pull → total deploy-pipeline stall until a manual `git checkout uv.lock` on M₁.

## Evidence (Makefile)
```
160: @uv run python tools/render_quadlet.py ...   # quadlet render (host path)
165: @uv run python tools/render_quadlet.py ...
323: @uv run python scripts/render_acl_spec.py
324: @uv run python scripts/render_acl_parity.py
```
None carry `--frozen`. (Dev-only targets — test/lint at 367+ — are out of scope.)

## Fix
Add `--frozen` (or `--locked`) to host-side `uv run` in the render/install path so a stale lock **fails loudly** instead of mutating the working tree and jamming the dirty-guard. Scope: the converge/quadlet-install render targets, not the dev workflow targets.

## Acceptance
- host-side render targets use `uv run --frozen`
- a deliberately-stale `uv.lock` makes `make quadlet-install` (render path) error, not rewrite the lock
- factory-quadlet-sync ff-only pull no longer at risk from render-induced lock churn

PR #1908, opened 2026-06-16
chore(deploy): declare dynamic per-bot token secret class

MickaelV0:
Declares the `bot-token` secret class in `deploy/secrets-policy.toml` and
wires the two gate scripts to handle it correctly.

## Changes

- `deploy/secrets-policy.toml` — adds `[secret-class.bot-token]` section with
  `pattern`, `policy = "bot-provisioned"`, `source = "n/a"`; updates header
  comment block to document the new `bot-provisioned` policy and
  `secret-class` TOML namespacing.

- `tools/check_secrets_drift.sh` — adds bot-class pattern loader and
  `is_bot_class_member()` helper to exempt dynamic bot secrets from the
  reverse direction of check (d). Class entries never reach `POLICY_SET`
  (parsed under `data["secret-class"]`, separate TOML key), so the forward
  direction is unaffected.

- `tools/check_secrets_source.sh` — adds explicit `bot-provisioned` case
  (WARN, not FAIL) to the policy dispatch. Belt-and-suspenders: class members
  are already skipped via `source = "n/a"` before the dispatch is reached.

- `docs/QUADLET-DEPLOYMENT.md` — adds subsection `### (e) Bot-token secret
  class` documenting the class pattern, why members are absent from
  `required_secrets`, and how they are provisioned.

## Gate results

- `check_secrets_drift.sh` → `check_secrets_drift: all checks passed — OK`
- `check_secrets_source.sh` → `SKIP` (data dir absent on dev machine — expected)

## Adversarial proof

A stray `[secret.factory-bot-xyz]` (no matching `required_secrets`) still
fails check (d) with:
`orphan policy entry not referenced by any required_secrets: factory-bot-xyz`

Refs #1906


MickaelV0:
Superseded by #1909. This branch was accidentally cut from `chore/1905-frozen-host-uv` instead of `staging`, so it inherited the #1905 Makefile `--frozen` commit (5 files instead of 4) and failed the red-gate Makefile-shape test. #1909 is the clean rebuild off staging (4 files only). Companion projects-meta PR: #6.

PR #1909, opened 2026-06-16
chore(deploy): declare dynamic per-bot token secret class

MickaelV0:
Refs #1906

Declares a dynamic `[secret-class.bot-token]` in `deploy/secrets-policy.toml` and teaches the drift/source gates to recognise per-bot tokens (`factory-bot-{platform}-{bot_id}`, incl. `-webhook`) as a *provisioned class* rather than false-flagging them.

**Files (4):**
- `deploy/secrets-policy.toml` — `[secret-class.bot-token]` section + header note
- `tools/check_secrets_drift.sh` — bot-class loader + `is_bot_class_member()`; class entries excluded from `POLICY_SET` via TOML namespacing (check (d) stays green)
- `tools/check_secrets_source.sh` — explicit `bot-provisioned` case (WARN, not FAIL)
- `docs/QUADLET-DEPLOYMENT.md` — `### (e) Bot-token secret class`

**Verified locally:** `check_secrets_drift.sh` → all checks passed (incl. check (d) bidirectional); `check_secrets_source.sh` → OK. Adversarial: a stray static `[secret.factory-bot-xyz]` with no `required_secrets` match still fails check (d) (exemption is class-scoped, not blanket).

**Supersedes #1908** — that branch accidentally inherited the #1905 Makefile `--frozen` commit (branched off `chore/1905` instead of `staging`). Rebuilt cleanly off `staging` (cherry-pick of the 4-file commit only). Companion: projects-meta PR #6 (Control 11).


MickaelV0:
## Code Review — PR #1909

Fresh multi-domain review (devops · security-auditor), findings verified empirically by the orchestrator against the running gate.

### 🔴 Blocking — dead + latently-broken exemption code

`issue:` `tools/check_secrets_drift.sh` ships `BOT_CLASS_PATTERNS` + `is_bot_class_member()` that are **dead and broken**.
- **Dead:** `is_bot_class_member()` has **zero call sites** — neither check (d) loop (`in_quadlet_not_policy`, `in_policy_not_quadlet`) invokes it.
- **Broken:** the ERE builder calls `re.escape()` **before** `re.sub(r'\{[^}]+\}', '[^-]+', …)`, producing `factory\-bot\-\[^-]+\-\[^-]+` — the `\[` is a literal bracket, so it matches **no** real secret name (verified: `factory-bot-telegram-mybot` → no match).
- **Misleading:** the comment says the load is "used only … for the reverse exemption guard" — it is used nowhere.
- Root cause: the exemption was implemented as runtime bash logic, but the class is already isolated **structurally** — `tomllib` parses `[secret-class.bot-token]` under `data["secret-class"]`, a separate key from `data["secret"]`, so it never enters `POLICY_SET`; and bot tokens never enter `QUADLET_SECRETS` (rendered dynamically via `{{bot_secrets}}`, absent from `required_secrets`). **The gate is already correct via TOML namespacing.**
- Confidence: 97%
- Severity correction: this is **not** gate-breaking — check (d) passes green today. The defect is dead/broken/misleading code shipping in a gate-honesty PR.
- Solutions:
  1. **(recommended)** Remove `BOT_CLASS_PATTERNS` + `is_bot_class_member()`; keep the explanatory comment crediting TOML namespacing as the isolation mechanism. Likewise drop the structurally-unreachable `bot-provisioned` branch in `check_secrets_source.sh` (its Python loop iterates `data["secret"]` only, so class members never reach it).
  2. If a real runtime exemption path is desired later, wire it into the check (d) loops **and** fix the ERE order (`re.sub` first, then escape only literal segments) **and** allow hyphens in `bot_id` (`[^-]+` rejects `my-bot`).

`praise:` Using a distinct top-level table (`[secret-class.*]` vs `[secret.*]`) instead of a naming convention structurally prevents parse-confusion — `data.get("secret", {})` can never return class metadata. Correct design choice.

### Verdict: **Request changes** — remove the dead/broken exemption code (the TOML structure already does the work). `secrets-policy.toml` declaration + docs are good.

Issue #2041, opened 2026-06-28
[factory/projects-meta] Timers + dead-code cleanup: OnFailure, Persistent stagger, setup.sh, orphan_ignore, TOCTOU stamp, blobstore-sweep


MickaelV0:
**Repo:** roxabi-factory + projects-meta. Low-sev confirmed findings, batched.

## Items
- `podman-auto-update.service`: add `OnFailure=factory-deploy-failure.service` (pull/restart failures currently invisible).
- 3 deploy timers `Persistent=true` collapses the 2-min stagger at boot → reassess.
- delete `deploy/nats/setup.sh` (exits 1 immediately, dead artifact).
- `hosts.toml orphan_ignore`: orphan detection only checks `*.container` → `.network`/`.volume` entries are inert (cover them or drop).
- `converge.sh:124` `write_convergence_state` recomputes independently (TOCTOU) → pass the computed state.
- `factory-blobstore-sweep.{service,timer}` not installed by `quadlet-sync-install` → add.

## Acceptance
- dead artifact removed; timers have failure handling; stamp passes computed state; sweep units installed

MickaelV0:
✅ Livré via PR #2045 + #2046 (Refs #2041, jamais Closes).

| Item | Statut |
|---|---|
| OnFailure sur podman-auto-update | ✅ |
| Persistent=false sur timers offset (*:2/5, *:4/5) | ✅ |
| Suppression deploy/nats/setup.sh | ✅ |
| factory-blobstore-sweep dans quadlet-sync-install | ✅ |
| TOCTOU stamp (write_convergence_state) | ✅ #2046 |
| hosts.toml orphan_ignore (.network/.volume) | ⏭️ hors scope — repo projects-meta |

Les 5 items roxabi-factory sont en prod. Le point orphan_ignore reste un micro-ticket projects-meta si encore pertinent.

Issue #2039, opened 2026-06-28
[factory] Stale >=9 guards → dynamic count + active_containers ordering + AGENTS.md converge sequence


MickaelV0:
**Repo:** roxabi-factory.

## Problem
- `converge.sh:103` and `quadlet-install-verify.sh:21` guard `>= 9` but quadlet.toml has **20** components → a partial parse (9–19) passes silently.
- `deploy.sh` marks `active_containers[$file]=1` (line 168) **before** the missing-source check (line 170) → `--prune` protects uninstallable phantoms.
- `deploy/AGENTS.md` converge sequence omits STEP 3 (`deploy.sh --prune`) → S2 invisible to doc-first debugging.

## Acceptance
- guards derive expected count dynamically (e.g. from `quadlet_containers()`)
- active_containers marked only after source confirmed
- AGENTS.md sequence matches converge.sh

Issue #1930, opened 2026-06-17
chore(ops): retire deprecated deploy targets and fix doc drift


MickaelV0:
## Scope

1. Replace `nats-setup` with cold-path `factory-acl genkeys --ack-external-distribution`
2. `deploy/nats/setup.sh` → exit 1
3. Fix runbook, `deploy/CLAUDE.md`, GETTING-STARTED, DEPLOYMENT, systemd Description
4. Deprecate `full-deploy` / remote `deploy`

**Epic:** E3 · **Wave:** W1

PR #2045, opened 2026-06-28
fix(systemd): fix Persistent= on offset timers, wire blobstore-sweep + OnFailure drop-in (#2041)

MickaelV0:
## Summary

- **Persistent= on offset timers** (`factory-post-autoupdate.timer`, `podman-auto-update.timer.d/override.conf`): `Persistent=true` on offset timers (*:2/5 and *:4/5) collapses the *:0/5→*:2/5→*:4/5 boot stagger — all three fire simultaneously at boot, defeating the collision-avoidance design. Fixed: both offset timers now have `Persistent=false`. Primary timer (`factory-quadlet-sync.timer` at *:0/5) and non-stagger timers (daily `blobstore-sweep`, weekly `logrotate`) are untouched.
- **`OnFailure=` drop-in** (`deploy/systemd/podman-auto-update.service.d/override.conf`): new drop-in wires `factory-deploy-failure.service` to trigger on `podman-auto-update.service` failure. Install wired into `quadlet-sync-install`.
- **`factory-blobstore-sweep` in Makefile**: both unit files existed in `deploy/systemd/` but were absent from `quadlet-sync-install` — install gap closed.
- **`deploy/nats/setup.sh` deleted**: dead since #1930 (file exits 1 immediately with "retired" message). No in-scope Makefile targets or scripts reference it.

## Validation

- `grep -n Persistent deploy/systemd/*.timer deploy/systemd/podman-auto-update.timer.d/override.conf` — confirmed: primary *:0/5 = `Persistent=true`; offset *:2/5 and *:4/5 = `Persistent=false`; non-stagger daily/weekly = `Persistent=true`
- `make -n quadlet-sync-install` — dry-run shows all 14 cp operations resolve; all new units included
- `git status --short` — exactly 5 scoped files (M Makefile, D deploy/nats/setup.sh, M factory-post-autoupdate.timer, A podman-auto-update.service.d/override.conf, M podman-auto-update.timer.d/override.conf)
- `bash -n Makefile` — exit 0 (Makefile-native syntax warnings are expected; not shell errors)
- `systemctl --user restart` + `enable --now` commands on M₁ not locally validatable (requires the host); `make quadlet-sync-install` should be run post-merge on M₁

## Deferred (out of scope)

`deploy/AGENTS.md` has one stale reference to `deploy/nats/setup.sh` (UFW note) — that file is owned by another agent (#2039).

Refs #2041

Review-gated: do not add the reviewed label; do not merge.


MickaelV0:
## Code Review + /fix (`c8a58b6e`)

Multi-domain review (devops/security). **1 high blocker found + fixed:**

| Sev | Finding | Fix |
|---|---|---|
| 🔴 **high** | `Persistent=false  # comment` — systemd has **no inline comments**; the `#…` is parsed into the boolean → directive **voided** → base `Persistent=true` survived → the *:0/5→*:2/5→*:4/5 stagger was **not actually preserved** (the bug this PR set out to fix). In override.conf + factory-post-autoupdate.timer. | comments moved to their own lines. Verified: no `Directive=value # …` inline pattern remains; only the two **offset** timers drop Persistent (primary `*:0/5` + daily/weekly keep `Persistent=true` — correct). |
| 🟡 medium | factory-deploy-failure message named 2 of 3 trigger units | now names all three (podman-auto-update gained OnFailure→here in this PR). |

Follow-up (non-blocking, **not** fixed — archival doc): ADR-045 has a passing mention of the now-deleted `deploy/nats/setup.sh`. ADRs are point-in-time records, so left as-is rather than rewriting history.

Review-gated.

Issue #2034, opened 2026-06-28
epic: Deploy-infra hardening — single SSoT, one installer, scoped converge


MickaelV0:
## Post-audit epic (deploy infrastructure)

**Root cause:** `~/projects/deploy.sh` is a strict, all-or-nothing, **cluster-wide** installer that the single-repo `roxabi-factory/deploy/converge.sh` hard-depends on under `set -e`. One foreign-repo secret, one ghost git worktree, or one template-only unit aborts the entire factory deploy. Automated converge on M1 has been **silently failing every 5 min** (OnFailure firing) since telegram/discord became `.container.tmpl`; M1 coasts on manual `make quadlet-install`.

**Axial drift (DEPLOY domain):** no single SSoT for *what runs where* (5 registries) and two installers share one write namespace with incompatible semantics.

### Children
- **A** unify installers (RENDER action) + `template=true` flag — *keystone, fixes S2*
- **B** scope cluster_plan glob (managed_repos allowlist) — *fixes S1*
- **C** per-repo error isolation + optional-secret warn + llmCLI host_roles — *fixes S3*
- **D** CI lint: component-has-source — *prevents S2 regression*
- **E** stale `>=9` guards + active_containers order + AGENTS.md doc
- **F** CI: build image in PR + gate publish + pin base images
- **G** timers + dead-code cleanup
- **H** [OPS] M1 cleanup (manual) — *blocked by A*

> Code for A/B/C/G lives in repo **projects-meta** (`~/projects`, manual-merge) and **llmCLI**; filed here for one coherent epic.

MickaelV0:
## Dev-cycle status — 7 PRs open (all review-gated, none merged)

| PR | Issue(s) | Repo | What |
|---|---|---|---|
| [projects-meta#7](https://github.com/Roxabi/projects-meta/pull/7) | #2035 #2036 (+#2039/#2041 partial) | projects-meta | RENDER action + managed_repos allowlist + safe orphan detection + active_containers ordering |
| [#2044](https://github.com/Roxabi/roxabi-factory/pull/2044) | #2035 | factory | `template=true` flag on telegram/discord |
| [llmCLI#138](https://github.com/Roxabi/llmCLI/pull/138) | #2037 (partial) | llmCLI | proxy host_roles `any`→`llm-worker` |
| [#2043](https://github.com/Roxabi/roxabi-factory/pull/2043) | #2038 | factory | component-source CI gate |
[...]

PR #2046, opened 2026-06-28
fix(deploy): dynamic unit-count guards + converge-sequence doc + TOCTOU stamp

### LINKED ISSUE #2039: [factory] Stale >=9 guards → dynamic count + active_containers ordering + AGENTS.md converge sequence

MickaelV0:
Part of epic #2034.

## Changes
- **#2039 — stale guard:** `converge.sh` and `quadlet-install-verify.sh` replace the hardcoded `>= 9` sanity guard (manifest has **20** components) with a lower bound derived from `deploy/quadlet.toml` `[component.*]` count at runtime — auto-scales, still catches an empty/partial `quadlet_containers` parse.
- **#2039 — doc:** `deploy/AGENTS.md` Convergence sequence was missing **STEP 3 = `deploy.sh --prune`** (the cluster role-aware installer that runs before the factory template render) — corrected + renumbered 3..10. Also dropped the dangling `deploy/nats/setup.sh` UFW reference (script deleted as dead in #2041; firewall base policy is `deploy/provision.sh`).
- **#2041 (TOCTOU):** `write_convergence_state` now accepts a pre-computed fingerprint (default = recompute, back-compat); `converge.sh` computes the post-converge state once and stamps exactly that.

## Validation (local, M2)
- `bash -n` converge.sh + quadlet-install-verify.sh → OK
- dynamic guard floor resolves to 20 (= component count) ✓
- `tools/check_doc_drift.py` → 0 new violations · `tools/check_doc_semantic_drift.py` → 0
- scope: exactly 4 files

> Recovery note: a prior dev agent thrashed on context here; its correct partial work (converge.sh guard + deploy-common.sh capability) was salvaged and completed.

Closes #2039 · Refs #2041 (TOCTOU)
Review-gated: do not add the reviewed label; do not merge.


MickaelV0:
## Code Review

**PR #2046** — `fix(deploy): dynamic unit-count guards + converge-sequence doc + TOCTOU stamp`

---

### Blockers

#### issue(blocking): `grep -cE` under `set -e` aborts script before user-friendly guard fires
`deploy/converge.sh:104` · `deploy/quadlet-install-verify.sh:21`

Root cause: `grep -c` exits **1** when it finds 0 matches, and exits **2** on a missing file. Both scripts run under `set -euo pipefail`. The assignment `_min_units=$(grep -cE '^\[component\.' ...)` therefore terminates the process immediately — before the actionable error message at line 105 / line 22 can execute. Any case where `quadlet.toml` is absent or has zero `[component.*]` sections produces a silent abort rather than the intended `ERROR: quadlet_containers returned N units (<M)`.

Class: `bash-error-suppression`  
Raw callsites: `[{deploy/converge.sh:104}, {deploy/quadlet-install-verify.sh:21}]`

Fix (both sites):
```bash
# 1. suppress grep's non-zero exit so set -e doesn't fire
_min_units=$(grep -cE '^\[component\.' "$(dirname "${BASH_SOURCE[0]}")/quadlet.toml" || true)
# 2. guard explicitly against an empty/missing parse result
[[ ${_min_units:-0} -gt 0 ]] || {
    echo "ERROR: could not parse component count from quadlet.toml" >&2; exit 1
}
# 3. existing unit-count guard is now reachable
[[ ${#_all_svcs[@]} -ge ${_min_units} ]] || { echo "ERROR: ..."; exit 1; }
```

Note: the same script already uses `|| true` correctly on the client-service filter at `converge.sh:106` — the omission here is inconsistent.

Confidence: 95%

---

### Suggestions / Thoughts

#### thought: `mapfile ... < <(quadlet_containers)` masks `quadlet_containers` failure — interacts with the grep fix
`deploy/converge.sh:102` · `deploy/quadlet-install-verify.sh:19`

Root cause: Bash process substitution reports `mapfile`'s exit status (0 on success), never the subprocess's. If `quadlet_containers` returns 1, `_all_svcs` / `UNITS` is empty and `set -e` does not trip. The `_min_units` guard compensates — but only while `_min_units > 0`. If finding #1 is fixed with `|| true`, a missing `quadlet.toml` sets `_min_units=0`, and `0 -ge 0` passes silently.

Fix: add an explicit floor check on the array before the `>= _min_units` comparison:
```bash
[[ ${#_all_svcs[@]} -gt 0 ]] || { echo "ERROR: quadlet_containers returned 0 units" >&2; exit 1; }
```

Confidence: 65%

#### suggestion: Component-header count and `container =` line count can silently diverge
`deploy/converge.sh:104`

Root cause: `_min_units` counts `[component.*]` TOML section headers; `quadlet_containers` counts `container = "..."` value lines. Both happen to equal 20 today, but a future `[component.*]` header added without a `container =` key would make the floor exceed the actual count → false-fail on every deploy. A `container =` outside a component section would cause the inverse.

Fix: derive the floor from the same grep that produces `quadlet_containers` output, or add a comment asserting the invariant (`# assumes every [component.*] section has exactly one container = key`).

Confidence: 75%

---

### Nitpicks

#### nitpick: AGENTS.md step 2 places voiceCLI pull at the wrong position in the sequence
`deploy/AGENTS.md:128`

Root cause: Step 2 says "(and `~/projects/voiceCLI` if present)" alongside the factory `git pull`, implying they happen together. In `converge.sh` the voiceCLI pull is step 5 — after Quadlet install (step 3) and template render (step 4). AGENTS.md therefore has 10 steps covering the logical flow but silently merges converge.sh's step 5 into step 2, making internal step numbers off-by-1 from step 5 onward.

Fix: either add a separate step 5 "(Pull voiceCLI if present)" to match `converge.sh`'s comment, or note in step 2 that the voiceCLI HEAD pull occurs later in code (step 5).

Confidence: 90%

---

### Praise

- `write_convergence_state(state="${1:-}")` back-compat is solid: `[[ -n "${state}" ]]` gates correctly on empty-string-safe check; falls back to `compute_convergence_state` for zero-arg callers. There is exactly one real caller in the repo (the test stub is arg-agnostic), so risk is nil.
- The "TOCTOU" framing is technically precise: capturing `_final=$(compute_convergence_state)` in the caller rather than inside `write_convergence_state` guarantees the exact post-converge state is stamped, preventing a second independent recompute from racing against an external change.
- UFW reference update is accurate: `deploy/nats/setup.sh` was removed in #2041; `deploy/provision.sh` base policy + manual 4222 rule is the correct replacement.
- The `|| true` guard on the client-services grep filter (`grep -v '^factory-nats$' || true`) is correct and consistent — which makes the omission on `grep -cE` stand out as an oversight rather than intent.

---

**Verdict: Request changes** — one confirmed `bash-error-suppression` bug (two callsites) prevents the user-friendly error guard from ever firing; requires fix before merge.

MickaelV0:
## Code Review + /fix (`73a4f855`)

Multi-domain review (devops/security/tester). **1 critical blocker found + fixed:**

| Sev | Finding | Fix |
|---|---|---|
| 🔴 **critical** | `_min_units=$(grep -cE '^\[component\.' …)` — `grep -c` exits **1 on zero matches** (2 on missing file); under `set -euo pipefail` this aborted the script **before** the friendly lower-bound guard — i.e. the empty/corrupt-manifest case the guard exists to report was silently swallowed. In **both** converge.sh:104 and quadlet-install-verify.sh:21. | added explicit `\|\| { echo "…no [component.*] sections…"; exit 1; }` handler. Verified: empty manifest now prints the error + exits, happy path counts 20. |
| 🟡 doc | `write_convergence_state` docstring said state was verified at the **START**, but the caller passes the **END**-computed state | corrected docstring |

Tester/security **praise**: dynamic `_min_units` correctly replaces the stale `9`; `local _min_units` split from assignment preserves set-e; TOCTOU stamp fix is a genuine improvement. Tautology concern (both sides parse same file) — noted: the patterns differ (`^\[component\.` headers vs quadlet_containers' `^container =` lines), so it's a real cross-check, not tautological.

`bash -n` ✓ · trap behavior verified. Review-gated.

Issue #2037, opened 2026-06-28
[projects-meta/llmCLI] Per-repo error isolation in deploy.sh + optional-secret warn + llmCLI host_roles


MickaelV0:
**Repo:** projects-meta + llmCLI. **Fixes S3 + axial fault-tolerance/ownership.**

## Problem
`deploy.sh` keeps a flat `errors` counter and `exit 1` on any failure; `converge.sh` hard-depends under `set -e`. A foreign repo's missing secret aborts the **factory** converge. `verify_secrets()` ignores policy optionality (optional secrets treated as fatal). llmCLI proxy `host_roles=['any']` lands its secret check on M1.

## Approach
- `deploy.sh`: tiered exit (hard vs warn) and/or `--repo` filter / per-repo error buckets; optional secrets WARN not fail.
- `llmCLI/deploy/quadlet.toml`: `host_roles ['any'] → ['llm-worker']` (proxy) — confirm M1 still gets the components it needs via explicit roles.

## Acceptance
- a missing foreign-repo secret WARNS, does not abort factory converge
- converge.sh can scope deploy.sh to roxabi-factory

MickaelV0:
**Partial:** the llmCLI `host_roles` half is shipped in Roxabi/llmCLI#138 (`["any"]→["llm-worker"]`), removing the cross-repo blast radius on M1.

**Deferred (needs your decision):** the `deploy.sh` per-repo **error-model tiering** — i.e. what failure contract a cluster-wide installer should have when a *foreign* repo's secret/source is missing. Options:
1. Tiered exit codes (2=hard/abort, 1=warn/continue) + converge treats warn as non-fatal.
2. `deploy.sh --repo <name>` scope flag so converge only hard-fails on its own repo.
3. Per-repo error buckets (foreign-repo errors warn, primary-repo errors abort).

The `managed_repos` allowlist (#2036) + the host_roles fix already remove the *known* blast sources, so this is no longer urgent — but the contract itself is a design call. Not implementing autonomously.

MickaelV0:
[...]

PR #2050, opened 2026-06-28
fix(deploy): handle deploy.sh tiered exit in converge (#2037)

MickaelV0:
Part of epic #2034 (#2037). Consumer side of the projects-meta `deploy.sh` tiered-exit change.

converge.sh captures `deploy.sh --prune`'s exit code without tripping `set -e`: **abort on ≥2** (hard — missing unit source), **warn + continue on 1** (soft — missing secret; secrets are installed at step 7 and a genuinely-missing one fails the restart at step 9).

⚠ Merge **after** the projects-meta deploy.sh PR (it defines the exit codes).
`bash -n` ✓. Review-gated.


MickaelV0:
## Code Review + /fix

2 read-only agents (devops + security). **Core fix confirmed correct** — `_split_tsv` sound for all field counts, `printf -v '%s'` safe against `%`/`\`/manifest values, no new injection surface, tier categorization correct (missing-source→hard exit 2, missing-secret→soft exit 1), `set -e` capture idiomatic. Fixed the findings:

| Finding | Fix |
|---|---|
| `deploy.sh` exit 2 also covers unknown-flag/syntax errors → converge's "missing source" message would mislead | converge message now generic + **propagates the exit code** (`exit ${_deploy_rc}`) |
| `_split_tsv` silently drops surplus fields | documented in the header |
| security: soft-exit installs unit before secret exists | confirmed **safe** — podman enforces `Secret=` at container start (no degraded boot); converge ordering (install→provision→restart) already documented inline |

Validation: `bash -n` ✓ · `_split_tsv` unit test ✓ · dry-run rc=1 (soft, no missing source) ✓. Review-gated.

PR #2083, opened 2026-07-01
fix(converge): daemon-reload after quadlet-install NO_RESTART=1

MickaelV0:
## Bug (audit 2026-06-30 — deploy provenance A, confirmed live at HEAD)

`converge.sh` step 4 runs `make quadlet-install NO_RESTART=1`. That flag **skips `daemon-reload`** — the reload lives solely in `deploy/quadlet-install-verify.sh`, invoked only in the non-`NO_RESTART` else-branch of the Makefile `quadlet-install` target (confirmed `Makefile:165-191` + `docs/CONFIGURATION.md:693`). `converge.sh` itself had **zero** `daemon-reload` calls.

**Consequence:** Quadlet unit content installed/rendered in steps 3-4 — a bot add/remove (→ new/removed `Secret=` mount line in `factory-telegram`/`-discord.container`), an image digest bump, a hardening/volume edit — is copied to disk but systemd never regenerates the `.service` before converge's own restarts fire. The restarts relaunch from the **stale generated unit** and silently ignore the new content until an out-of-band `make quadlet-install` or reboot.

`deploy/AGENTS.md:157` falsely claimed `NO_RESTART=1` includes `daemon-reload` (it does not); `docs/CONFIGURATION.md` was already correct.

## Fix
- `converge.sh`: explicit, **unconditional** `systemctl --user daemon-reload` right after step 4, before any restart (steps 3-4 run for every drift kind; only the restart strategy branches auth/structural).
- Tests: the two `test_no_reload` cases asserted `not any("reload" ...)` — they locked in the buggy "never reloads" status quo. Replaced each with `test_daemon_reload_called` (asserts the reload now runs) + `test_no_unit_scoped_reload` (converge still must NOT use the unit-scoped `reload factory-nats` SIGHUP path — that lives only in `make nats-add-identity`). **11 passed** locally.
- `deploy/AGENTS.md`: corrected step 4 + added step 4b.

## Deploy note
M₁ auto-converges on merge; this changes converge to reload before restarts. Verify on M₁ post-merge that a converge run logs `daemon-reload` and picks up unit changes (no stale-unit relaunch).

### DIFF
--- a/deploy/AGENTS.md
+++ b/deploy/AGENTS.md
@@ -154,7 +154,8 @@ Legacy 4-field stamps (pre-image-digest schema) are normalized to `:none:none` o
 1. **Change-gate** — skip if already converged.
 2. **Pull** — `git pull origin staging` in `~/projects/roxabi-factory` (and `~/projects/voiceCLI` if present).
 3. **Install cluster Quadlets** — `bash ~/projects/deploy.sh --prune` (role-aware SSOT: installs every host-matched unit across all managed repos from `hosts.toml` × `*/deploy/quadlet.toml`, generates host-override drop-ins, prunes orphan `.container` files). Runs under `converge.sh`'s `set -e` — **must succeed before step 4**. Template-only components (telegram/discord) are emitted as RENDER actions and rendered in step 4, not copied here.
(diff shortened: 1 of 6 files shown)


MickaelV0:
Superseded by clean rebuild https://github.com/Roxabi/roxabi-factory/pull/2086 (bot merged an unrelated intermediate staging state that conflicted + polluted this branch's diff).

PR #2084, opened 2026-07-01
fix(gate): stop hardcoding container count in stale_container_count message

MickaelV0:
## Bug (audit 2026-06-30 — gate hygiene)

`tools/check_doc_semantic_drift.py`'s `stale_container_count` rule exists to catch stale container counts in docs — but its own fix-message hardcoded one: *"production stack is nine Quadlet containers"*. The live count is now **26** (`grep -c 'container = ' deploy/quadlet.toml`) and has climbed for weeks, so the gate's guidance became the exact stale fact it's meant to correct (self-referential; previously bumped six→nine once already).

## Fix
- Message → points at the generated SSoT (`docs/architecture/CURRENT.generated.md` Process Topology / `deploy/quadlet.toml`), no hardcoded number.
- Regression test: asserts the message contains no bare cardinal + `container(s)` phrase, so a future edit can't silently reintroduce a count.
- Behavior-neutral (regex/exit-code unchanged) — gate still green, 6 tests pass.

Verified this gate has **no canonical copy in roxabi-plugins** → edited project-side directly.

## Deferred follow-up (separate PR)
The regex only matches `four|five|six containers` and can't catch the live `nine containers` in `docs/runbooks/quadlet-install.md:3`. Generalizing it to any cardinal cascades into fixing that (and other) docs → needs a dry-run sweep first, so it's intentionally out of this PR.


MickaelV0:
## Code Review (security-auditor · backend-dev · tester) + fix

**Verdict: Approve.** Message-only change verified behavior-neutral (regex/exit-code unchanged); regression test correctly targets the rule via RULES lookup and would have failed against the old hardcoded message (praise).

**Fix applied** (1 confirmed suggestion): the regression guard only matched spelled-out `one..twelve` — but the live count is **26** (digit-form), so a future regression re-stated as "26 containers" would slip through. Extended the alternation to `\d+` + a wider word range. 6 tests pass.

PR #2086, opened 2026-07-01
fix(converge): daemon-reload after quadlet-install NO_RESTART=1

MickaelV0:
Rebuild of #2083 on clean current-staging base (the auto-update bot had merged an intermediate staging state carrying an unrelated `user_store.py`/bootstrap-migration change that conflicted + polluted the diff). Net diff is now exactly the 3 converge files.

## Bug (audit 2026-06-30 — deploy provenance A)
`converge.sh` step 4 runs `make quadlet-install NO_RESTART=1`, which **skips `daemon-reload`** (the reload lives only in `deploy/quadlet-install-verify.sh`, run in the non-`NO_RESTART` else-branch — `Makefile:165-191`, `docs/CONFIGURATION.md:693`). `converge.sh` had zero `daemon-reload` calls, so Quadlet unit content installed/rendered in steps 3-4 (bot add/remove → `Secret=` churn in factory-telegram/-discord.container, image digest bumps, hardening/volume edits) was written to disk but the restarts relaunched from the **stale generated unit**, silently ignoring it.

## Fix
- `converge.sh`: explicit, unconditional `systemctl --user daemon-reload` right after step 4, before any restart.
- Tests: two `test_no_reload` cases (which locked in the buggy "never reloads") → `test_daemon_reload_called` + `test_no_unit_scoped_reload`. **11 passed** locally.
- `deploy/AGENTS.md`: corrected step 4 (falsely claimed NO_RESTART=1 includes daemon-reload).

Deploy note: M₁ auto-converges on merge → verify a converge logs `daemon-reload` + picks up unit changes.


MickaelV0:
## Code Review (security-auditor · devops · tester) + fix

**Verdict: Approve.** All 3 agents confirmed the fix is correctly sequenced (praise ×3): the unconditional `systemctl --user daemon-reload` sits after step 4 and before the drift branch, so it runs for both auth+structural drift and always precedes every restart; verified against Makefile:165-190 that NO_RESTART=1 truly skips the reload, and that `install.sh --secrets-only` never touches units.

**Fix applied** (1 confirmed suggestion): `test_daemon_reload_called` only asserted the reload appeared *somewhere* — but ordering is the whole point. Now asserts `reload_idx < restart_idx` (daemon-reload precedes restart factory-nats), so a future refactor moving the reload after the restart goes RED. 11 tests pass.

PR #2136, opened 2026-07-02
perf(ci): cut CI wall-clock ~32%, revive local hooks, stop dependabot fan-out

MickaelV0:
## Summary

20-agent audit of CI/CD (7 axes + adversarial verification of every change here) — full analysis in `artifacts/analyses/ci-cd-optimization-2026-07.claude.md`.

Three measured problems, three fixes:

**1. CI wall-clock 7m07 → ~4m50 (−32%)**
- 90s WAL-checkpoint stall in the slowest test — root cause was product-side: `AuthStore._cleanup_bare_ids` leaves a 0-row DELETE transaction open, pinning the WAL read-mark (3×30s busy_timeout in tests, stalled periodic checkpoints in prod). One-line fix.
- vitest setup was 102s of 140s cumulative (the `@phosphor-icons/react` barrel re-executed per test file) → esbuild prebundle: gate 49.6s → ~24.7s, 117/117 pass.
- `tests/integration` ran **3×** per run and `tests/e2e` **2×** → each suite now runs exactly once (`--ignore` on the coverage step; integration keeps the real-NATS docker job).
- `integration` unchained from `needs:[ci]` (ordering-only edge, job is self-contained) and its `typing_enabled` matrix dropped (behaviorally inert: flag read by zero integration tests, publisher never wired there; disabled branch is unit-covered).
- playwright `--only-shell` (the 175 MiB full Chrome was downloaded but never executed) + vestigial `portaudio19-dev` dropped (no pyaudio dep exists).

**2. Red-CI notifications — local hooks never ran**
`pre-commit install` refuses when `core.hooksPath` is set at any scope (it is, globally + locally) → `.git/hooks` held only samples and `dev-setup.sh` aborted 100% of the time. Same broken state exists in roxabi-plugins/roxabi-live/llmCLI/voiceCLI.
- `tools/install-hooks.sh` writes the `pre-commit hook-impl` dispatchers directly (+ re-chains shadowed ccc hooks); wired into dev-setup + worktree-setup (self-heal).
- Stage rebalance: pyright → pre-push (commits back to ~2-4s), new `pytest_smoke` pre-push gate, `doc_drift_bundle` → pre-push (2/18 recent reds were repeat doc-drift pushes), `lint_js` files-regex fixed (the `$`-anchored pattern matched no nested path).
- `scripts/qg` exports `UV_NO_SYNC=1` so hook-driven gates in worktrees can't corrupt the shared venv.
- De-flaked `test_publisher_to_listener_e2e_discord` (3 timeout hits on 2026-07-01 alone).

**3. Dependabot fan-out = 67.7% of ALL workflow runs (8-day window, 13,766 runs)**
`update-behind-prs` edit-poisoned every dependabot PR on every staging push → 13 zombie PRs re-CI'd ~120×/day each. Now scoped to `reviewed`-labeled PRs (+ arm-time branch update), dependabot majors/actions grouped, docker ecosystem added for base-image digest bumps, no-op `synchronize` triggers trimmed.

## Operator follow-ups (not in this PR)

One-time zombie cleanup (merge/close the 13 open dependabot PRs), `gh label create dependencies`, then the phase-2 items ranked in the analysis artifact (3-way ci job split needs a coordinated branch-protection change; merge queue; staging-red alert; image-carried converge after lot-4b).

Every run of this PR's CI exercises the new pipeline — expect the `ci` job at ~4m50.


MickaelV0:
## Code Review

Multi-domain review (security, architect, tester, backend, frontend, devops, axial — 7 fresh agents). Base: staging.

### Blockers

**issue(blocking): `make hooks-install` + CONTRIBUTING.md still call the broken `pre-commit install`** — `Makefile:406-408`, `CONTRIBUTING.md:87,93-94` — C=90
The exact bug this PR fixes (pre-commit refuses under `core.hooksPath`) stays live in the documented parallel entrypoint: a contributor following CONTRIBUTING gets zero hooks, silently. Fix: point `hooks-install` at `bash tools/install-hooks.sh` + update the CONTRIBUTING "Equivalent" block. Class: parallel-path-drift.

**issue(blocking): the WAL commit-ordering fix ships with zero regression coverage** — `src/factory/infrastructure/stores/identity/auth_store.py:64` — C=88 (tester) / 78 (backend, empirically verified the bug in a sqlite REPL)
Reverting the fix passes every existing test. Cheapest pin: `assert not store._require_db().in_transaction` right after `connect()` in the 0-deletion path of `tests/core/test_auth_store_seed_prefix.py`; stronger: second connection asserting `wal_checkpoint(TRUNCATE)` returns promptly.

### Warnings (non-blocking)

- **`|| true` on the new update-branch step swallows ALL failures, not just the documented 422** — `.github/workflows/auto-merge.yml:60` — flagged independently by 5/7 agents (C≤72). Best-effort step with update-behind-prs as backstop, so acceptable — but either scope to 422 or widen the comment. Class: bash-error-suppression.
- **install-hooks.sh reuse footgun (item 12 of the phase-2 plan)** — `tools/install-hooks.sh:534` — C=72. In a repo whose only `core.hooksPath` is the GLOBAL one, the script writes dispatchers into the machine-shared hooks dir (and the `global_dir == hooks_dir` guard then skips ccc re-chaining). Safe in THIS repo (local override verified). Guard before replicating to roxabi-plugins/roxabi-live/llmCLI/voiceCLI.
- **`pytest_smoke` has no floor against 0 collected tests** — `.claude/stack.yml` — C=55. If the 2 smoke-marked files lose the marker, every push fails with pytest exit 5. Consider a loud companion assert or accept documented coupling.
- **`vitest.config.ts` is outside the tsc program; Vitest 4 renames `deps.optimizer.web`→`client`** — C=65. Add the file to `tsconfig.node.json#include` so the future rename fails typecheck instead of silently reverting the 102s setup cost.
- **First `uv sync --package roxabi-contracts --extra testing` may be redundant** — `ci.yml:80` — C=65. `[dependency-groups] dev` already lists `roxabi-contracts[testing]`. Verify then drop the line (saves an install round-trip) or comment why it stays.

### Notes
- CI sync-order ↔ UV_NO_SYNC coupling is load-bearing, prose-only — optional fast-fail assert (`uv run ruff --version && uv run pyright --version`) after install.
- `synchronize`-drop in pr-title/dependabot-automerge is safe only while neither is a required check — one-line cross-ref comment suggested.
- Heredoc `${hooks_dir}` interpolation: defense-in-depth charset guard suggested (local-only threat model).

### Praise
- WAL fix is a correct root-cause fix (verified empirically: 0-row DELETE does open an implicit transaction; 4 connections share auth.db; hub-exclusive RW). C=90.
- `files:` regex fixes verified against the exact `[[ =~ ]]` matcher scripts/qg uses — old pattern matched zero nested paths, new one correct.
- stages ↔ run_order parity clean; `pytest_smoke` scoping verified (9 tests collected); axial review: no drift, single-module fix.

### Verdict: **Request changes** (2 blockers, both cheap: Makefile/CONTRIBUTING redirect + one regression assert)

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 2 finding(s)
**Applied via 1b1:** 2 finding(s)
**Skipped:** 5 finding(s)

### Auto-Applied
- [applied] issue: seed-prefix cleanup left implicit transaction open on zero-deleted path — regression test `test_cleanup_commits_on_zero_deleted` pins the WAL fix (falsification-gated: red with bug reintroduced) (97%)
- [applied] issue: `make hooks-install` + CONTRIBUTING still pointed at `pre-commit install`, which refuses under core.hooksPath — both now route through `tools/install-hooks.sh`

### Applied (1b1)
- [applied] issue: install-hooks.sh could clobber a GLOBAL core.hooksPath dir (dispatchers would fire in every repo) — hooks dir now resolved from `--local` scope only, repo-local hooksPath pinned when only a global exists (prerequisite for replicating the script to other repos) (72%)
- [applied] suggestion: `vitest.config.ts` added to `tsconfig.node.json#include` (65%) — **caveat:** tsconfig.node.json is referenced by no executed tsc program (typecheck = tsconfig.json → src only); wiring it surfaces 21 pre-existing errors (missing @types/node, `__dirname` in ESM configs) → follow-up material

### Skipped
- suggestion: `|| true` scoping on update-branch call (72%) — best-effort by design, update-behind-prs is the backstop
- suggestion: pytest_smoke exit-5 floor when 0 smoke tests collected (55%) — coupling documented in stack.yml
- suggestion: drop possibly-redundant first `uv sync --package contracts` in ci.yml (65%) — ~3s gain vs CI risk; verify outside this PR
- nitpick: heredoc charset guard on hooks_dir (60%) — local-only threat model
- nitpick: dependabot actions group minor/major split (35%) — volume-reduction goal

PR #2125, opened 2026-07-01
feat(dashboard): live jobs view over SSE + active-jobs registry write-path (#1772)

MickaelV0:
## Why

The active-jobs dashboard (#1800) and registry (#1796) were closed as done, but the feature was **dark end-to-end in prod**: the KV registry was provisioned yet never written to (no `open()/close()` callsite), and the browser polled `/api/bff/jobs` every 5s against an always-empty store. A 3-lens panel (architect / axial / devops) confirmed the *design* is sound (CQRS: KV current-state + `factory.job.>` trace + `factory.event.*` external), so the fix is wiring, not redesign.

This PR makes the live jobs view actually work, and swaps the browser poll for SSE.

## What

**A · write-path (`core/`)** — populate the registry from the pool run lifecycle
- `ports/active_jobs.py`: `ActiveJobsRecorder` — narrow open/close port (subset of `ActiveJobsPort`, satisfied structurally by `RegistryCoordinator`)
- `hub.py`: `active_jobs_recorder()` accessor backing `PoolContext`
- `pool_processor.py`: `open()` at run start, `close()` in the run `finally` — one entry per run (`job_id = run`), superseding the `pool._current_task` it mirrors. **Best-effort side-channel: any registry failure is swallowed so a turn is never broken.**

**B · SSE (`dashboard/` + FE)** — push instead of poll
- `dashboard/jobs_stream.py` + `routes/bff.py`: `GET /api/bff/jobs/stream` (+ stream-token), server digest-polls the registry and pushes a frame only on change — mirrors the existing pipeline stream
- `lib/api.ts` / `hooks/useJobsLive.ts` / `JobsPage.tsx`: `useQuery(refetchInterval: 5s)` → `EventSource` hook (mirrors `usePipelineRuns`)

## Validation (local, all green)

| Python | Frontend |
|---|---|
| ruff · pyright 0 · 70 pytest | biome · vitest 114 · build+dist |

New tests: `test_active_jobs_writeback.py` (open/close + **proves any registry error — incl. a `MagicMock` ctx — never breaks a turn**), `test_jobs_stream.py`. Existing `test_simple_agent`/`test_voice_command` (MagicMock ctx) stay green.

## Reviewer notes

- **Two `except Exception`** (`pool_processor`, `jobs_stream`) → auto `axial-adr-review` + `security-auditor` labels. Intentional: the registry write is a best-effort side-channel that must never break message processing, and the SSE generator must surface hub errors as a frame rather than kill the stream. Do not strip the labels.
[...]

Issue #2069, opened 2026-06-30
feat(obs): OTel raw telemetry — adapter hooks + collector JSONL + dashboard raw


MickaelV0:
## Goal

Collect structured OTel spans (`pool_id`, `job_id`, skill, model, `blob_ref`) from all NATS workers (clipool, omp, voice, image, llm) and llmcli proxy; store raw (JSONL + SQLite index) without Langfuse v1 for later post-processing and dashboard raw viewer.

**SSoT execution plan:** `artifacts/goal/otel-raw-telemetry-goal.md`

## Parent / related

- Epic #1760 (control-plane dashboard)
- Engines #1759
- Amends ADR-092 / ADR-094; new ADR-097 (otel-raw store)

## Design validated (2026-06-30)

Panel review: DevOps · Axial Drift · Product · Architect — **GO design, NO-GO implementation** until Block 0 (design artifacts) merged.

### Key decisions

- Three stores: blobstore (bytes) · turns.db (text) · otel-raw (metadata)
- OTel via `MessageLifecycleHooks` in `NatsAdapterBase` — **no** OTel SDK in `roxabi-nats` core
- New package `roxabi-otel`; optional `telemetry_attributes()` per worker CLI
- Langfuse deferred; collector exports JSONL only in v1
- Hub codec SSoT (`mint_work_envelope_fields`) **blocking** before worker OTel

## Execution order

1. Block 0 — ADR-097 + specs (design-only PR)
2. Block 1 — Hub codec SSoT
3. Blocks 2–5 — contracts, roxabi-otel, collector, factory workers
4. Blocks 6–7 — voiceCLI / imageCLI / llmCLI PRs (external repos)
5. Block 8 — Dashboard raw BFF + UI

## Acceptance (summary)

- [ ] Design artifacts merged (ADR-097, span attributes, correlation IDs, otel-raw store)
- [ ] Hub codecs propagate trace_id/job_id (no local uuid4 remint)
- [ ] Collector writes JSONL; Langfuse not required at startup
- [ ] Spans carry blob_ref pointers, never bytes/PII
- [ ] E2E: clipool job → JSONL line with coherent job_id
- [ ] Dashboard `GET /api/bff/spans` raw viewer

See full AC list in goal file.

PR #2137, opened 2026-07-02
fix(dashboard): mint root trace for operator-launched jobs (launch 502) + quiet empty-registry read

MickaelV0:
## Why

Operator "Launch OMP job" from the dashboard returned **502 Bad Gateway**:
```
1 validation error for DashboardJobsLaunchResponse
accepted  Field required [input_value={'error': 'internal_error'}]
```
Reproduced live on M₁ + traced to the hub.

## Root cause (regression from #2069)

`handle_jobs_launch` calls `mint_work_envelope_fields(pool_id=…)` with **no `trace_id`** and **no ambient `TraceContext`** — an operator launch is a fresh root entry point, unlike hub work-path codecs (which run inside an inbound-established trace). #2069 made `trace_id` required → `mint_work_envelope_fields` raised `ValueError` → the RPC replied `{'error':'internal_error'}` → the dashboard BFF failed to parse that as `DashboardJobsLaunchResponse` → **502**.

**Why CI was green:** #2069 also added an **autouse** `conftest` fixture that sets a default `TraceContext` for every test — so `handle_jobs_launch` tests passed with an ambient trace that **prod never has**. Classic fake-masks-reality.

## Fixes

1. **`handle_jobs_launch`** — mint a root trace explicitly: `trace_id=TraceContext.get_trace_id() or TraceContext.generate()` (idiom from `middleware_guards`). + regression test marked `@pytest.mark.no_default_trace` that reproduces the **no-ambient-context** prod condition (fails without the fix).
2. **`active_jobs_kv.list_all()`** — was logging `NoKeysError` (empty bucket) as `ERROR` + traceback every ~5s on the dashboard poll. Empty bucket = no active jobs, not a failure → catch `NoKeysError` → return `[]` silently.

## Validation (local)
ruff · pyright 0 · **32 pytest** (dashboard-jobs RPC incl. the no-ambient-trace repro + active-jobs KV).

Regression from #2069. Refs #2125 (jobs live view).

### DIFF
--- a/src/factory/bootstrap/factory/dashboard_jobs_rpc.py
+++ b/src/factory/bootstrap/factory/dashboard_jobs_rpc.py
@@ -61,11 +61,12 @@ async def handle_jobs_launch(hub: Hub, nc: NATS, payload: dict[str, Any]) -> dic
         req.pool_id
         or RoutingKey(Platform.WEB, _WEB_BOT, f"agent:{req.agent}").to_pool_id()
     )
-    # Operator-initiated launch is a fresh root entry point — unlike hub
-    # work-path codecs it has no ambient TraceContext, so mint a root trace
-    # explicitly (mint_work_envelope_fields raises without one, #2069).
+    # Operator-initiated launch is a fresh root entry point — it runs in its own
+    # dashboard-RPC subscription task that TraceMiddleware never touches, so it
+    # always mints a NEW root trace (like inbound), never reuses ambient context.
+    # mint_work_envelope_fields raises without a trace_id (#2069).
     fields = mint_work_envelope_fields(
-        trace_id=TraceContext.get_trace_id() or TraceContext.generate(),
+        trace_id=TraceContext.generate(),
         pool_id=pool_id,
     )
     job_id = fields.job_id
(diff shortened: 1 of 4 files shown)


MickaelV0:
## Code Review — #2137 (security · backend · tester)

**Verdict: Approve with comments** — no blockers. 3 quality/coverage findings, all applied (ec649318).

### Applied
1. **trace_id resolution** (backend, 70%) — `get_trace_id() or generate()`: the `get_trace_id()` branch is dead in prod (the dashboard-RPC task is never touched by TraceMiddleware) and a launch is a NEW root by definition → unconditional `TraceContext.generate()`.
2. **list_all coverage + latent tautology** (tester, 90%) — the `NoKeysError → []` fix had no test, and asserting `== []` alone is tautological (the generic `except nats.errors.Error` handler also returns `[]`). Added `TestListAll` asserting `[]` **and** `log.exception` not called (proves the dedicated branch is taken).
3. **launch-test assertion** (tester, 55%) — parse the published `trace_id` as a UUID instead of a raw-bytes substring check.

### Verified / praise
- trace_id has **no client-controlled input**; matches `control_trace_id`/`wire_trace_id_hex` precedent (security, 92%).
- `NoKeysError` catch is **empty-only** — verified against nats-py source (raised only on empty key list, never on ACL/permission failures) → never masks real failures; clause ordering correct (security 85% / backend 95%).
- The `no_default_trace` regression test **genuinely reproduces prod** (pre-fix `ValueError` propagates uncaught) — not vacuous (tester, 90%).
- **502 root cause fully addressed** within scope; no second downstream failure; outer `_wrap` catches residuals (backend, 80%).

Local: ruff · pyright 0 · **33 pytest**.

PR #2150, opened 2026-07-02
chore: promote staging — drop decision presentation protocol

MickaelV0:
Promote staging to main.

Includes drop of the decision presentation protocol (DP(A|B|C)) across plugin skills — replaced with direct present choice / ask user instructions.

No new gate format yet; blank slate for future redesign.


MickaelV0:
Superseded: DP drop landed directly on main (56449777)

PR #2176, opened 2026-07-02
perf(qg): honor files: filters at CI stage (PR diffs, fail-open)

MickaelV0:
## Mechanism

`scripts/qg` previously bypassed gates' `files:` regex whenever `stage == ci`, so every gate ran on every CI run. Now:

- `ci.yml` exports `QG_DIFF_RANGE: ${{ github.event_name == 'pull_request' && format('origin/{0}...HEAD', github.base_ref) || '' }}` to the qg step — **set only on pull_request events** (checkout uses `fetch-depth: 0`, so `origin/<base>` exists).
- When `QG_DIFF_RANGE` is non-empty at stage `ci`, qg computes `git diff --name-only <range>` (triple-dot = merge-base diff; **deletions included deliberately** — deleted files must trigger gates) and sets `CI_FILTER_ACTIVE=1`. Gates with a `files:` pattern skip unless a changed file matches.
- `CI_FILTER_ACTIVE` is the sentinel separating **"diff unavailable"** (fail open, run everything) from **"diff empty"** (filter normally: filtered gates skip, unfiltered gates run).
- pre-commit / pre-push semantics are byte-for-byte unchanged.

Saves ~56s on non-frontend PRs (~53% of PRs): `dashboard_build` (~30s), `dashboard_unit_test` (~25s vitest), plus `lint_js`.

## Fail-open guarantees

| Event / condition | `QG_DIFF_RANGE` | Behavior |
|---|---|---|
| `pull_request`, diff resolves | `origin/<base>...HEAD` | `files:` filters honored against merge-base diff |
| `pull_request`, diff resolves, **empty diff** | set | filtered gates skip (nothing matched), unfiltered gates run |
| `pull_request`, **git diff fails** (bad ref, shallow clone) | set | `::warning::` printed, **ALL gates run** (fail-open, never fail-closed) |
| `push` to staging/main | empty | ALL gates run (previous behavior) |
| `workflow_dispatch` | empty | ALL gates run |
| future `merge_group` | empty (expression only matches `pull_request`) | ALL gates run |
| local `scripts/qg run --stage ci` (no env) | unset | ALL gates run |

## New filters (this PR only)

`dashboard_build` and `dashboard_dist_assert` gain the exact filter `dashboard_unit_test` already had:
`^(apps/dashboard/|packages/shared/|brand/)|^(bun\.lock|package\.json)$`

- `dashboard_dist_assert` **must** stay identical to `dashboard_build`'s filter (it asserts the build output exists — running it when build was skipped is a guaranteed red). A comment in stack.yml pins this invariant.
- Dashboard's only workspace dep is `@roxabi-factory/shared` (`packages/shared/`) — covered. Root `package.json`/`bun.lock` (toolchain bumps) — covered.
- No other gate gains a filter in this PR (each new filter is a new silent-skip risk).

## Audit of existing filters (now effective in CI for the first time)

| Gate | Pattern | Verdict |
|---|---|---|
| `lint_js` | `^(apps/\|packages/\|brand/)\|^(biome\.json\|package\.json\|bun\.lock)$` | OK — over-broad in the safe direction (python `packages/*` changes also trigger it). `$`-anchoring bug already fixed (see stack.yml comment). |
| `dashboard_unit_test` | `^(apps/dashboard/\|packages/shared/\|brand/)\|^(bun\.lock\|package\.json)$` | OK — sole workspace dep (`packages/shared`) and toolchain bumps covered. |
| `codes_sync` | `^packages/roxabi-contracts/(src/roxabi_contracts/errors\.py\|docs/error-codes\.md\|scripts/check_codes_sync\.py)$` | OK — narrow, but the check's only inputs are exactly those three files. |

⚠️ **Flagged (not changed):** none of the filters match `.claude/stack.yml`, `scripts/qg`, or `.github/workflows/ci.yml` themselves — a PR touching only QG/CI infra (like this one) skips the filtered frontend gates. Acceptable (those gates don't consume the runner), but worth knowing: if a filtered gate's *script/config* moves into an unmatched path, its filter should be widened in the same PR.

## Tests

`tests/scripts/test_qg.sh` section E (fixture git repo): non-matching diff → filtered skips / unfiltered runs; matching diff → runs; no `QG_DIFF_RANGE` → fail-open; invalid range → fail-open + `::warning::`; empty diff → filtered skips without warning. 11/11 pass locally.

## Merge ordering

⚠️ Sibling PR `perf/ci-job-split` touches the same `ci.yml` region (Quality gates step) and **must merge AFTER this one** (rebase on the conflict).


MickaelV0:
## Code Review

3 fresh agents (security-auditor, devops, tester). Verdict: **Request changes** — all actionable findings applied in-PR (see follow-up comment).

### Blockers
- issue(blocking) [devops, 90%]: dashboard_build's new `files:` filter silently disables the "Dashboard visual e2e" step for PRs that can regress it — the e2e fixture `pytest.skip`s when `apps/dashboard/dist/index.html` is missing, and the SPA is served by `src/factory/adapters/web/` (untracked by the filter). PRs touching the web adapter or `tests/e2e/dashboard/` would merge unverified. → extend both filters.
- issue(blocking) [security 95% + devops 85% + tester 70% — 3 agents]: `git diff --name-only` emits C-quoted paths under default `core.quotePath` (`"apps/dashboard/caf\303\251.ts"`) → a non-ASCII/control-char filename silently skips every filtered gate (fail-open→silent-skip inversion, fork-reachable via filenames). → NUL-delimited `-z` enumeration.
- suggestion(blocking) [security 90% + devops 85%]: rename detection lists only the destination — moving a file OUT of a filtered tree never triggers the gate (contradicts the "deletions covered" comment). → `--no-renames`.
- suggestion(blocking) [tester, 95%]: the "deletions trigger gates" invariant has no test — an `ACMR`-filter regression (the exact idiom used 250 lines up in the pre-commit path) would go green. → E6 deletion case. Class: parallel-path-drift (scripts/qg pre-commit diff vs CI diff).

### Non-blocking (applied)
- parallel-path-drift [security 85% + devops 90%]: filter regex triplicated (unit_test/build/dist_assert) with prose-only identity invariant; YAML anchors REFUTED as fix (yq returns unresolved alias text without explode()). → test_qg.sh identity assertion dashboard_build ≡ dashboard_dist_assert.
- bash-error-suppression [security 80% + devops 80%]: `2>/dev/null` discards git's real diagnostic; the fail-open warning guesses. → capture stderr into the ::warning::.
- tester 80%: no test that pre-commit/pre-push ignore QG_DIFF_RANGE (the stage guard lives in 2 places; a leak would silently clobber staged-diff semantics).
- tester nitpicks: E2 tests unset but production emits set-but-empty (+ env leak into the else branch); E3 greps bare `::warning::`; E5 can't distinguish match from fail-open; WORK3 tmpdir leaks on abort. devops nitpick: fixture commits inherit global gpgsign/hooks.
- devops 80%: doc_drift_bundle short-circuits before the files:/enabled checks — guard comment pinning the invariant.

### Noted, no action
- security thought [70%]: `files:` on a ci gate is now a soft gate-evasion channel in PR-controlled stack.yml (equivalent power already existed via `enabled: false`); review of stack.yml diffs remains the boundary.
- devops thought [90%]: stale-merge-ref/stacked-PR gap is covered by run-ALL on push events — invariant: never add filtering to push events.
- devops thought [85%]: codes_sync + lint_js filters audited complete today; their `files:` ↔ script-scope coupling is now load-bearing at CI.

### Praise
- Fail-open architecture (CI_FILTER_ACTIVE sentinel, env:-block injection, mapfile-without-phantom-element) called out independently by security + devops; tester confirmed section E drives the real production path against a real git fixture (no tautology).

MickaelV0:
## Review Fixes Applied

Counts: auto-applied 0 / applied 13 / documented 3 / no-action 0 — commit b457767e5. `tests/scripts/test_qg.sh`: 15/15 pass.

### Blockers
- [applied] e2e-input filter gap: `dashboard_build` + `dashboard_dist_assert` filters extended with `src/factory/adapters/web/` and `tests/e2e/dashboard/` (kept byte-identical; `dashboard_unit_test` untouched — vitest doesn't run e2e). Comment rewritten to state the e2e `pytest.skip` coupling.
- [applied] core.quotePath silent-skip: CI diff now `git diff --name-only -z --no-renames` + `mapfile -d ''`. Verified load-bearing: without `-z`, `match/café.txt` enumerates as `"match/caf\303\251.txt"` and matches nothing (E5 now pins this with a real non-ASCII fixture path).
- [applied] rename escape: `--no-renames` — a move out of a filtered tree surfaces as D+A; the deletions comment updated to state both invariants truthfully.
- [applied] deletion coverage test: E6 commits a matching file, `git rm`s it, and asserts the filtered gate still runs on the deletion-only range (guards against an ACMR regression drifting in from the staged pre-commit idiom).

### Non-blocking
- [applied] triplicated-regex drift: section F asserts `dashboard_build.files` ≡ `dashboard_dist_assert.files` byte-identically against the REAL `.claude/stack.yml` via yq.
- [applied] stderr suppression: git's stderr is captured and embedded in the `::warning::` (newlines flattened for the annotation); E3 asserts both the exact fail-open message and a `fatal:` fragment.
- [applied] stage-guard leak test: E7 runs `--stage pre-commit` with `QG_DIFF_RANGE` exported pointing at a matching range while the staged file does NOT match — filtered gate skips (staged diff wins).
- [applied] E2 split: E2a uses `env -u QG_DIFF_RANGE` (true unset, no env leak), E2b uses `QG_DIFF_RANGE=""` (the exact production push/workflow_dispatch value).
- [applied] E5 ambiguity: negative assertion that no fail-open warning appeared (genuine filter match proven).
- [applied] fixture robustness: `commit.gpgsign false` + `core.hooksPath` isolation on WORK2/WORK3; EXIT trap replaced by a `cleanup()` covering WORK/WORK2/WORK3 (`${VAR:+...}`, set -u safe).
- [applied] pre-commit/pre-push parity: `changed_files()` gets the same `-z` + `mapfile -d ''` quoting fix; `--diff-filter=ACMR` deliberately kept on the staged branch and the divergence is now documented in-line.
- [applied] doc_drift_bundle: guard comment pinning the deliberate short-circuit before enabled/stage/files checks (a future `files:` there would be dead config).
- [applied] test header: Verifies list updated for E (quotePath, deletions, stage guard, set-but-empty) and new section F.

### Documented (no code change)
- [documented] `files:` as soft gate-evasion channel in PR-controlled stack.yml — equivalent power already existed via `enabled: false`; stack.yml diff review remains the boundary.
- [documented] stale-merge-ref/stacked-PR gap — covered by run-ALL on push events; invariant: never add filtering to push events (ci.yml comment already states this).
- [documented] codes_sync + lint_js `files:` ↔ script-scope coupling now load-bearing at CI — audited complete; no change needed.

Verification: `bash tests/scripts/test_qg.sh` 15/15; smoke `QG_DIFF_RANGE=origin/staging...HEAD scripts/qg run --stage ci lint_js` → correctly skipped (PR diff has no frontend files); bad-range smoke shows the warning carrying git's real stderr and running all gates.

PR #2177, opened 2026-07-02
perf(ci): split ci job into 3 parallel jobs (aggregate check stays 'ci')

MickaelV0:
## What

Splits the monolithic `ci` job (~4m50 serial) into 3 parallel jobs, plus an aggregate gate:

| Job | Content | Installs |
|---|---|---|
| `gates` | `scripts/qg run --stage ci` + 3 gate self-tests | uv, bun, node 24, yq, deps (both `uv sync`s, root dev LAST) — **no nats-server/nk** |
| `tests` | dashboard build + visual e2e + factory unit coverage | uv, bun, nats-server, nk, playwright headless-shell — **no yq/node** |
| `package-coverage` | roxabi_nats / roxabi_contracts / roxabi_obs coverage | uv, nats-server — **no nk/bun/yq/node** |
| `ci` | aggregate (`needs:` all three, `if: always()`) | none |

Expected wall-clock: **~2m45** (from the measured audit) vs ~4m50 today. `integration` and `docker-build` jobs are untouched.

## Why the aggregate job is named `ci`

Staging branch protection requires status contexts exactly `["ci", "trufflehog"]` (strict). Keeping a final job whose check context is exactly `ci` means this split needs **zero branch-protection changes** — a renamed/removed `ci` context would deadlock auto-merge repo-wide.

`if: always()` on the aggregate is mandatory: without it, a failed needed job leaves the aggregate **skipped**, and a skipped required check deadlocks merges exactly like a missing one. The aggregate step fails unless every needed job's result is `success` (so `failure`, `cancelled`, and `skipped` all block).

## Install trims — evidence

- **gates drops nats-server/nk**: no ci-stage gate script references either binary (grep over `tools/` + `scripts/`); `acl_authconf_drift` runs `factory-acl genkeys --template-only`, which fabricates placeholder pubkeys (`scripts/gen_nkeys.py::_cmd_genkeys`) without invoking `nk`.
- **tests keeps nats-server/nk**: `tests/factories/nats_server.py` spawns the real binary (used by `tests/nats/`, `tests/bootstrap/test_embedded_nats.py`); `tests/scripts/test_nk.py` + `test_genkeys_modes.py` exercise `nk`.
- **tests builds the dashboard explicitly**: the visual e2e fixture `pytest.skip()`s when `apps/dashboard/dist/index.html` is missing — in the monolithic job the `dashboard_build` gate produced it as a side effect; the explicit `bun run build:dashboard` step prevents silent skips.
- **package-coverage keeps nats-server only**: `packages/roxabi-nats/tests/conftest.py` spawns a real nats-server and auto-skips without it (the 70% floor must keep measuring real-server coverage); zero `nk`/bun references in `packages/*/tests`.
- **checkout depth**: `fetch-depth: 0` kept only in `gates` (`check_debt_expiry` reads git dates); the pytest suites stub git (`tests/deploy/test_converge_nats_restart.py`), so `tests`/`package-coverage` use default depth.

## Verification

- YAML validated with PyYAML; job graph confirmed (`ci` needs `[gates, tests, package-coverage]`, `if: always()`).
- GitHub Actions cannot be run locally — **this PR's own CI run is the real test** of the split.

## Merge ordering

⚠️ A sibling PR (`perf/qg-ci-files-filters`) touches the same region of `ci.yml`. This PR should merge **after** it and absorb staging (re-merge staging into this branch, resolve, re-run CI) before going in.


MickaelV0:
**Round 2 pushed** — two changes since the first run:

1. **Root cause of the `package-coverage` failure** (16 caplog failures in `test_version_check.py`): a latent boot-window bug in `roxabi_nats/_version_check.py`, not a missing install. The log rate limiter used `0.0` as the "never logged" sentinel; `time.monotonic()` counts from boot on Linux, so on a freshly booted runner VM (`monotonic < 60 s`) `now - 0.0 < _LOG_INTERVAL_S` silenced the very **first** drop log. The monolithic job masked it by reaching this step ~5 min after boot; the split job reaches it ~40 s after boot. Fixed by keying "never logged" off absence of prior state + regression test pinning a young clock (`2a9057814`). Reproduced locally on the old code: young clock → 0 error logs fired.

2. **Absorbed staging** (`36fcb690f`): #2176's `QG_DIFF_RANGE` env block auto-merged onto the "Quality gates (stack.yml)" step inside the `gates` job (which carries the `fetch-depth: 0` it depends on), plus the #2182/#2183 Dockerfile python pin fix (docker-build should be green now).

MickaelV0:
## Code Review

2 fresh agents (devops, security-auditor). Verdict: **Approve with comments** — 2 non-blocking hygiene fixes applied in-PR (commit 0325cdf38).

### Non-blocking (applied)
- suggestion [devops, 85%]: nats-server installer duplicated across tests + package-coverage jobs with per-step `NATS_VERSION` — a future bump edited in one job leaves the other on the old binary (coverage floors then measure against different servers). → hoisted to a single workflow-level `env:` declaration.
- nitpick [devops 80% + security 85%]: `oven-sh/setup-bun@v2` was the only tag-pinned action in the file and the split doubled it to 2 call sites. → SHA-pinned both (`0c5077e5 # v2`), matching file convention.

### Noted, no action
- thought [devops, 75%]: wall-time halves but billed runner minutes ~double (3 checkouts/syncs, dashboard built twice — the tests-job explicit build is correct: e2e's dist fixture lives on a different runner now). Accepted: repo is public (free minutes); artifact hand-off would re-serialize the graph.
- note [security, 90%]: the aggregate `ci` needs: covers exactly the old required-check scope (gates/tests/package-coverage) — integration/docker-build were never part of the required context and still aren't; the in-file comment documents it.
- note [security, 90%]: nats/nk/yq installers are TOFU (checksums fetched from the same release) — pre-existing pattern, replicated verbatim.

### Praise
- Aggregate `ci` fan-in is deadlock-proof by construction: `if: always()` + fail on ANY `needs.*.result != 'success'` (stricter than the common `contains(failure)` idiom — skipped/cancelled also fail), results via `env:` (no expression-in-script injection), job id keeps context "ci" → zero branch-protection change. Empirically proven in this PR: the package-coverage red run concluded `ci=failure`, not skipped.
- The package-coverage first-red fix is a root-cause product fix, not a CI workaround: `_version_check.py` used `monotonic 0.0` as "never logged" sentinel → on a runner booted <60s the FIRST drop log was silenced (also a latent prod bug on fresh containers); fixed by keying on absence + regression test pinning the boot window.
- 1:1 step mapping verified against the base monolith (17/17 steps, flags byte-identical, comments moved alongside); `uv sync` order invariant preserved in gates; dependency-pruning claims verified against the tree (no ci gate shells nats-server/nk; roxabi-nats tests skipif on missing nats-server → installing it in package-coverage protects the 70% floor from skip-inflation).

### Timings (green run)
gates 61s · tests 2m33 · package-coverage 56s · aggregate 2s → required-check wall ~2m45 (was ~4m50 serial).

MickaelV0:
## Review Fixes Applied

**Applied:** 2 · **Documented/no-action:** 3 · **Deferred:** 0 · **Failed:** 0

### Applied
- [applied] suggestion: NATS_VERSION hoisted to workflow-level env (single declaration site) — .github/workflows/ci.yml
- [applied] nitpick: setup-bun SHA-pinned at both call sites (`0c5077e5 # v2`)

### No action (documented in review)
- runner-minutes trade-off accepted (public repo)
- aggregate scope mirrors the old required check (in-file comment)
- TOFU checksum pattern pre-existing

PR #2251, opened 2026-07-04
test(nats): matrix-driven live ACL round-trip — close postmortem P0 gap

### LINKED ISSUE #2247: test(nats): matrix-driven live ACL round-trip — close the postmortem's P0 gap
**Origin:** postmortem action item, `docs/history/nats-acl-inbox-case-postmortem.md` — the P0 fix ("real NATS server + ephemeral nkeys + cross allow/deny assertions") marked "skipped" after the 2026-04-27 incident, surfaced during epic #2193 triage and validated by a 5-agent panel review (architect/product-lead/axial-drift/security/devops, 2026-07-04, unanimous on priority).

**Do not build from scratch** — `tests/scripts/test_parity_e2e.py` already boots an ephemeral `nats-server` with the real rendered `auth.conf` + ephemeral nkeys, and asserts allow/deny for a handful of hardcoded identities (hub, voice-tts, clipool) plus retired-identity rejection. CI already installs `nats-server` + `nk`.

**Gap**: coverage is hand-picked, not matrix-driven, and there is no live cross-identity request/reply round-trip through the real dynamic inbox — the exact mechanism that broke in April (`_INBOX.hub.<NUID>`, case-sensitive, authorized only via `allow_responses`). A static (identity, subject) string-match would NOT reproduce April; only a live round-trip against the broker's real accept/reject does.

**Scope**:
1. Parametrize the existing live-server tests over every active `(identity, subject)` pair in `deploy/nats/acl-matrix.json`, and every entry in `request_reply_flows`, so new identities/subjects are auto-covered.
2. Add a live request/reply round-trip per `request_reply_flows` entry: requester publishes with a reply-to, responder replies to the broker-minted dynamic inbox, assert the reply is actually delivered (positive) and that non-flow pairs are denied (negative).
3. Skip `status:retired` identities.
4. Keep the existing static string-compare tests (`test_acl_canonical_llm.py`, `test_acl_gh_helper.py`) as a fast structural pre-check layer — do not fold them in carelessly; cross-reference instead. **Do not add a 3rd per-identity test file** — this must be one data-driven harness (axial-drift finding: 2 per-identity files already exist, a 3rd institutionalizes N×M test-axis drift).
5. CI: the job must **hard-fail**, not skip, when `nats-server`/`nk` are absent — a skipped ACL gate is a false-green (existing `requires_nats_server` skip guard needs the same scrutiny).
6. Security: ephemeral nkeys generated fresh per run into runner tmpfs only, never committed/cached as artifacts, broker bound to `127.0.0.1` on an ephemeral port with a per-run tmp datadir, never read a real `~/.roxabi` seed path.

**Not filed under epic #2193** (doc-strategy) — this is CI/test-hardening, standalone.

**Acceptance**:
- [ ] Live ACL test parametrized over the full active identity+subject matrix from acl-matrix.json
- [ ] Live cross-identity request/reply round-trip test reproduces April's failure class (denies a case-mismatched or non-flow reply-to)
- [ ] `make test-acl` alias runs this file
- [ ] CI hard-fails (not skips) when nats-server/nk absent
- [ ] Existing static tests kept as fast pre-check, cross-referenced not duplicated

Ref: validated by architect/product-lead/axial-drift/security/devops panel review, 2026-07-04 (unanimous "highest priority of the four" #2202/2203/2204 follow-ups).

MickaelV0:
## Summary
- Parametrizes `tests/scripts/test_parity_e2e.py`'s live-`nats-server` harness over every active `(identity, subject)` pair in `deploy/nats/acl-matrix.json` (via `effective_grants()`) and every `request_reply_flows` entry, replacing the old hand-picked coverage (hub/voice-tts/clipool only) with matrix-driven coverage that auto-extends when the matrix changes — closing the postmortem's P0 gap (`docs/history/nats-acl-inbox-case-postmortem.md`).
- Adds a live positive request/reply round-trip per flow (real reply delivered through the broker's dynamic inbox) and a live one-way negative check per flow (non-flow-paired identity's inbox publish is actually denied — a real captured `"permissions violation"`, not a timeout). An earlier round-trip/case-mismatch/timeout design for the negative case was rejected during spec review: a case-mutated publish times out regardless of ACL outcome, so that assertion would pass unconditionally and prove nothing.
- Adds a drift-detection test (`test_acl_matrix_v3_current_in_sync`) comparing `acl-matrix.json` against the `v3-current.json` fixture, a stricter CI hard-fail guard (`GITHUB_ACTIONS=true` + missing `nats-server`/`nk`/`nats-py` now fails collection instead of silently skipping — verified both directions manually), and a `make test-acl` alias.
- Cross-references added to the two existing static per-identity test files (`test_acl_canonical_llm.py`, `test_acl_gh_helper.py`) — no 3rd per-identity test file created, per the issue's explicit constraint. No production code touched: `acl-matrix.json`, `scripts/_renderer.py`, `scripts/_effective.py`, `factory.cli.ops` are all reused as-is.

## Live gap found during implementation (not fixed here — out of scope)

While implementing the matrix-driven deny-probes, 3 of 68 new test nodes failed deterministically (not flaky — confirmed by isolated reproduction against a real `nats-server`, including a check after `nc.drain()`): `dashboard-reader` (publish), `gh-helper` (subscribe), `ingress` (subscribe).

Root cause: **`nats-server` 2.10.22 treats an explicit empty allow list (`publish: { allow: [] }` / `subscribe: { allow: [] }`) as unrestricted, not deny-all.** A user's `permissions` block replaces `default_permissions` wholesale rather than merging with it, and an empty allow list leaves that direction's allow-set nil — so the top-level `deny: [">"]` fallback rendered by `scripts/_renderer.py` never applies for that direction. Practically: today, `dashboard-reader` can publish to any subject, and `gh-helper`/`ingress` can subscribe to any subject, contrary to their documented least-privilege intent (e.g. `test_acl_gh_helper.py` explicitly asserts "gh-helper must have no subscribe permissions").

This is squarely what a live round-trip test is supposed to catch (vs. the static string-match layer, which only checks the JSON matrix's *intent*, not the broker's *actual enforcement*) — the test is working as designed. Disposition in this PR: `pytest.mark.xfail(strict=True)`, computed from `EFFECTIVE_GRANTS` (not hardcoded identity names — any future identity with an empty allow-list direction is automatically caught the same way). `strict=True` means the moment the renderer bug is fixed, the probe starts passing, the suite fails on the unexpected XPASS, and the marker must be removed by whoever lands that fix.

**Recommend filing a follow-up issue** to fix `scripts/_renderer.py` (e.g. emit an explicit `deny: [">"]` for a direction whose intended allow-list is empty). Not filed here per the "no issue creation without user approval" rule, and out of scope for this test-only issue regardless. Severity is a human call — exploiting this requires a valid nkey for one of the 3 affected identities (defense-in-depth gap, not remote-unauth), so I'm not stamping a priority on it myself.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #2247: matrix-driven live ACL round-trip — close the postmortem's P0 gap | OPEN |
| Frame | [2247-matrix-driven-live-acl-round-trip-frame.mdx](artifacts/frames/2247-matrix-driven-live-acl-round-trip-frame.mdx) | Present |
| Spec | [2247-matrix-driven-live-acl-round-trip-spec.mdx](artifacts/specs/2247-matrix-driven-live-acl-round-trip-spec.mdx) | Present |
| Plan | [2247-matrix-driven-live-acl-round-trip-plan.mdx](artifacts/plans/2247-matrix-driven-live-acl-round-trip-plan.mdx) | Present |
| Implementation | 6 commits on `feat/2247-matrix-driven-live-acl-round-trip` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (68 new/parametrized nodes: 65 passed, 3 xfailed strict) | Passed |

## Test Plan
- [x] `make test-acl` runs exactly `tests/scripts/test_parity_e2e.py` (verified directly)
- [x] `uv run pytest tests/scripts/test_parity_e2e.py -v` — 65 passed, 3 xfailed (strict) locally with `nats-server`/`nk` installed
- [x] `GITHUB_ACTIONS=true PATH=/usr/bin:/bin uv run pytest tests/scripts/test_parity_e2e.py` — hard collection failure (exit 1), not a skip (verified directly)
- [x] Same command without `GITHUB_ACTIONS` set — 68 skipped gracefully, exit 0 (verified directly)
- [x] `git grep -n "test_parity_e2e" tests/scripts/test_acl_*.py` — cross-reference present in both static files
- [x] Full repo suite: `uv run pytest -q` — 6548 passed, 26 skipped, 4 xfailed (1 pre-existing + 3 new), 0 failed
- [x] Full repo `ruff format .` / `ruff check .` / `pyright` — clean

## SC → Test Matrix

| SC | Test(s) | Status |
|----|---------|--------|
| SC1: every active (identity, subject) pair exercised live | `test_parity_e2e.py :: test_acl_matrix_publish[*]`, `test_acl_matrix_subscribe[*]` | ⚠ NO FALSIFY — e2e (65 passed, 3 xfailed strict — see gap above) |
| SC2: live positive round-trip + negative non-flow-pair denial per flow | `test_parity_e2e.py :: test_flow_round_trip_positive[*]`, `test_flow_round_trip_denies_non_flow_pair[*]` | ⚠ NO FALSIFY — e2e (18/18 passed) |
| SC3: acl-matrix.json ↔ v3-current.json drift guard | `test_parity_e2e.py :: test_acl_matrix_v3_current_in_sync` | ⚠ NO FALSIFY — e2e (passed) |
| SC4: retired identities excluded + non-empty-list guard | `test_parity_e2e.py :: test_acl_matrix_active_identities_nonempty`, pre-existing `test_retired_identity_connect_rejected` | ⚠ NO FALSIFY — e2e (passed) |
| SC5: `make test-acl` exists, runs exactly this file | — | ✓ proven (manual: `make test-acl` executed directly, 65 passed/3 xfailed) |
| SC6: CI hard-fails (not skips) on missing binaries under `GITHUB_ACTIONS=true`; skips gracefully otherwise | — | ✓ proven (manual: both directions executed directly per spec's own S3 check command) |
| SC7: static test files functionally unchanged, only cross-ref comment added | — | ✓ proven (structural: `git diff` shows docstring-only changes) |
| SC8: no 3rd per-identity static ACL test file added | — | ✓ proven (structural: `git status` shows no new files) |
| SC9: ephemeral nkeys fresh per run in pytest tmpfs, seed files `chmod(0o600)` | `rendered_auth_conf` fixture (pre-existing + chmod addition); implicitly proven by every live-server test's successful `_read_seed()`-gated connection | ✓ proven |
| SC10: `uv run pytest tests/scripts/test_parity_e2e.py -v` passes locally | full file run | ✓ proven (65 passed, 3 xfailed) |

**Falsification Gate note:** this PR creates zero new production/source files — only test files, the Makefile alias, and docs artifacts are touched. Every new/changed test in the matrix above exercises a real `nats-server` subprocess against existing, unmodified production ACL config, matching this file's own filename (`test_parity_e2e.py`) and its pre-existing tests' e2e classification. The Falsification Gate's source-stash mechanism has nothing to stash (no new `src/` changes), so all rows are marked e2e-exempt per the gate's own carve-out, consistent with how this file's pre-existing tests have always been treated.

Closes #2247

---


MickaelV0:
## Code Review

Reviewed by 5 domain agents (security-auditor, architect, product-lead, tester, devops), each reading the full diff and full changed files independently — several claims empirically re-verified (live nats-server reproductions, local `pytest` runs, CI job topology traced) rather than trusted from the PR description.

**Verdict: Request changes** — one process-level blocking gap (below), no code-correctness blockers. The test implementation itself is assessed by 4/5 agents as sound and ready to merge on its own technical merits.

---

### Blockers

**todo (blocking): live production ACL gap discovered by this PR has no filed tracking issue or assigned priority**
`tests/scripts/test_parity_e2e.py:110-163` · `deploy/nats/acl-matrix.json`
— product-lead (85%), corroborated by security-auditor (`issue`, 90%) and architect (`note`, 78%)

N2/N3's exhaustive matrix-driven deny-probes found that `nats-server` 2.10.22 treats an empty `allow: []` on either direction as **unrestricted, not deny-all** — a real, live production gap on M1 today, affecting `dashboard-reader` (publish), `gh-helper` (subscribe), `ingress` (subscribe). Security-auditor independently reproduced this against a live broker (rendered the real `acl-matrix.json`, confirmed an empty-allow identity could subscribe to an ungranted subject with zero violation). This directly contradicts an existing documented invariant (`test_acl_gh_helper.py`'s own docstring claims "no subscribe permissions").

The scope call to *not fix* `scripts/_renderer.py` in this PR is correct (out of scope per the approved spec, separate blast radius, needs its own review). The disposition — `xfail(strict=True)` computed dynamically from `EFFECTIVE_GRANTS`, documented in the test docstring, spec Design Note, and PR description — is sound engineering (self-healing: XPASS fails the suite the moment the renderer is fixed, forcing marker removal; auto-covers future empty-allow identities). What's missing is a **durable, prioritized, tracked artifact**: right now the only record of a live least-privilege break is PR-description prose and an xfail reason string — exactly the "found once, then forgotten" failure mode the originating postmortem exists to prevent (see prior art: `sweep_candidates.py honors native edges only`, `prose "#n" ≠ Closes` in this repo's own memory).

Security-auditor's severity calibration: **Medium generally, High for `ingress`** (externally-facing intake path) — not remotely exploitable on its own (requires possession of the identity's 0o600-permissioned nkey seed), so not an emergency, but a real defense-in-depth gap that should be prioritized rather than left to institutional memory.

Solutions:
1. File a tracking issue against `scripts/_renderer.py` ("renderer must emit explicit `deny: [\">\"]` when an intended allow-list is empty"), sized S, recommended priority P1/Medium (High for `ingress`), and reference it from both the xfail `reason` strings and this PR's description.
2. At minimum, do not consider the originating postmortem's P0 gap fully closed until that issue exists and is prioritized by a human.

**Not actioned in this run**: filing a GitHub issue unilaterally is outside this pipeline's authority (Safety Rule requires human approval before `gh issue create`). Flagging here for the human operator to action.

*Note on dissent*: 2 of the 3 agents who flagged this (security-auditor, architect) framed it as non-blocking/advisory rather than merge-blocking, focused more on severity-communication than on process/tracking. Only product-lead framed it explicitly as blocking. Given 3-way independent corroboration of the underlying fact and the direct relevance to what #2247's originating postmortem was trying to prevent, the blocking classification is upheld — but this is a closer call than a typical code-correctness blocker, and is presented transparently as such.

---

### Warnings

**suggestion: allow-probe direction may inherit `_probe`'s insufficient-settle race, risking a silent false-green on the dangerous side**
`tests/scripts/test_parity_e2e.py:695-757` · root cause `src/factory/cli/ops.py:211`
— architect (80%)

`_settle_deny` (added this PR) hardens the *deny* direction against `_probe`'s single-`asyncio.sleep(0)`-yield race. The *allow* direction wasn't hardened the same way — architect's hypothesis (not empirically reproduced, but logically traced): a broken allow-grant on the **last** probed subject could have its late `-ERR` land inside the *next* probe's window, where an `expect_deny=True` check would treat it as expected denial — silently absorbing a real regression instead of surfacing it. This is the more dangerous direction (undetected over-permissive grant vs. undetected under-permissive one), and the root cause is shared with production (`factory ops verify` calls the same unpatched `_probe`). Out of scope for #2247 (ops.py is reused-as-is), but worth a tracked follow-up alongside the renderer fix above.

**issue: test broker binds `0.0.0.0`, not `127.0.0.1` as the spec's own stated security constraint requires**
`tests/scripts/test_parity_e2e.py:333-345`
— security-auditor (90%)

Confirmed: no `-a`/`--addr` flag on the `nats-server` Popen invocation, and `nats-server` defaults to `0.0.0.0`. Both the client port and the unauthenticated HTTP monitor port (`-m`) listen on all interfaces during the test's lifetime. On a shared runner or LAN-reachable dev host, a co-located party could hit the monitor endpoint (`/varz`, `/connz`, `/subsz`) unauthenticated. Client port itself is still nkey-gated (no `no_auth_user`), so publish/subscribe isn't directly reachable — security-auditor rates this Low in practice, but it's a real mismatch between the spec's stated constraint and the implementation. One-line fix: add `"-a", "127.0.0.1"` to the Popen argv.

**issue: `_settle_deny`'s bounded polling contradicts the spec's explicit "no bespoke timing budget" statement**
`tests/scripts/test_parity_e2e.py:230-263` vs. `artifacts/specs/2247-matrix-driven-live-acl-round-trip-spec.mdx:86,168,183`
— product-lead (90%)

The spec states three separate times that deny-probe timing is governed solely by `_probe`'s own `flush(timeout=2)`, with no bespoke budget. The shipped `_settle_deny` is a bounded 5×0.2s poll layered on top — necessary in practice (its own docstring: a one-shot wait left 3/68 nodes flaky) but an unreconciled deviation from the as-written spec. Not a correctness concern; a documentation-sync gap. Recommend amending the spec's Design Note / Breadboard N5 / Edge Cases to reflect the as-built behavior, the same way the `GITHUB_ACTIONS` vs `CI` judgment call is already documented elsewhere in this spec.

**issue: N6's CI hard-fail guard has zero automated regression test**
`tests/scripts/test_parity_e2e.py:65`
— tester (90%), related: devops (78%) on collection-scope pinning

The new module-level `GITHUB_ACTIONS`-gated hard-fail guard is verified only by manual command in the PR's Test Plan, not by an executable test — nothing falsifies it if the guard is deleted or its condition silently inverted in a future edit. Mitigating context: product-lead independently ran both directions manually this session (hard-fail confirmed exit 1 under `GITHUB_ACTIONS=true` + hidden binaries; graceful skip confirmed otherwise) and devops confirmed the `tests` CI job installs nats-server/nk before reaching this file's collection — so the guard is live-verified today, just not self-defending against future drift. Devops adds a related, lower-confidence point: nothing pins `test_parity_e2e.py` into the `tests` job's collection scope either, so a future `--ignore` change could silently drop ACL coverage with no failing signal (recommends a grep-based CI-stage gate rather than importing the module from a binary-less job, to avoid tripping N6 itself).

---

### Suggestions

- **`_non_flow_pair_target` always resolves to the same identity ("blobstore") across all 9 flows** — target axis has zero diversity despite 9 parametrize IDs implying independent scenarios (tester, 62%). Not a tautology (each node exercises a real live probe), but a coverage-shape gap worth documenting alongside the spec's other two disclosed limitations, or fixing via index/hash-based target rotation.
- **`_settle_deny`'s docstring symptom (b) attributes flakiness to timing for the exact 3 identities that are actually deterministically blocked by the empty-allow-list gap** — mutually exclusive explanations for the same identities; will mislead future maintainers editing this helper (architect, 70%). Recommend separating the two causes in the docstring.
- **`_settle_deny` not backported to 3 pre-existing sibling tests** (`test_hub_publish_acl_enforced`, `test_voice_tts_publish_acl_enforced`, `test_clipool_worker_subscribe_acl_enforced`) using the identical single-sleep idiom (tester, 45%). Legitimately out of this PR's scope; flagged as `parallel-path-drift` for a future pass.
- **`_settle_deny`'s 1s bound could itself produce a false failure** on a CPU-starved CI runner — opposite direction from the false-pass it was built to prevent (devops, 45%). Recommend logging the observed settle time on first failure rather than pre-emptively widening.
- **N4 positive round-trip doesn't isolate the static inbox grant from the dynamic `allow_responses` grant** — already documented as a known limitation in the spec (postmortem Fix 3 §8), not a tautology, just flagging so it isn't mistaken for full isolation (security-auditor, 70%).
- Minor doc nits (all low-confidence, non-blocking): plan artifact's `make test-acl -- --collect-only` doesn't forward args on GNU Make (devops, 88%, historical doc only); Makefile comment for `test-acl` omits nats-py as a hard requirement alongside the two binaries (devops, 35%); "verified by isolated reproduction" claims (timing race + empty-allow-list bug) leave no repro artifact in the diff, evidence-trail-only nit (tester, 30%).

---

### Praise

- **`xfail(strict=True)` as a self-removing security tripwire** — computed dynamically from `EFFECTIVE_GRANTS`, not hardcoded identity names; XPASS forces marker removal the instant the renderer bug is fixed. Independently praised by both security-auditor and architect as correct security-engineering judgment.
- **Reuse-vs-new split is architecturally correct** — reuses `factory.cli.ops` primitives and `roxabi_contracts.verify.verify_deny` (the SSoT deny-subject) rather than re-deriving; new subscribe helpers are legitimately new code since `ops.py` is publish-only. The 2-instance publish/subscribe mirror is below the `target-axis-trap` promotion threshold (architect).
- **CI ride-along empirically verified sufficient** — devops traced the actual job topology and confirmed nats-server/nk are installed before the `tests` job's broad `pytest tests/` collection reaches this file (92%).
- **`GITHUB_ACTIONS` (not generic `CI`) is the right env-var choice** — devops verified this empirically against `pytest_smoke`'s no-marker-filter collection behavior, where a generic `CI=true` would false-positive hard-fail on any dev shell exporting it (90%).
- **Spec compliance independently re-verified, not trusted** — product-lead ran the actual test suite locally (65 passed, 3 xfailed, matches PR claim exactly), confirmed the hard-fail/skip guard fires correctly in both directions, and confirmed no 3rd test file was created and both static test files are docstring-only. All ~10 success criteria and both issue-level "do NOT" constraints are genuinely met.

---

### Blind-spot sweep

All 5 agents independently ran the repo's `review-blind-spots.md` checklist. No violations found: bounded loops/timeouts throughout, no bare/broad excepts, no shell/SQL injection surface, no hardcoded secrets (nkeys minted fresh per test run into `tmp_path_factory`, never committed, `chmod(0o600)`), pathlib used throughout, no unlocked shared-mutable-state races (single-threaded asyncio, `xdist_group` pins the file to one worker), no missing `await`. One non-finding noted by security-auditor: a brief non-atomic write-then-chmod window on seed files, mitigated by the containing tmp directory already being `0o700`.

MickaelV0:
## Code Review (iteration 2 — post-fix)

**Methodology note:** this pass reviews the fix commit (`b7183832f`) applied against the first review's four addressable findings, in this execution context without independent Task-spawned subagents — conducted as a single rigorous pass rotating through the same domain lenses (security, architecture, tests, devops, product/spec) against the fix diff, the full file's current state, and a live re-run of the suite. It complements, not replaces, the first pass's 5-agent review of the original 476-line diff, which is unchanged in substance by this commit.

**Verdict: Request changes** — carried forward from the first review's process-level blocker only. No new code-correctness issues found in the fix commit itself; 3 of 4 addressable findings are fixed and verified, the 4th was investigated and correctly left as a documented non-fix per the review's own guidance.

---

### Blockers (carried forward, unchanged)

**todo (blocking): live production ACL gap discovered by #2247 still has no filed tracking issue**

Unchanged from the first review. Not addressed in this iteration by design — filing a GitHub issue is outside this pipeline's authority (Safety Rule requires human approval before `gh issue create`). A ready-to-paste issue draft has been prepared and handed to the human operator out-of-band (not posted here to avoid an unauthorized `gh issue create`). This remains the sole blocker; the affected identities (`dashboard-reader`, `gh-helper`, `ingress`) and nats-server's empty-`allow: []`-is-unrestricted behavior are unchanged by this commit.

---

### Findings on the fix commit

**praise: N6 CI hard-fail guard's decision logic (not just the missing-deps list) extracted into a pure, unit-tested function**
`tests/scripts/test_parity_e2e.py:65-96`
— architect/tester lens, 90%

`_ci_hardfail_missing_deps(*, github_actions, nats_available, nk_available, nats_py_available) -> list[str]` now encapsulates the *entire* decision, including the `github_actions` gate itself — not just the dep-availability comprehension. Three new tests (`test_ci_hardfail_guard_noop_outside_github_actions`, `test_ci_hardfail_guard_fires_in_github_actions_with_missing_deps`, `test_ci_hardfail_guard_silent_in_github_actions_with_all_deps_present`) exercise all three meaningful branches, including the specific "silently-inverted `== \"true\"` condition" failure mode the original finding named. Verified: manual repro (`GITHUB_ACTIONS=true PATH=/usr/bin uv run pytest tests/scripts/test_parity_e2e.py --collect-only`) still hard-fails with the expected message post-refactor; full file re-run is 68 passed, 3 xfailed (unchanged from pre-fix baseline); ruff/pyright clean.
Confidence: 92%

**issue → resolved: test broker now binds `127.0.0.1` instead of `0.0.0.0`**
`tests/scripts/test_parity_e2e.py:363-364, 1013-1014`
— security-auditor lens, 90%

`"-a", "127.0.0.1"` added to *both* `nats-server` `Popen` invocations in the file — the PR-introduced `nats_server` module fixture the finding cited, and the pre-existing (untouched by #2247) `test_retired_identity_connect_rejected` helper, which has the identical pattern and identical exposure. Empirically verified directly (isolated `nats-server -a 127.0.0.1 -p ... -m ...` run): both the client and monitor ports now listen on `127.0.0.1` only, versus `*` (all interfaces) without the flag. Closes the unauthenticated-monitor-port (`/varz`, `/connz`, `/subsz`) exposure on a shared runner or LAN-reachable dev host that the first review flagged as Low-severity-in-practice but a real spec/implementation mismatch.
Confidence: 95%

**issue → resolved: spec/implementation doc drift around `_settle_deny`'s timing budget**
`artifacts/specs/2247-matrix-driven-live-acl-round-trip-spec.mdx` (Design Note new item 4; Expected Behavior item 5; Breadboard N5 row; Edge Cases N4/N5 row)
— product-lead lens, 88%

A 4th Design Note judgment call now documents `_settle_deny`'s bounded 5×0.2s follow-up poll as an intentional as-built deviation from the spec's original "no bespoke timing budget" wording, framed the same way the pre-existing `GITHUB_ACTIONS` vs `CI` judgment call documents its own deviation from precedent (recorded, not silently made). The three call-out sites the finding named (Expected Behavior item 5, Breadboard N5, Edge Cases N4/N5) are all updated to cross-reference it. Grepped the file post-edit for the old "no bespoke budget" / "governs timing" phrasing — no stale copies remain.
Confidence: 90%

**thought: allow-probe settle-race — investigated, not reproduced, correctly left as-is**
`tests/scripts/test_parity_e2e.py:695-757` (unchanged) · root cause `src/factory/cli/ops.py:211` (unchanged, reused-as-is, out of scope)
— architect lens, 70%

The architect's hypothesis (broken allow-grant on the *last* probed subject leaking a late `-ERR` into the immediately-following deny-probe's window, silently absorbing a real over-permissive regression) was not empirically reproduced this iteration — doing so deterministically requires synthetic ACL fault injection (a scratch matrix where an intended-allow subject is actually denied) plus a way to force the `error_cb` delivery past the single-yield window, neither of which exists in the current harness without new test infrastructure. Two options were considered and rejected: (a) a per-subject settle-and-check loop mirroring `_settle_deny` (rejected — asymmetric with the existing 5×0.2s bound, adds an unconditional per-node sleep, and risks manufacturing exactly the false-failure-on-a-slow-runner problem devops already flagged about `_settle_deny` itself, in new unreviewed code during what's meant to be the converging iteration); (b) leaving it fully undocumented (rejected — loses the finding). Landed on: no code change, folded as a second bullet into the tracking-issue draft handed to the human operator, since the architect explicitly called it "worth a tracked follow-up alongside the renderer fix" — same durable home as the primary gap, not a rushed patch.
Confidence: 65% (this is a scope/process judgment call, not a technical finding)

---

### Not re-litigated (unchanged from first review, out of this iteration's scope per task instructions)

- Target-axis diversity gap (`_non_flow_pair_target` always resolves to `blobstore`)
- `_settle_deny` docstring's two-symptom conflation (timing vs. deterministic empty-allow-list gap)
- `_settle_deny` not backported to 3 pre-existing sibling tests
- CI collection-scope pinning for `test_parity_e2e.py` (devops's lower-confidence point)
- N4 static/dynamic grant isolation
- Minor doc nits (Makefile arg-forwarding, nats-py requirement omission, repro-artifact-in-diff nit)

None of these were in the four items this iteration was scoped to fix; carrying them forward unchanged rather than scope-creeping.

---

### Blind-spot sweep (fix commit only)

Bounded loops/timeouts throughout (new `_ci_hardfail_missing_deps` is a plain list comprehension, no loop risk). No bare/broad excepts introduced. No shell/SQL injection surface (both `Popen` calls remain list-form argv, no `shell=True`). No hardcoded secrets. No unlocked shared-mutable-state races. No missing `await`. Pure-markdown spec edit carries zero runtime risk.

MickaelV0:
## Review Fixes Applied

**Process blocker resolved:** Filed tracking issue #2267 for the live empty-allow-list ACL gap (`scripts/_renderer.py` must emit explicit deny when allow-list is empty).

**Changes:**
- `tests/scripts/test_parity_e2e.py`: xfail reason strings and docstring now reference #2267

**Deferred (out of scope #2247):** renderer fix itself → #2267

PR #2254, opened 2026-07-05
feat(api): add /health/ready endpoint with NATS round-trip check

### LINKED ISSUE #2202: feat(api): add /health/ready endpoint with NATS round-trip check
**Origin:** #2195

Split liveness (/health, process alive) from readiness (/health/ready, can serve traffic).

Current /health endpoint does not verify NATS connectivity. The 2026-04-27 incident showed the hub was alive but unable to communicate with adapters for 3h15m, with /health still returning 200.

Add /health/ready endpoint that performs a real NATS/JetStream round-trip: `kv.get('hub.ready')` against the `factory-state` bucket (the same probe `wait_for_hub()` already uses at adapter boot, `packages/roxabi-nats/src/roxabi_nats/readiness.py`).

**Do NOT use `factory.system.ready` request/reply** (the mechanism this issue originally proposed): ADR-065 demoted that path off the critical boot-readiness route after it caused a 20-min stall + ~1200 ACL violations/s under `allow_responses:false`; the hub only *subscribes* to `factory.system.ready` today (no publish grant), so a self-probe would need a new ACL grant to originate the request — for zero extra detection value, since it's a hub-to-itself loopback either way.

**Do NOT wire this into the Quadlet HealthCmd/auto-restart** (reversing the issue's original "update Quadlet healthcheck to use this endpoint" ask): `make converge` restarts `factory-nats` routinely, which would flip the hub "unhealthy" on every deploy and risk a crashloop during an actual ACL outage. Keep the container's process-liveness `HealthCmd` unchanged. `/health/ready` is a **separate, unauthenticated** endpoint consumed by deploy-smoke/canary gates (`docs/runbooks/deploy-smoke-canary.md`) and external monitoring — never tied to restart.

**Honest scope**: this endpoint proves the hub's own message-plane substrate (NATS/JetStream reachability) is live — strictly better than today's `_probe_nats` which only checks `nc.is_connected` (a passive TCP flag that stayed "true" throughout the April incident). It does **not** detect a peer identity's ACL break (the actual April fault — clipool's reply path). That class of failure is caught pre-prod by a standalone ACL test-acl issue and, in production, by the incident-response detection-reactivation work split from #2203.

Ref: validated by architect/product-lead/devops panel review, 2026-07-04 (KV mechanism 3/4 confidence, Quadlet-decoupling unanimous).

MickaelV0:
## Summary
- Add `/health/ready` — an unauthenticated readiness probe that round-trips `kv.get('hub.ready')` on the `factory-state` JetStream KV bucket, bounded to a 5s single attempt (`_probe_nats_ready`).
- Unlike `/health` (pure process liveness) or the passive `nc.is_connected` TCP flag used by `/health/detail`'s NATS check, this proves the message plane is actually reachable — the 2026-04-27 incident had the hub reporting 200 on `/health` for 3h15m while adapters couldn't reach it, because the TCP socket stayed up even though JetStream/ACL access was wedged.
- Deliberately **not** wired into the Quadlet `HealthCmd`/auto-restart: ADR-065 already demoted request/reply readiness off the critical boot path after a 20-min stall + ~1200 ACL violations/s incident. This is a pull-style plane③ probe (ADR-091) for `deploy-smoke`/canary gates and external monitoring to consume — wiring those consumers is out of scope for this issue.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #2202: feat(api): add /health/ready endpoint with NATS round-trip check | OPEN |
| Implementation | 2 commits on `feat/2202-health-ready-endpoint` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (9 new) | Passed |

## Test Plan
- [x] `TestHealthReady` (9 cases): no-auth-required, ready-true on successful round-trip, and 503 for each failure mode — `nc=None`, `BucketNotFoundError`, `KeyNotFoundError`, unexpected KV value, generic `nats.errors.Error`, timeout (probe timeout monkeypatched to 0.05s), and an unexpected exception (asserts `log.exception` fired).
- [x] Manual falsification: stashed `health.py`, confirmed all 9 new tests fail (404 — route doesn't exist), restored, confirmed green.
- [ ] Manual: `curl localhost:8443/health/ready` against a running hub — 200 when `announce_hub_ready()` has run (confirmed via code read: `hub_standalone.py` calls it before `create_health_server`), 503 otherwise.

Fixes #2202

---

### DIFF
--- a/src/factory/bootstrap/infra/health.py
+++ b/src/factory/bootstrap/infra/health.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import hmac
 import logging
 import os
@@ -11,7 +12,7 @@ from pathlib import Path
 from typing import Any
 
 from fastapi import FastAPI, Header, HTTPException
-from fastapi.responses import PlainTextResponse
+from fastapi.responses import JSONResponse, PlainTextResponse
 
 from factory.core.hub import Hub
 from factory.paths import factory_data_dir
@@ -82,6 +83,57 @@ def _probe_nats(nc: Any | None) -> str | None:
     except AttributeError as exc:
         log.debug("_probe_nats: unexpected exception from nc.is_connected: %s", exc)
         return "unreachable"
+
+
+# Bounded readiness probe — a live server reports its *current* state, it does
+# not block waiting for one (unlike roxabi_nats.readiness.wait_for_hub, which
+# is a boot-time barrier with retry + KV-watch fallback). Keeps the HTTP
+# handler fast even under a wedged NATS/JetStream backend (#2202).
+_READY_TIMEOUT_S = 5.0
+
+
+async def _probe_nats_ready(nc: Any | None) -> tuple[bool, str | None]:
+    """Round-trip through NATS/JetStream KV to prove the message plane is live.
+
+    Fetches ``hub.ready`` from the ``factory-state`` KV bucket — the same key
+    ``announce_hub_ready()`` writes on hub startup and ``wait_for_hub()``
+    (``roxabi_nats.readiness``) reads at adapter boot — but as a single
+    bounded attempt rather than a retry/watch loop.
+
+    This is a strictly stronger signal than ``_probe_nats``'s ``nc.is_connected``
+    check: that TCP flag stayed ``True`` throughout the 2026-04-27 incident
+    (hub alive, adapters unreachable for 3h15m) because it only reflects
+    socket state, not JetStream/ACL health (#2202).
+
+    Returns ``(ready, reason)`` — ``reason`` is ``None`` when ready, else a
+    short machine-readable explanation for the failure.
+    """
+    if nc is None:
+        return False, "nats client not configured"
+
+    import nats.errors
+    from nats.js.errors import BucketNotFoundError, KeyNotFoundError
+
+    try:
+        async with asyncio.timeout(_READY_TIMEOUT_S):
+            js = nc.jetstream()
+            kv = await js.key_value("factory-state")
+            entry = await kv.get("hub.ready")
+    except TimeoutError:
+        return False, "nats round-trip timed out"
+    except BucketNotFoundError:
+        return False, "factory-state bucket not provisioned"
+    except KeyNotFoundError:
+        return False, "hub.ready key not found"
+    except nats.errors.Error as exc:
+        return False, f"nats error: {exc}"
+    except Exception:
+        log.exception("_probe_nats_ready: unexpected error during KV round-trip")
+        return False, "unexpected error"
+
+    if entry.value != b"true":
+        return False, "hub.ready value unexpected"
+    return True, None
 
 
 def _collect_hub_detail(  # noqa: C901 — optional sections (nats/reaper/circuits)
@@ -246,6 +298,13 @@ def create_health_app(
     surfaces NATS reachability under the ``nats`` key and an overall
     ``status`` of ``ok``/``degraded``. When ``NATS_URL`` is unset both
     fields are omitted.
+
+    ``/health/ready`` (#2202) is a separate, unauthenticated readiness probe:
+    process liveness (``/health``) says nothing about whether the hub can
+    actually reach NATS/JetStream, so it round-trips a KV read instead of
+    trusting the passive ``nc.is_connected`` TCP flag. Deliberately **not**
+    wired into the Quadlet ``HealthCmd``/auto-restart — see the probe's
+    docstring and #2202 for the incident/ADR-065 rationale.
     """
     _secrets = secrets or Secrets()
     app = FastAPI(title="factory Hub")
@@ -253,6 +312,13 @@ def create_health_app(
     @app.get("/health")
     async def health() -> dict:
         return {"ok": True}
+
+    @app.get("/health/ready")
+    async def health_ready() -> JSONResponse:
+        ready, reason = await _probe_nats_ready(nc)
+        return JSONResponse(
+            {"ready": ready, "reason": reason}, status_code=200 if ready else 503
+        )
 
     @app.get("/health/detail")
     async def health_detail(authorization: str = Header(default="")) -> dict:
--- a/tests/test_health_endpoint_status.py
+++ b/tests/test_health_endpoint_status.py
@@ -1,14 +1,17 @@
 """Tests for /health endpoint — status, authenticated detail, and hub timestamps.
 
-Covers: issue #111, SC-1, SC-2, SC-3, #207.
-Classes: TestHealthUnauthenticated, TestHealthEndpoint, TestHubTimestamps.
+Covers: issue #111, SC-1, SC-2, SC-3, #207, #2202.
+Classes: TestHealthUnauthenticated, TestHealthEndpoint, TestNatsHealthProbe,
+TestHealthReady, TestHubTimestamps.
 """
 
 from __future__ import annotations
 
+import asyncio
+import logging
 import time
 from datetime import datetime, timezone
-from unittest.mock import Mock
+from unittest.mock import AsyncMock, MagicMock, Mock
 
 import pytest
 from httpx import ASGITransport, AsyncClient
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Code Review

Reviewed by security-auditor, tester, and backend-dev (parallel domain agents), all cross-checking claims against the actual installed `nats-py` source and this repo's sibling wiring modules rather than assuming. Single chunk (330-line diff), no cross-chunk join needed. Axial-adr-review and product-lead/architect/frontend-dev/devops were scoped out (diff touches `infra/`, not `infrastructure/`; no spec artifact for this Tier-S issue; no arch/frontend/config surface touched).

**Verdict: Approve with comments.** No blocking findings — every reviewing agent's own bottom line agrees this does not block merge. Several real, independently-corroborated non-blocking items below, worth a fast follow-up.

### Warnings

suggestion (parallel-path-drift): hardcoded `"factory-state"` literal instead of the `FACTORY_STATE_BUCKET` constant
  `src/factory/bootstrap/infra/health.py:120,125` (also referenced in the docstring at :98)
  -- tester
  Root cause: `src/factory/infrastructure/kv/factory_state.py:7` defines `FACTORY_STATE_BUCKET = "factory-state"`, and three sibling wiring modules (`kv_agent_roster.py`, `kv_watch_channels.py`, `kv_bot_roster.py`) all import and use the constant. `health.py`'s new probe hardcodes the string literal instead — verified directly, not assumed.
  Class: [parallel-path-drift]
  Raw callsites: [{file: "src/factory/bootstrap/infra/health.py", line: 120}, {file: "src/factory/bootstrap/infra/health.py", line: 125}]
  Solutions:
    1. `from factory.infrastructure.kv.factory_state import FACTORY_STATE_BUCKET` and use it in both call sites (recommended)
    2. Leave as-is — no functional risk, just a lost single-source-of-truth guarantee if the bucket name ever changes
  Confidence: 70%

suggestion (candidate/error-detail-leak): raw NATS exception text interpolated into the unauthenticated response body, with asymmetric logging
  `src/factory/bootstrap/infra/health.py:128-129`
  -- security-auditor, backend-dev (independently, converging)
  Root cause: every other failure branch in `_probe_nats_ready` returns a static, purpose-built reason string; the `except nats.errors.Error as exc:` branch instead does `f"nats error: {exc}"`, forwarding raw client-exception text to an unauthenticated caller — and (unlike the `except Exception` branch two lines below) does not `log.exception` it server-side, so the operator's own logs see less than the anonymous caller does. Checked against the real `nats-py` source: all `nats.errors.*` messages are static strings (no credentials/URLs/tokens); the one dynamic path (`APIError.__str__`) interpolates only a server-provided stream/subject description. So the residual leak is low-value internal-architecture reconnaissance, not a credential leak, and is additionally gated behind the Quadlet's `PublishPort=127.0.0.1:8443:8443` loopback bind — one config change away from being network-reachable.
  Class: [candidate/error-detail-leak]
  Raw callsites: [{file: "src/factory/bootstrap/infra/health.py", line: 129}]
  Solutions:
    1. Return a generic `reason` (e.g. `"nats error"`) for this branch and move the detail to a server-side `log.warning(...)`, matching every other branch's pattern (recommended)
    2. Keep interpolating but bound/sanitize (e.g. `type(exc).__name__` only)
  Confidence: 63%

issue (missing-error-handling): `entry.value` accessed outside the try/except that's supposed to make this function exception-safe
  `src/factory/bootstrap/infra/health.py:134-136`
  -- backend-dev
  Root cause: `_probe_nats_ready`'s whole contract is "never raise, always return `(bool, reason)`" — every NATS/JetStream failure is mapped to a tuple inside the try block (lines 117-132), but `if entry.value != b"true":` runs after the try/except closes, unprotected. The function's own docstring cites `wait_for_hub()` (`roxabi_nats/readiness.py:203-206`) as precedent, and that sibling protects the equivalent check *inside* its try block — this diff diverges from the precedent it names. Today this is likely unreachable (nats-py's `KeyValue.get()` contractually returns an `Entry` with `.value` or raises `KeyNotFoundError`), so severity is low.
  Raw callsites: [{file: "src/factory/bootstrap/infra/health.py", line: 134}]
  Solutions:
    1. Move the `.value` check inside the try block, matching `wait_for_hub`'s pattern (recommended)
    2. Leave as-is, given today's nats-py contract makes the path unreachable
  Confidence: 55%

issue: `KeyDeletedError` (tombstoned `hub.ready` key) is a sibling of `KeyNotFoundError`, not caught by it — falls through to the generic branch, untested
  `src/factory/bootstrap/infra/health.py:126`
  -- tester
  Root cause: verified directly against the installed `nats-py` in this worktree's venv: `KeyDeletedError` and `KeyNotFoundError` are both `(KeyValueError, NotFoundError, APIError, Error)` — siblings, not parent/child. `except KeyNotFoundError:` does not catch `KeyDeletedError`, so a soft-deleted (vs. never-existing) `hub.ready` key falls to `except nats.errors.Error as exc:`, yielding `"nats error: nats: key was deleted"` instead of a purpose-built reason. Still fails safe (503, no crash, logged via the generic branch's sibling handling), just an unhandled/untested branch not mentioned in the docstring's failure-mode list.
  Raw callsites: [{file: "src/factory/bootstrap/infra/health.py", line: 126}]
  Solutions:
    1. Add `except KeyDeletedError: return False, "hub.ready key deleted"` next to the `KeyNotFoundError` clause, with a matching test (recommended)
    2. Document that the generic fallback intentionally covers this case
  Confidence: 55%

### Suggestions

nitpick: `test_ready_no_auth_required` is a strict subset of `test_ready_true_when_kv_round_trip_succeeds` — no auth mechanism exists for it to falsify
  `tests/test_health_endpoint_status.py:149`
  -- tester
  Root cause: `/health/ready` takes no `Header`/`Authorization` parameter at all (unlike `/health/detail`), so there's no guard this test could ever break by itself — removing any real code wouldn't make it fail differently than its sibling success test.
  Solutions:
    1. Drop it — the class docstring already documents "no auth header required" as context (recommended)
    2. Or make it distinguishing: pass a garbage `Authorization` header and assert still-200 (unlike `/health/detail`'s 401)
  Confidence: 70%

suggestion (candidate/missing-integration-coverage): all 9 `TestHealthReady` tests mock the NATS/JetStream chain; no test exercises a real `nats-server`
  `tests/test_health_endpoint_status.py:142`
  -- tester
  Root cause: the mocked shape (`nc.jetstream()` sync, `key_value()`/`get()` async) is verified contract-accurate against the installed `nats-py` today, so this isn't a live tautology — but a future `nats-py` version bump changing that shape would keep all 9 mocked tests green while breaking the real endpoint. The repo already has real-`nats-server` test infrastructure usable here (`tests/factories/nats_server.py`, and `packages/roxabi-nats/tests/test_readiness.py` tests the identical `factory-state`/`hub.ready` key against a live server).
  Class: [candidate/missing-integration-coverage]
  Raw callsites: [{file: "tests/test_health_endpoint_status.py", line: 142}]
  Solutions:
    1. Add one `@requires_nats_server`-style test mirroring `packages/roxabi-nats/tests/test_readiness.py` (recommended)
    2. Defer explicitly with a `NO TEST — infra-not-wired` note if intentionally out of scope for this Tier-S slice
  Confidence: 65%

suggestion: `/health/ready` performs 2 NATS round-trips per call, not 1 — `key_value()`'s handle isn't cached
  `src/factory/bootstrap/infra/health.py:119-121`
  -- backend-dev
  Root cause: `js.key_value("factory-state")` internally calls `stream_info()` (a real JetStream RPC) before `kv.get()` does its own round-trip — so every hit is 2 RPCs. The bucket's stream metadata is static, so re-fetching it every call is redundant, though likely immaterial given the endpoint is explicitly not wired into Quadlet `HealthCmd`/auto-restart (low-frequency, human/monitoring-driven per the docstring).
  Solutions:
    1. Leave as-is — acceptable given documented low-frequency usage (recommended)
    2. Cache the `KeyValue` handle after first success, re-resolve on failure
  Confidence: 55%

### Praise

praise: `asyncio.timeout()` correctly bounds the entire KV round-trip, and the timeout test is not tautological
  `src/factory/bootstrap/infra/health.py:118-121`, `tests/test_health_endpoint_status.py` (`test_ready_false_on_timeout`)
  -- tester, backend-dev
  Confirmed via falsification (removing the `asyncio.timeout` guard makes the test fail with a *different* error message, proving it genuinely exercises the guard) and via exception-ordering review (`TimeoutError` is caught before the broader `nats.errors.Error`, correct regardless of which layer originates the timeout). Exactly the "unbounded network call without timeout" blind-spot, handled correctly.
  Confidence: 95%

praise: AAA structure and test naming match house style; broad `except Exception` in the probe is intentional, tested, fails closed
  `tests/test_health_endpoint_status.py`, `src/factory/bootstrap/infra/health.py:130-132`
  -- tester, backend-dev, security-auditor
  All 9 new test names are behavior-descriptive and unambiguous. The generic `except Exception` branch logs full traceback via `log.exception` and returns `ready=False`/503 rather than swallowing silently or crashing — mirrors the established pattern in `roxabi_nats/readiness.py`'s `announce_hub_ready`/`_kv_watch_for_ready`. Does not swallow `asyncio.CancelledError` (that's `BaseException`).
  Confidence: 90%

### Blind-spot audit (review-blind-spots.md)

Applicable and cleared: unbounded network call without timeout (this PR *is* the fix — bounded correctly), broad except-without-reraise (intentional fail-closed pattern, logged, tested), missing `await` on coroutine (verified line-by-line against real `nats-py` signatures — none missing). Not applicable: injection surfaces, hardcoded secrets, path traversal, mutable defaults, subprocess/shell/DB — this diff has no such surface (zero external input reaches `/health/ready`: no query params, body, or headers).

### Production-path note (not a finding, for context)

The production entrypoint (`factory-hub.container` → `factory hub` → `hub_standalone.py`) calls `announce_hub_ready(nc)` *before* starting the health server, so `/health/ready` will correctly report `ready:true` once live in prod. Unified/single-process mode (`bootstrap_lifecycle.py`) never calls `announce_hub_ready` — `/health/ready` would always read 503 there — but unified mode is not the deployed path, so this is a pre-existing gap this PR doesn't create or worsen.

PR #2255, opened 2026-07-05
chore(security): wire nkey rotation logging + check-seed-age

### LINKED ISSUE #2246: chore(security): enforce nkey rotation policy — wire rotation_log_append() on routine path + make check-seed-age
**Origin:** #2204 (credential rotation policy doc) — split out per architect/security/devops panel review, 2026-07-04.

**Problem**: `rotation_log_append()` (`src/factory/operator_audit.py`) is only called from the disaster-recovery path (`src/factory/cli/secrets_reset.py:161`). Routine per-identity nkey rotation (`docs/runbooks/nkey-rotation.md`) never logs — so the rotation-log (`~/.roxabi/factory/rotation-log.md`), the only authoritative age source, is structurally incomplete. A documented "max seed age: 90 days" policy with no enforcement mechanism and no reliable data source is inert prose, not a control.

**Scope**:
1. Wire `rotation_log_append()` into the routine rotation path (the `factory-acl`/`gen-nkeys.sh` flow used by `nkey-rotation.md`), not only disaster-recovery.
2. Add a `make check-seed-age` target that reads `rotation-log.md` as the source of truth (do not rely on seed-file mtime as primary — Syncthing/rsync/restore silently reset mtimes; mtime may be used only as a documented fallback for identities with no log entry).
3. Threshold: warn at 75 days, hard-fail at 90 days (per #2204's policy) — avoid alert fatigue / spurious CI red on identities legitimately near the boundary.
4. Bootstrap: identities rotated before this wiring lands have no log entry — document the "unknown age → assume now" grace on first run.
5. Confirm rotation includes revocation (old seed removed from `auth.conf` + consuming unit restarted) — an aged-out policy is meaningless if the old credential still authenticates.

**Acceptance**:
- [ ] Routine rotation path calls `rotation_log_append()` on every rotation
- [ ] `make check-seed-age` reads the log, warns ≥75d, exits non-zero ≥90d
- [ ] Bootstrap grace documented for pre-existing unlogged seeds
- [ ] Wired as a scheduled/cron or pre-push check (not merge-blocking on unrelated PRs)

Ref: validated by architect/security/devops panel review, 2026-07-04.

MickaelV0:
## Summary
- Wires `rotation_log_append()` into both nkey provisioning paths (`_mode_add_identity` guarded to `state=="added"`, `_mode_full_provision` per active identity) so routine rotations are auto-logged instead of relying only on the DR blanket entry.
- Adds `factory-acl check seed-age` / `make check-seed-age` — a pure checker (`scripts/check_seed_age.py`) that warns at ≥75d / fails at ≥90d based on `rotation-log.md`, with unconditional bootstrap grace for identities with zero log entries (seed-file mtime surfaced only as an informational `NOTE:`, never gating) — wired into a daily systemd timer, never a CI/merge gate.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #2246: chore(security): enforce nkey rotation policy — wire rotation_log_append() on routine path + make check-seed-age | OPEN |
| Spec | [2246-enforce-nkey-rotation-policy-spec.mdx](artifacts/specs/2246-enforce-nkey-rotation-policy-spec.mdx) | Present |
| Plan | [2246-enforce-nkey-rotation-policy-plan.mdx](artifacts/plans/2246-enforce-nkey-rotation-policy-plan.mdx) | Present |
| Implementation | 4 commits on `feat/2246-enforce-nkey-rotation-policy` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (15 new: 14 in `test_check_seed_age.py` + 1 in `test_genkeys_modes.py`; +2 rotation-log assertion blocks in `test_genkeys_modes_add_identity.py`) | Passed |

## Test Plan
- [x] `uv run --frozen pytest tests/scripts/test_genkeys_modes.py tests/scripts/test_genkeys_modes_add_identity.py tests/scripts/test_check_seed_age.py tests/test_operator_audit.py tests/cli/test_cli_secrets_reset.py -q` — 69 passed
- [x] Full repo suite (`uv run --frozen pytest -q`) — 6519 passed, 0 failures (26 pre-existing skips, 1 pre-existing xfail)
- [x] `uv run --frozen ruff check .` — clean; `ruff format --check` on touched files — clean
- [x] `uv run --frozen pyright` (repo-wide) — 0 errors
- [ ] Manual: on M1, `make quadlet-sync-install` picks up the new timer and `systemctl --user status factory-check-seed-age.timer` shows scheduled

## SC → Test Matrix

| SC | Test(s) | Status |
|----|---------|--------|
| SC1: `_mode_add_identity` logs only on `state=="added"`, not `"repaired"`/`"noop"` (N1, N10) | `tests/scripts/test_genkeys_modes_add_identity.py::test_add_identity_rotation_log_entry_only_on_added` | ✅ passed |
| SC2: `_mode_full_provision` logs once per active identity every invocation (N2, N10) | `tests/scripts/test_genkeys_modes.py::TestRotationLogWiring::test_full_provision_writes_one_rotation_log_line_per_active_identity` | ✅ passed |
| SC3: `_mode_regen_authconf` + `secrets_reset.py:161` DR call site unmodified | `git diff staging...HEAD -- src/factory/cli/secrets_reset.py` empty; no `_mode_regen_authconf` edits | ✅ verified |
| SC4: `check seed-age` exits 0 when all identities are <75d logged or have no entry | `tests/scripts/test_check_seed_age.py::TestWarnFailBoundaries` (74d case), `TestBootstrapGrace*` | ✅ passed |
| SC5: WARN in [75,90) exits 0 w/ `WARN:`; FAIL ≥90 exits 1 w/ `FAIL:`; no-entry identities never gate | `tests/scripts/test_check_seed_age.py::TestWarnFailBoundaries` (75/89/90/91d cases) | ✅ passed |
| SC6: zero-entry identity → grace; seed mtime surfaced as `NOTE:` only, never flips verdict even past 90d | `tests/scripts/test_check_seed_age.py::TestBootstrapGraceWithSeedFile::test_zero_entries_with_old_seed_mtime_never_gates` | ✅ passed |
| SC7: `make check-seed-age` invokes `uv run --frozen` | `grep -A1 'check-seed-age:' Makefile \| grep -- '--frozen'` | ✅ verified |
| SC8: systemd unit+timer mirror `factory-fleet-digest-poll` shape; registered in `quadlet-sync-install` | `deploy/systemd/factory-check-seed-age.{service,timer}` diffed against `factory-fleet-digest-poll.{service,timer}`; `grep -c factory-check-seed-age Makefile` = 5 | ✅ verified |
| SC9: `.claude/stack.yml` gains no new gate entry (never CI-blocking) | `grep -c 'check-seed-age\|check_seed_age' .claude/stack.yml` = 0 | ✅ verified |
| SC10: both subprocess `env={...}` and in-process `monkeypatch.setenv` sites set `ROTATION_LOG`+`OPERATOR_LOG`; no real-path writes (N9) | `tests/scripts/test_genkeys_modes.py` + `test_genkeys_modes_add_identity.py` diffs (14 + 11 call sites hardened) | ✅ passed |
| SC11: `nkey-rotation.md` documents auto-logging + bootstrap-grace design (N12) | `grep -q 'rotation_log_append\|auto-logged' docs/runbooks/nkey-rotation.md` | ✅ verified |
| SC12: `operator-log.md` + `deploy/AGENTS.md` no longer say "future nkey rotations" (N12) | `! grep -rq 'future nkey rotations' docs/runbooks/operator-log.md deploy/AGENTS.md` | ✅ verified |
| SC13: full targeted suite green | `uv run --frozen pytest tests/scripts/test_genkeys_modes.py tests/scripts/test_genkeys_modes_add_identity.py tests/scripts/test_check_seed_age.py tests/test_operator_audit.py tests/cli/test_cli_secrets_reset.py -q` | ✅ 69 passed |

## Judgment calls (not fully spelled out in the issue — flagging for review)

1. **Bootstrap-grace redesign.** The issue's own wording leaned toward an mtime-based fallback for identities with no log entry. Analysis (independently corroborated by /spec's product-lead + architect reviewers) showed this would misfire on rollout: all 18 active identities currently have zero log entries, and several seeds' real mtimes already sit near the 75-day WARN threshold — an mtime-gated check would produce a fleet-wide false-alarm storm on day one, contradicting the issue's own alert-fatigue and "assume now" intent. Redesigned to: log entry → gates; no entry → unconditional grace; mtime → informational-only `NOTE:`, never gating.
2. **Systemd mirror target.** Issue text suggested `factory-operator-logrotate` as the shape reference; switched to `factory-fleet-digest-poll` instead, since `factory-operator-logrotate` is bare-bash and lacks `WorkingDirectory=`/`PATH=`, which would leave this Python-based checker's relative-path resolution broken under systemd `--user`'s minimal environment.
3. **`--frozen` requirement.** Added a mandatory `--frozen` to both the Makefile target and the systemd unit's invocation, citing a documented M1 incident (`deploy/fleet-digest-poll.sh:16`) where a bare `uv run` dirtied `uv.lock` and jammed deploy for 12h.
4. **Test-hermeticity scope.** Broadened beyond the issue's literal wording to require both subprocess `env={...}` *and* in-process `monkeypatch.setenv` isolation — the latter wasn't explicitly called out but at least one existing test drives the same code path in-process, and would otherwise leak writes to the real `~/.roxabi/factory/rotation-log.md`/`operator.log` once this PR's wiring landed.
5. **`run()` return contract.** Specified as a pure `(warnings, failures, notes) -> tuple[list[str], list[str], list[str]]` with all printing owned by the CLI layer, rather than having the core module print directly — kept the age-computation logic trivially unit-testable without capturing stdout/stderr.
6. **DR gets "free" granular log lines.** `_mode_full_provision`'s new per-identity logging also fires during `_mode_regenerate`'s DR path, in addition to the existing blanket `secret:nats-nkeys` entry from `secrets_reset.py`. Kept as-is (both reviewers agreed this is a net improvement, not a duplication concern) — flagging in case a different DR-logging shape was expected.
7. **Environment note (process, not code):** this session's execution environment has no `TaskCreate`/`TaskUpdate`/`TaskList` tooling, so `/plan`'s micro-tasks were tracked via the plan artifact's own T-numbered blueprint rather than session tasks — documented in the plan's `## Task IDs` section, no impact on the shipped code.
8. **Buffer-then-flush rotation-log redesign (undisclosed at PR-open time).** `_mode_full_provision` originally called `rotation_log_append()` inline, once per identity, inside the per-identity seed-generation loop. Iteration-2 review flagged a log/rollback desync: that loop runs inside `_mode_regenerate`'s backup/rollback boundary, so an inline append would leave a false "freshly rotated" line for any identity already processed before a *later* identity's seed generation raises — `_restore_on_failure` reverts the seeds to their pre-rotation state, but the log would still (falsely) claim a fresh rotation happened, suppressing `check_seed_age.py`'s WARN/FAIL for that identity for up to ~75-90 days. Redesigned to buffer every identity's pending log entry in a list and only flush (`rotation_log_append()`) after *every* identity's seed and `auth.conf` write has durably succeeded — a mid-loop failure now yields **zero** log entries for that run instead of a partial, misleading set. This is the most consequential implementation change vs. the approved plan and should have been listed here at PR-open time alongside items 1-7; added retroactively per `/code-review` follow-up.
   This redesign has a second, co-dependent half in `src/factory/operator_audit.py`: `_run_bash` was hardened to fail-soft — added `_BRIDGE_TIMEOUT_S = 10` and wrapped the subprocess call in `except (OSError, subprocess.TimeoutExpired): pass` — so a hang or OS error during the post-seed flush degrades to a missing log line rather than propagating. Without this, a flush-time failure occurring *after* seeds are durably written would look identical to a real seed-generation failure and could wrongly trigger `_mode_regenerate`'s rollback of already-good seeds (the reverse of the forward invariant above). Also omitted at PR-open time; added retroactively per `/code-review` follow-up (2026-07-06).

Fixes #2246

---


MickaelV0:
## Code Review

Multi-domain review (architect, security-auditor, backend-dev, tester, product-lead, devops) against `6628f018c` (HEAD). One blocking finding was found, fixed, and verified during this review; one design-level gap remains open for a human call before merge.

**Verdict: Request changes** — not because of an open defect in the shipped code, but because one real architectural gap (below) needs an explicit human decision (fix now vs. accept-and-document) before this is mergeable as a security-hardening PR.

### Already fixed during this review

**issue(blocking) → fixed in `6628f018c`**: `tests/scripts/test_genkeys_seeds_dir.py` (pre-existing file, not otherwise touched by this PR) calls `_mode_full_provision()` in-process twice (`test_full_provision_rootless_does_not_call_require_root`, `test_full_provision_opt_in_calls_require_root`) against a fixture matrix containing `"hub"` — a real production identity name. Both tests overrode `SEEDS_DIR`/`AUTH_DIR` but not `ROTATION_LOG`/`OPERATOR_LOG`. Since this PR wires `rotation_log_append()` into `_mode_full_provision`'s per-identity loop, every run of these two tests wrote a real `secret:hub | reason:seed-generated | trigger:factory-acl-genkeys` line to the developer's actual `~/.roxabi/factory/rotation-log.md` — directly contradicting this PR's own SC10 claim ("no real-path writes").
- Found by: product-lead (96% confidence). Independently verified by direct read of both call sites, and empirically: **6 fabricated `secret:hub` entries had already accumulated in the real file** during this session's own test runs.
- Fix: added the same `ROTATION_LOG`/`OPERATOR_LOG` → `tmp_path` override pattern already used elsewhere in this PR to both tests (`tests/scripts/test_genkeys_seeds_dir.py`, commit `6628f018c`).
- Cleanup: the 6 fabricated lines in the real (out-of-repo) `~/.roxabi/factory/rotation-log.md` were removed, restoring the file to header-only — matching the documented pre-PR bootstrap state (zero log entries across all identities).
- Verified: `tests/scripts/test_genkeys_seeds_dir.py` 10/10 pass; full repo-wide suite (6507 passed) re-run twice post-fix; real file confirmed unchanged at 35 bytes (header only) both times. `ruff format`/`ruff check`/`pyright` clean on the touched file.

### Open — needs a human decision before merge

**issue (candidate/log-rollback-desync)**: `rotation_log_append()` is now called inside `_mode_full_provision`'s per-identity loop (`scripts/_modes.py:594`), and `_mode_regenerate` (`scripts/_modes.py:471`) calls that loop under a seed-only backup/rollback boundary (`_backup_seeds`/`_restore_on_failure`). If seed generation fails partway through (e.g. identity #7 of 18), identities #1–6 already have `rotation_log_append()` fire with today's date **before** the exception propagates and seeds are restored to their pre-rotation values. `rotation-log.md` is append-only and outside the backup/restore transaction, so the log permanently claims those identities were freshly rotated even though their seed reverted. `check_seed_age.py` then reads that false-fresh date and suppresses WARN/FAIL for a seed that never actually changed — in the exact DR/security-emergency scenario (`--regenerate`, e.g. after a suspected compromise) that this feature exists to gate. This is not indefinitely silent (the false entry ages out and starts warning again after 75–90 days from the fake date), but if the failed regenerate isn't retried immediately, a genuinely-compromised, never-rotated seed can read as fresh for up to ~3 months.
- Found by: backend-dev (72% confidence).
- Root cause: this PR's spec explicitly reasoned about the DR happy-path ("18 granular lines for free — a net improvement") but the edge-case table only covers `rotation_log_append()` itself failing softly — it never considered a *later* step in the same loop failing *after* logging already succeeded. No existing test exercises a real (non-mocked) failure partway through `_mode_full_provision`'s loop; `test_regenerate_restores_on_provision_failure` fully mocks out `_mode_full_provision` with an immediate `side_effect`, so it never reaches the per-identity loop or a partial log write.
- Options (from backend-dev, not applied — this is the human decision point):
  1. Buffer `(name, reason, trigger)` tuples during the loop and only call `rotation_log_append()` for all of them once the entire operation is confirmed durable (recommended, but a real design change to `_mode_full_provision`/`_mode_regenerate`'s interaction — out of scope for an unattended fix).
  2. Accept the risk explicitly: document in `docs/runbooks/nkey-rotation.md` that a partial-failure-then-rollback during `--regenerate` can desync the log from real seed state, and that an operator should re-run `check-seed-age` with fresh eyes after any DR failure.
- Not fixed here: this is a design-level tradeoff on the DR/security-emergency path, not a mechanical hermeticity gap — it belongs at a human go/no-go decision, not an autonomous fix mid-pipeline.

### Non-blocking notes (accept as documented tradeoffs, or fold into follow-up at maintainer's discretion)

- **note (fail-open enforcement, 72%)** — `check_seed_age.py`: every seed minted *before* this PR ships has zero rotation-log entries and sits in permanent bootstrap grace until its first post-deploy rotation; there's also no reconciliation/alert if `rotation_log_append` ever fails-soft (bridge is `check=False`, bash side `|| true`), so a silently-failed append also yields permanent grace. The control can pass green forever while genuinely stale credentials exist, with nothing surfacing that gap. — security-auditor. Suggested: backfill a dated `bootstrap-backfill` log entry per active identity on first post-deploy run, or emit a distinct "N identities never logged" counter.
- **suggestion (55%)** — `check_seed_age.py:52-55,118-121`: `_latest_log_dates()` and the mtime fallback both catch bare `OSError`, conflating "file doesn't exist yet" (intended grace) with permission errors / disk I/O failures — a broken read silently degrades to the same ungated grace with no operator-visible signal the check itself is broken. — backend-dev. Suggested: narrow to `FileNotFoundError`, let other `OSError` subclasses propagate or surface as a distinct notes entry.
- **suggestion (55%)** — `test_genkeys_modes.py:1095`, `test_genkeys_modes_add_identity.py:237`: wiring tests assert only on `secret:{name}` substring, never on `reason:`/`trigger:` values, even though the spec pins `trigger` to differ by call site (`factory-acl-add-identity` vs `factory-acl-genkeys`). Swapping the two trigger strings would stay green. — tester.
- **suggestion (45%)** — `src/factory/operator_audit.py:19` (file not touched by this PR): `_run_bash`'s `subprocess.run` has no `timeout=`; this PR raises its call frequency from 1 (DR-only) to up to 18/invocation on the routine hot path, so a hang now blocks `factory-acl genkeys` itself. — backend-dev.
- **nit (55%)** — `deploy/systemd/factory-check-seed-age.service`: `Type=oneshot` with no `RemainAfterExit=` means a genuine stale-seed FAIL leaves the unit `failed` for up to 24h, folding into `systemctl --user is-system-running` → `degraded` — worth confirming this is the intended signal (arguably desirable) and documenting it in the runbook so it isn't mistaken for unrelated drift during a future M1 triage. — devops.
- **nit (45%)** — `deploy/systemd/factory-check-seed-age.timer`: `After=`/`Wants=network-online.target` was copied from the fleet-digest-poll shape reference but this checker reads only local files — inert-but-harmless. — devops.
- **suggestion (25%)** — `scripts/gen_nkeys.py:187`: `_cmd_check_seed_age`'s exit-code mapping has no CLI-layer test (only the underlying pure `run()` is tested) — consistent with the existing convention for sibling `check` subcommands, flagged for visibility only. — tester.

### Praise (representative, not exhaustive)

- Seed/key material never reaches the audit log or the checker — only identity names, dates, fixed reasons/triggers, and mtimes flow through; `check_seed_age.run()` never opens seed file contents. — security-auditor
- The mtime-informational-only redesign (bootstrap grace: log-gates the verdict, mtime is a non-gating `notes` hint) is faithfully and completely implemented with no leak path from mtime into `warnings`/`failures`. — architect
- Date-boundary math (74/75/89/90/91-day edges, UTC-anchored throughout) is correct and the parametrize table is hardcoded rather than derived from the constants under test — genuinely regression-resistant. — tester
- `check_seed_age.py`'s CLI wiring (`_cmd_check_seed_age`) faithfully mirrors the existing `_cmd_check_grants` pattern; `_repo_root()` correctly avoids a circular import by not reaching into `gen_nkeys.py`. — backend-dev
- `deploy/systemd/factory-check-seed-age.{service,timer}` is a directive-by-directive match of the correct shape reference (`factory-fleet-digest-poll`, not `factory-operator-logrotate`, which would have left `acl-matrix.json`'s cwd-relative default unresolvable under systemd); correctly never wired into `.claude/stack.yml` or CI (scheduled-only, not a merge gate). — devops
- Full hermeticity audit of both genkeys wiring test files (11+ subprocess `env={}` dicts, all `monkeypatch.setenv` sites) confirms every code path capable of calling `rotation_log_append()` is isolated; the sole gap was the file above, now fixed. — tester

### CI

Green on `6628f018c` (`ci`, `trufflehog`, `tests`, `gates`, `integration`, `docker-build`, all required + non-required checks pass). First run on the prior commit hit the pre-existing documented flaky `test_post_autoupdate_contract`; reran clean, confirmed unrelated via file-scope diff.

MickaelV0:
## Code Review (iteration 2 — fresh pass, fix→review loop cap reached)

Fresh review against `f0d5648a7` (HEAD), which addressed the prior review's Blocker #2 (log/rollback desync) and all non-blocking notes. Reviewed as security-auditor / backend-dev / tester / devops / architect lenses; no implementation context carried over from the fix pass.

**Verdict: Approve**

### Verification performed (not just re-reading the diff)

- **Real-file hermeticity**: `~/.roxabi/factory/rotation-log.md` confirmed unchanged (35 bytes, header-only) before and after running the full local suite twice, including the two new mid-loop-failure regression tests.
- **Full local test suite**: 6509 passed, 26 skipped, 1 xfailed (isolated via `PYTHONPATH=<worktree>/src` + venv `bin/` on `PATH`, since the shared cross-worktree `.venv`'s editable install/console-script shebangs were pointing at a sibling worktree at the time — a known environment gotcha, not a code defect; re-verified against *this* worktree's actual source after diagnosing it).
- **`ruff check` / `ruff format --check`** on all touched files: clean.
- **`pyright`** on all touched files: 0 errors, 0 warnings.
- **Flush-placement soundness (the load-bearing claim of the fix)**: read every line of `_mode_full_provision` between the flush loop (`scripts/_modes.py:634-635`) and the function's `return` (line 642) — it's pure in-memory dict/list iteration over already-validated data (`identity.get("deploy")`), no I/O, no subprocess, cannot raise. `_repo_root()` is a pure `Path` computation. `_run_bash` now catches `(OSError, subprocess.TimeoutExpired)` around a `timeout=10` `subprocess.run`, so a flush-time logging failure degrades to a missing log line rather than propagating. The comment's claim ("nothing below this line can fail in a way that triggers rollback") holds concretely in this codebase today, not just as an aspiration.
- **Doc-claim cross-check**: `deploy/lib/operator-log.sh`'s `|| true` fail-soft behavior (cited in the new runbook section) verified directly against the script (lines 64-65, 83-86). `docs/runbooks/operator-log.md` / `deploy/AGENTS.md` no longer contain "future nkey rotations" (SC12). `nkey-rotation.md` documents auto-logging + bootstrap grace (SC11).
- **CI**: confirmed green on `f0d5648a7` — `ci`, `tests`, `gates`, `integration`, `docker-build`, `trufflehog`, all required + non-required checks `SUCCESS`, zero non-passing.
- **Pre-existing dead-code interaction re-checked**: `_restore_on_failure`'s `not seeds_dir.exists()` guard (flagged as pre-existing/out-of-scope in the fix commit) not actually reverting a seed after a real mid-loop failure does **not** reopen Blocker #2 — even if the on-disk seed ends up "new" while the log has zero entries for it, `check_seed_age.py` then either applies bootstrap grace (first-ever rotation) or falls back to a *stale* prior log date (subsequent rotation), which biases toward WARN/FAIL, never toward false-fresh. The failure mode is fail-safe in the direction this feature exists to guarantee.

### Notes (non-blocking, no action required)

- **thought (40%)** — `scripts/_modes.py`: the buffer-then-flush comment block explains *why* the boundary is safe but doesn't name the two-part invariant test pair (`test_regenerate_mid_loop_failure_writes_no_rotation_log_entries`, `test_full_provision_log_failure_does_not_trigger_seed_rollback`) by name in a cross-reference — a future editor touching this loop has to rediscover them. Optional: add a one-line `# see tests/scripts/test_genkeys_modes.py::TestRegenerateMode` pointer. Not required; the docstring-level explanation is already unusually thorough.
- **praise** — The two new regression tests exercise a *real* mid-loop failure (a `FakeNkeyProvider` subclass that raises on the second identity, and a real `subprocess.run` monkeypatch raising `TimeoutExpired`) rather than mocking out the function under test — exactly the gap the original review flagged in the pre-existing `test_regenerate_restores_on_provision_failure` (full mock, never reaches the loop).
- **praise** — Narrowing `except OSError` → `except FileNotFoundError` in both `check_seed_age.py` call sites is the correct fix for the security-auditor's "broken check silently degrades to green" concern — a real permission/I/O error now propagates instead of being swallowed into bootstrap grace.
- **praise** — Every fail-open gap in the logging path (fail-soft bridge, buffer-then-flush's own tradeoff window, narrowed `OSError` handling, `degraded`-on-FAIL systemd behavior) is now documented in `nkey-rotation.md` with the specific operator-facing consequence, not just "here be dragons."

### Not fixed (carried from iteration 1, still correctly out of scope)

- `_restore_on_failure`'s seed-rollback guard being dead code on any real mid-loop failure — pre-existing bug (predates this PR, confirmed against `staging` merge-base), independently re-verified in this pass not to undermine this PR's own invariant (see above). Correctly flagged for a separate human decision rather than folded into this security-hardening PR's scope.
- Fail-open bootstrap-grace blind spot (pre-#2246-rollout seeds sit in permanent grace until first rotation; a silently-failed append also yields permanent grace with no reconciliation/alert) — addressed via explicit runbook documentation of the operator-facing consequence rather than implementing the security-auditor's suggested `bootstrap-backfill` log entries or a distinct "N identities never logged" counter. Reasonable for this issue's stated scope (a logging/checking mechanism, not a backfill/alerting feature); flagged in the runbook as something an operator should know to cross-check manually, not silently dropped.

### CI

Confirmed green on `f0d5648a7` at review time (`ci`, `tests`, `gates`, `integration`, `docker-build`, `trufflehog` — all `SUCCESS`).

MickaelV0:
## Review Fixes Applied

**Applied (minor, unclassified):** 2 findings
**Deferred (issues created):** 0 findings
**Skipped:** 0 findings
**Failed:** 0 findings

### Auto-Applied
- [applied] suggestion(blocking): `factory-acl check seed-age` never prints a CLI success line, unlike every other `check` subcommand (`gen_nkeys.py:191`) — added `check-seed-age: OK ({N} warning(s))` (92%, 2 agents)

### Applied (1b1)
- [applied] issue: `docs/runbooks/nkey-rotation.md` overstated `--regenerate`'s rollback guarantee on a genuine mid-loop failure — corrected to state this is a **pre-existing gap, predating this feature, not yet tracked** (no issue filed — see Deferred note below)
- [applied] suggestion(blocking): `factory-acl check seed-age` had no CLI-level (subprocess) test coverage — the whole flow was only exercised via internal `run()` calls, never through `argparse` + the actual entry point. Added `TestCliWiring` (4 tests) in `tests/scripts/test_check_seed_age.py`, invoking `python -m scripts.gen_nkeys check seed-age` via `subprocess.run` with `tmp_path`-scoped `ROTATION_LOG`/`SEEDS_DIR` env overrides. Falsification-verified: breaking the CLI's success-print line via a deliberate mutation correctly failed 3/4 of the new tests before the fix was restored (75%)

### Applied (minor)
- [applied] nitpick: `Makefile` `.PHONY` was missing `check-seed-age`
- [applied] nitpick: stale "RED (#2246)" framing left over from TDD authoring removed from section headers/docstrings in `test_genkeys_modes.py` and `test_genkeys_modes_add_identity.py`

### Deferred (not filed — awaiting human decision)
- `_restore_on_failure`'s guard at `scripts/_modes.py:436` — a pre-existing (predates this feature) gap where `--regenerate`'s backup/restore path does not guarantee an automatic seed rollback on a genuine mid-loop failure. This is now accurately described (not overstated) in `docs/runbooks/nkey-rotation.md`, but **no tracking issue has been filed** — filing one requires an explicit human decision, out of scope for this `/fix` cycle.

### Verification
- `ruff format --check` / `ruff check` — clean (2 line-length violations fixed during 1b1 apply)
- `python -m pyright` on all 5 touched Python files — 0 errors, 0 warnings
- `pytest tests/scripts/test_check_seed_age.py` — 14/14 passed (incl. 4 new CLI-wiring tests)
- Full suite (`pytest tests/`) — 5521 passed, 26 skipped, 1 xfailed, run twice; two isolated one-off failures both diagnosed as unrelated to this change (a `PATH`-artifact of test invocation, and a pre-existing flaky e2e/subprocess test), neither reproduces in isolation
- Hermeticity: `~/.roxabi/factory/rotation-log.md` MD5 unchanged across all test runs — new CLI tests never touch the real production rotation log

MickaelV0:
## Code Review (iteration 3 — 6-agent pass at f0d5648a7→db0492bb9, loop cap reached)

Full 6-domain pass (security-auditor, backend-dev, architect, tester, devops, product-lead) run fresh against the latest diff, after two prior fix→review cycles already resolved all blocking findings. Applying a stricter bar at this pass: only genuine correctness/security regressions get fixed; nitpicks, style, and coverage-gap suggestions are recorded here as informational rather than triggering a 4th round.

### Applied
- **issue (72%, devops):** `docs/runbooks/operator-log.md:76-77` claimed "—" (no operator.log event) for the two new rotation-logging paths, but `rotation_log_append()` unconditionally emits an `op_log rotation_log ...` event (`deploy/lib/operator-log.sh:87`, `src/factory/operator_audit.py:59`). Fixed in `db0492bb9`.

### Praise
- **security-auditor:** fully clean — no blocking or non-blocking security findings. Ship it.
- **architect (92%):** fail-soft isolation of the rotation-log write path (`_run_bash` timeout+except, buffer-then-flush in `_mode_full_provision`) is correct and well-reasoned; no circular deps; no target-axis-trap.
- **product-lead (92%):** all 13 spec Success Criteria independently re-verified by direct source inspection (not PR-body trust); 73 tests pass; ruff/pyright clean; runbook claims cross-checked against implementation — no drift beyond the item noted below.
- **tester:** buffer-then-flush zero-entries invariant and the full 74–91 day boundary matrix independently re-traced and confirmed non-tautological.

### Noted, not fixed (below auto-apply bar, non-blocking)
- **backend-dev (35%):** `check_seed_age.py:42` regex could theoretically over-capture on a hand-corrupted log line; unreachable in practice (sole writer always emits well-formed lines), fails safe if triggered.
- **architect (70%/66%):** `check_seed_age.py` mirrors `check_grants.py` at the `run()`-contract level only, not module-shape; `factory-check-seed-age.service` shells to `make` where sibling units shell to a script. Both self-assessed "not a defect" — intentional, defensible design choices; spec wording could be softened for precision.
- **tester (75%/60%/45%):** three test-coverage gaps on already-correct behavior — `_BRIDGE_TIMEOUT_S` kwarg not asserted, `OSError` arm of a guard untested, narrow `except FileNotFoundError` not proven to stay narrow. Code is correct; tests just don't pin these specific branches.
- **devops (55%/45%):** `factory-check-seed-age.service`/`.timer` omit `After=/Wants=network-online.target` and use the bare `OnCalendar=daily` alias without a comment, unlike sibling units. Cosmetic convention drift; functionally inert since the check is local-filesystem-only.
- **product-lead (80%):** `docs/runbooks/nkey-rotation.md:242`'s "no tracking issue filed yet" disclosure (for the pre-existing, out-of-scope `_restore_on_failure` dead-guard gap at `scripts/_modes.py:436`) will go stale once/if an issue is filed. Confirmed via `gh issue list` that no such issue exists yet. Not a merge blocker — filing a tracking issue is a human decision, not made here.

### Verdict
**Approve with comments.** No regressions found; one doc-accuracy fix applied; remaining items are non-blocking and recorded for optional future cleanup.

MickaelV0:
## Review Fixes Applied

**Skipped:** 8 finding(s) (nitpicks/coverage-gaps/notes, below actionability bar — see Code Review comment above for the full list)
**Failed:** 0 finding(s)

### Applied (1b1)
- [applied] issue: `docs/runbooks/operator-log.md:76-77` — corrected "operator.log event" column to reflect that `rotation_log_append()` unconditionally emits a `rotation_log` event (`db0492bb9`)

### Skipped
See "Noted, not fixed" section in the Code Review comment above — 8 findings (1 backend-dev, 2 architect, 3 tester, 2 devops), none meeting the C≥80 + 2-agent auto-apply bar, none representing correctness/security regressions. One additional finding (product-lead, 80%) flags a documentation loose-end contingent on a human decision to file a tracking issue for a pre-existing, out-of-scope bug — not acted on here.

MickaelV0:
## Code Review

**Verdict: Request changes**

6 domain agents reviewed (security-auditor, architect, product-lead, tester, backend-dev, devops). Spec compliance is strong (12/13 SCs independently re-verified, targeted suite re-run: 73 passed) and no security/architecture-blocking issues were found. One real blocker: the PR is now out of sync with `staging`.

### Blockers

- **issue: PR is `CONFLICTING` with staging (`mergeStateStatus: DIRTY`)** — independently verified (`gh pr view --json mergeable,mergeStateStatus`). `scripts/_modes.py` and `tests/scripts/test_genkeys_modes.py` were also touched by PR #2263 (merged to `staging` ~1h after this branch's last commit — verified via `gh pr view 2263 --json files`, exact same 2 files). `docs/runbooks/nkey-rotation.md:242` and a test comment (`tests/scripts/test_genkeys_modes.py:1638`) still present-tense-disclose the `_restore_on_failure` dead-guard gap as open/unfixed and "tracked in #2258" — but #2258 is now **CLOSED** (fixed via #2263).
  -- product-lead (90%), independently corroborated
  Solutions:
  1. Rebase/merge `staging` into this branch, resolve the `_modes.py` + test-file conflicts against #2263's fix, then flip the `nkey-rotation.md` caveat + test comment to past tense (recommended)
  2. Merge as-is post-conflict-resolution and file a fast-follow doc-drift issue for the caveat text

### Warnings / Suggestions

- **issue (non-blocking): future-dated rotation-log entry permanently suppresses the seed-age gate (fail-open)** — a single well-formed line with a future date wins the "latest" comparison with no upper-bound check, silently suppressing WARN/FAIL for that identity forever, no operator-visible signal.
  `scripts/check_seed_age.py:71,107` -- security-auditor (65%)
  Solutions: 1. Reject/ignore `log_date > today` as malformed (recommended) 2. Surface as a `NOTE:` instead

- **suggestion: undisclosed plan deviation** — the buffer-then-flush redesign of N2 (deferring `rotation_log_append()` until after the seed loop, to avoid poisoning the log on a rolled-back mid-loop failure) is the most consequential implementation change vs. the approved plan, but wasn't listed among the PR's 7 declared "judgment calls" alongside lower-stakes ones that were.
  `scripts/_modes.py:598` -- product-lead (75%)

- **thought: flush-ordering invariant is unenforced** — correctness of the buffer-then-flush design depends on nothing after the flush (still inside `_mode_regenerate`'s rollback boundary) ever raising. Holds today (pure dict iteration) but nothing guards a future edit from moving raising code into that tail.
  `scripts/_modes.py:637` -- architect (60%)
  Solutions: 1. Add a guard comment stating the invariant (recommended) 2. Extract to a provably-last-statement helper

- **nitpick (parallel-path-drift): `ROTATION_LOG` default path duplicated across bash and Python** with no test pinning them together — if the bash default ever moves, the Python reader silently reads a nonexistent file and every identity falls into bootstrap-grace (silent fail-open).
  `scripts/gen_nkeys.py:175`, `deploy/lib/operator-log.sh:13` -- architect (72%)
  Solutions: 1. Add a test asserting the two defaults match (recommended) 2. Resolve `ROTATION_LOG` by sourcing `operator-log.sh`

- **nitpick: untested error-propagation contract** — the deliberate narrow `except FileNotFoundError` (a broken log read must crash loudly, not degrade into bootstrap grace) has no test proving that contract; a natural-looking future refactor (`except OSError`/`except Exception`) would silently reintroduce the masked-failure risk.
  `scripts/check_seed_age.py:56,58` -- tester (55%)
  Solutions: 1. Add a test that forces a PermissionError/OSError and asserts `run()` raises (recommended)

- **nitpick (parallel-path-drift): `_repo_root()` duplicated** between `_modes.py` and `gen_nkeys.py` (cycle-constrained, so duplication is currently justified) — flagging for awareness if `scripts/` layout ever changes.
  `scripts/_modes.py:84` -- architect (50%)

- **nitpick: dangling doc reference** — Makefile comment cites "(see docs/runbooks)" for the M1 uv.lock-churn incident, but no runbook file actually documents it.
  `Makefile:357` -- devops (70%)
  Solutions: 1. Drop the parenthetical (recommended) 2. Add the missing writeup

- **thought: ops-signal conflation** — the new systemd unit adds a second, unrelated source of the `degraded` `systemctl --user is-system-running` aggregate signal (already used for M1 deploy-health). Documented/intentional, flagging for the record only.
  `deploy/systemd/factory-check-seed-age.service:10` -- architect (45%)

- **nitpick: single-layer `--frozen` guard** — unlike the sourced deploy-path convention (`UV_FROZEN=1` env + `--frozen` flag, belt-and-suspenders), this unit relies on the `--frozen` flag alone.
  `deploy/systemd/factory-check-seed-age.service:13`, `Makefile:361` -- devops (40%)
  Solutions: 1. Add `Environment="UV_FROZEN=1"` to the unit (recommended)

- **nitpick: wall-clock dependency** in 3 `TestCliWiring` tests instead of the file's fixed `_TODAY` convention — low actual flake risk (offsets have wide margin from the 75/90d boundaries).
  `tests/scripts/test_check_seed_age.py:388,412,432` -- tester (30%)

- **nit (informational): comment overstates restore behavior** — "`_restore_on_failure` reverts it to its pre-rotation value" overstates what happens given the (now-fixed-upstream, not-yet-in-this-branch) dead guard; the property that actually holds is "zero rotation-log entries get flushed." Comment-accuracy only.
  `scripts/_modes.py:586-597` -- backend-dev

### Praise

- Spec compliance independently re-verified: 12/13 SCs confirmed against spec text + live code + a fresh test run (73 passed), not just the PR's self-report — product-lead
- `pending_rotations` buffer-then-flush design is transactionally correct, not just defensive (verified against `_mode_regenerate`'s rollback boundary) — architect
- Test quality: tester empirically broke 3 safety-critical guards (WARN/FAIL boundary, bootstrap-grace-never-uses-mtime, buffer-then-flush) and confirmed each has a genuinely falsifying test, then restored and re-verified green
- No security-blocking issues: no injection, auth bypass, secret exposure, or privilege escalation; `rotation_log_append()` never leaks seed material — security-auditor
- No architecture-blocking issues: clean layering, no cycles, no double-logging vs. the existing DR call site — architect
- Systemd unit correctly mirrors `factory-fleet-digest-poll`'s shape, `--frozen` present, registered in `quadlet-sync-install`, correctly excluded from CI/merge gates — devops
- Fail-soft error handling in `rotation_log_append()` call sites correctly implemented and tested; no bare/overly-broad excepts introduced — backend-dev
- The PR's own "judgment calls" disclosure (1-7) is reasonable — mostly already pre-approved by the spec, not unilateral scope creep — product-lead

---

MickaelV0:
## Review Fixes Applied

**Blocker resolved:** PR was `CONFLICTING` with `staging` (sibling PR #2263 touched the same 2 files). Rebased `feat/2246-enforce-nkey-rotation-policy` onto `origin/staging`, manually resolved the textual-adjacency conflict in `tests/scripts/test_genkeys_modes.py` (both PRs independently added different test methods at the same insertion point — kept all 4), and corrected a test comment + `docs/runbooks/nkey-rotation.md:242` that still described #2258's dead-guard bug as open ("tracked in #2258") when it's now fixed (via #2263, pulled in by this rebase). Full validate (lint + format + 6515 tests + pyright) re-run clean post-rebase; force-pushed.

**Auto-applied (C≥80, 2+ agents):** 0 findings
**Applied via 1b1:** 5 findings
**Deferred (issues created):** 0 findings
**Skipped:** 5 findings (3 low-confidence/reviewer-endorsed-as-is, 2 non-actionable `thought` category)
**Resolved incidentally by rebase:** 1 finding
**Failed:** 0 findings

### Applied (1b1)

- [applied] issue (non-blocking): future-dated rotation-log entry silently suppressed the seed-age gate forever — `_latest_log_dates` now rejects/ignores `log_date > today` as malformed. `scripts/check_seed_age.py`
- [applied] nitpick: untested error-propagation contract — added `TestErrorPropagation`, which forces a `PermissionError` and asserts `run()` raises rather than degrading into bootstrap grace. `tests/scripts/test_check_seed_age.py`
- [applied] nitpick (parallel-path-drift): `ROTATION_LOG` default path duplicated across bash/Python with no parity check — added `TestRotationLogDefaultPathParity`, which extracts both literals from source and asserts they resolve identically under the same `HOME`. `tests/scripts/test_check_seed_age.py`
- [applied] nitpick: dangling doc reference — dropped the "(see docs/runbooks)" parenthetical citing a runbook that doesn't document the incident. `Makefile:357`
- [applied] suggestion: undisclosed plan deviation — added judgment call #8 to the PR body documenting the buffer-then-flush rotation-log redesign (the most consequential deviation from the approved plan, previously undisclosed).

Both new tests were falsified: reverting each fix reproduces the exact bug the test targets (future date wins forever / error silently swallowed / bash-Python drift undetected), confirming they're not tautological.

### Skipped

- nitpick (parallel-path-drift): `_repo_root()` duplicated between `_modes.py` and `gen_nkeys.py` — reviewer's own note: "duplication is currently justified" (cycle-constrained); flagged for awareness only, no action needed.
- nitpick: single-layer `--frozen` guard on the new systemd unit (40% confidence, speculative) — low-value belt-and-suspenders addition for a brand-new unit; not worth the inconsistency-avoidance tradeoff at this confidence level.
- nitpick: wall-clock dependency in 3 `TestCliWiring` tests (30% confidence) — reviewer's own note: "low actual flake risk" (wide margin from the 75/90d boundaries).
- thought: flush-ordering invariant unenforced — `thought` is a non-actionable category by the `/fix` taxonomy (discussion/awareness, not auto-fix scope).
- thought: ops-signal conflation (`degraded` aggregate) — same: non-actionable category, documented/intentional per the finding itself.

### Resolved incidentally by rebase

- nit (informational): `scripts/_modes.py:586-597`'s comment "reverts it to its pre-rotation value" previously overstated `_restore_on_failure`'s actual (buggy) behavior — now accurate, since #2263's dead-guard fix landed via this rebase.

---

MickaelV0:
## Code Review

**Verdict: Approve with comments**

6 fresh domain agents re-reviewed post-rebase (security-auditor, architect, product-lead, tester, backend-dev, devops) — no implementation context, full 2210-line diff each. No blocking findings. Spec compliance strong (12/13 SC rows independently re-verified). All three tests added by the prior `/fix` round (`TestErrorPropagation`, `TestRotationLogDefaultPathParity`) were empirically re-falsified (guard removed → test correctly fails, restored → green) — not tautological. The 4-way rebase-merged test conflict in `test_genkeys_modes.py` was checked for duplication/overlap — clean.

### Warnings / Suggestions

- **suggestion: `_latest_log_dates`'s future-date-rejection guard (added by this session's own prior `/fix` round) has zero test coverage** — independently found by 3 agents, empirically confirmed (removing the guard, full suite still passes).
  `scripts/check_seed_age.py:72` -- tester (92%), backend-dev (85%), product-lead (85%)
  Root cause: the guard itself was a review-driven addition; the fix round that added it didn't pair it with a regression test.
  Solutions:
    1. Add a test asserting a future-dated log line is rejected/ignored rather than winning "latest" (recommended)
    2. Defer — low risk, but next refactor of this function has nothing to catch a regression

- **suggestion: seed-mtime informational read only catches `FileNotFoundError`, asymmetric with the log-read path's deliberate fail-loud contract, and untested** — architect flagged the consequence (a `PermissionError` here would propagate uncaught, unlike the log-read path which is deliberately narrow); tester confirmed no test would catch a regression either way.
  `scripts/check_seed_age.py:95,127-129` -- tester (78%), architect (35%, as `thought`)
  Solutions:
    1. Widen to match the log-read path's contract (or explicitly document why asymmetry is intended) + add a test (recommended)

- **suggestion: PR body's "Judgment calls" #8 discloses only the buffer-then-flush half of this session's redesign, omitting the `operator_audit._run_bash` timeout/fail-soft hardening that the flush's correctness now depends on** — the reverse invariant (a flush-time `_run_bash` failure must not falsely mark a rollback as a fresh rotation) isn't mentioned.
  PR body, judgment call #8 -- product-lead (80%)
  Solutions:
    1. Amend judgment call #8 to name both halves of the design (recommended)

- **suggestion: 2 rebase-inherited tests (`test_regenerate_restores_on_genuine_mid_loop_failure`, `test_regenerate_restores_etc_auth_on_genuine_mid_loop_failure`, from sibling PR #2263) meet this PR's own hermeticity criterion (env-scoped `ROTATION_LOG`/`OPERATOR_LOG`) by accident of where their induced failure lands, not by construction** — safe today because the failure occurs before the post-loop flush point; a future refactor moving that boundary could silently leak writes to the real files.
  `tests/scripts/test_genkeys_modes.py` (TestRegenerateMode) -- product-lead (80%)
  Solutions:
    1. Add an explicit env-scope assertion/fixture so hermeticity is enforced by construction, not incidentally (recommended)

- **note: `ROTATION_LOG` default path literal duplicated between `scripts/gen_nkeys.py` and `deploy/lib/operator-log.sh`** — already mitigated by `TestRotationLogDefaultPathParity` (added in the prior `/fix` round), which pins both literals together. Informational only, no action needed.
  `scripts/gen_nkeys.py:175`, `deploy/lib/operator-log.sh:13` -- security-auditor (66%); praised independently by architect (90%) as correctly mitigated

- **thought: `operator_audit._run_bash`'s fail-soft exception-swallow (`except (OSError, subprocess.TimeoutExpired): pass`) is now load-bearing for `_modes.py`'s buffer-then-flush correctness, but the dependency is invisible from `_modes.py` alone** — nothing documents that the flush's safety depends on this function never raising.
  `src/factory/operator_audit.py` (`_run_bash`), referenced from `scripts/_modes.py:645` -- architect (62%, `candidate/fail-soft-contract-load-bearing`), same theme independently noted by backend-dev (`candidate/silent-fail-soft-no-breadcrumb`)
  Solutions:
    1. Add a comment at the `_run_bash` call site in `_modes.py` naming the invariant it depends on (recommended)

- **nitpick: buffer-then-flush comment/docstring (`_modes.py:597-608`, mirrored in `docs/runbooks/nkey-rotation.md:242,245`) overstates the rollback-boundary guarantee** — true only for the `--regenerate`-wrapped call path; the bare `factory-acl genkeys` default-dispatch path (`gen_nkeys.py:79`) isn't covered by the same boundary.
  `scripts/_modes.py:597-608`, `docs/runbooks/nkey-rotation.md:242,245` -- backend-dev (55%)
  Solutions:
    1. Scope the comment's claim to the `--regenerate` path explicitly (recommended)

- **suggestion: PR body's Test Plan / SC10 numeric claims are stale relative to the current (post-rebase, post-fix) diff** — SC10's cited call-site count and the Test Plan's "57 passed" no longer match a live run of the same command.
  PR body -- product-lead
  Solutions:
    1. Refresh the counts before merge (recommended, cheap)

- **thought: the new systemd unit's stale-seed FAIL escalates `systemctl --user is-system-running` to `degraded` for up to 24h** — documented behavior, not a bug, but wasn't an explicit ask in the issue; worth a conscious product sign-off that this is the intended blast radius for a policy-only check.
  `deploy/systemd/factory-check-seed-age.service` -- product-lead
  Solutions:
    1. Confirm intended, or scope the escalation differently (e.g. a separate non-`degraded` signal) — no action required if confirmed as-is

- **candidate/xdist-monkeypatch-flake: `TestErrorPropagation::test_permission_error_reading_rotation_log_propagates` failed once non-reproducibly across ~14 runs under default parallel xdist, always passed under `-n0`** — root cause unconfirmed (suspected monkeypatch/xdist worker interaction). Non-blocking.
  `tests/scripts/test_check_seed_age.py` -- devops (45%)

### Praise

- Spec compliance independently re-verified: 12/13 SC rows confirmed accurate against spec text + live code — product-lead
- `ROTATION_LOG` parity test properly implemented and effective — architect
- No security-blocking issues: no injection, auth bypass, secret exposure, or privilege escalation in the rebase-affected area — security-auditor
- No architecture-blocking issues: buffer-then-flush design verified transactionally correct against `_mode_regenerate`'s rollback boundary — architect
- Test quality: both `/fix`-round tests (`TestErrorPropagation`, `TestRotationLogDefaultPathParity`) confirmed genuinely falsifying via mutation, then restored clean — tester
- 4-way rebase-merged tests in `test_genkeys_modes.py` (`TestRegenerateMode`) verified non-overlapping, no duplication — tester
- Systemd unit + `quadlet-sync-install` Makefile target verified complete: correct unit copy, single daemon-reload, `enable --now`, passes `systemd-analyze verify` — devops

---

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 3 findings
**Applied via 1b1:** 3 findings
**Deferred (issues created):** 0 findings
**Skipped:** 4 findings
**Failed:** 0 findings

### Auto-Applied

- [applied] suggestion: `_latest_log_dates`'s future-date-rejection guard had zero test coverage (92%, 3 agents: tester/backend-dev/product-lead) — added `TestFutureDatedLogEntryRejected`; falsified (removing the guard reproduces the exact bug: a bogus future-dated line masks a genuinely stale entry, suppressing FAIL).
- [applied] suggestion: judgment call #8 in the PR body omitted the `operator_audit._run_bash` fail-soft/timeout half of the buffer-then-flush redesign (80%, product-lead, verified 80% by architect) — amended the PR body to disclose both co-dependent halves.
- [applied] suggestion: 2 rebase-inherited `TestRegenerateMode` tests met this PR's hermeticity criterion by accident of where their induced failure lands, not by construction (80%, product-lead, verified 92% by tester) — pinned explicit `ROTATION_LOG`/`OPERATOR_LOG` env overrides on both.

### Applied (1b1)

- [applied] suggestion: seed-mtime informational read had the same narrow-except contract as the log-read path but no test proving it (78%, tester; corroborated 35% by architect) — added `TestSeedMtimeErrorPropagation`, mirroring `TestErrorPropagation`; falsified (widening the except clause makes it fail with `DID NOT RAISE`).
- [applied] nitpick: buffer-then-flush comment overstated the rollback-boundary guarantee as universal (55%, backend-dev) — reworded to scope the `_mode_regenerate` rollback claim to the `--regenerate` path, noting the bare default-dispatch path still benefits from all-or-nothing flushing independent of any rollback. Comment/prose only, no logic change. (`docs/runbooks/nkey-rotation.md`'s mirroring text was already correct from an earlier commit in this PR — left untouched.)
- [applied] suggestion: PR body's Test Plan / SC10 / SC13 / Lifecycle counts were stale relative to the current diff (product-lead) — refreshed all four to the live re-run: targeted suite 57→69 passed, full suite 6507→6519 passed, SC10 call-sites 8+6→14+11, Lifecycle new-test count 11→15 (14 in `test_check_seed_age.py` + 1 in `test_genkeys_modes.py`).

All fixes independently verified by a fresh verifier (different domain than the originating finding) before commit — the two single-agent C≥80 findings were confirmed at 80% and 92% before promotion to auto-apply; the two genuinely falsifiable new tests were confirmed to fail when their guard is removed and restored clean.

### Skipped

- note: `ROTATION_LOG` default-path literal duplicated bash/Python (security-auditor, 66%) — already mitigated by the prior `/fix` round's `TestRotationLogDefaultPathParity`; no action needed.
- thought: `operator_audit._run_bash`'s fail-soft exception-swallow is load-bearing for the buffer-then-flush design but undocumented as such (architect 62% / backend-dev, `candidate/fail-soft-contract-load-bearing`) — `thought` is a non-actionable category per the `/fix` taxonomy (discussion/awareness, not auto-fix scope).
- thought: new systemd unit's stale-seed FAIL escalates `is-system-running` to `degraded` for up to 24h (product-lead) — a product sign-off question, not a code fix; documented/intentional per the finding itself.
- candidate/xdist-monkeypatch-flake: `TestErrorPropagation` failed once non-reproducibly across ~14 runs under default parallel xdist (devops, 45%) — non-reproducible, root cause unconfirmed; not actionable without a repro.

### Enforcement diagnostics

0 subsumption violations.

---

Issue #1035, opened 2026-05-03
epic(monitoring): Monitoring v2 — NATS event stream + Tauri desktop dashboard


MickaelV0:
**Supersedes (partially):** monitor portion of #701 — the `roxabi-ops/monitor/` task is now moot; host-timer monitoring is being replaced, not relocated. (#701's `roxabi-host.service` + cross-project Makefile relocation remain valid.)
**Related:** #667 (OTel observability pipeline — complementary stream of telemetry), #62 (Phase 4 — closed/frozen, partially superseded by this), #57 + #462 (NATS Prometheus exporter — infra-level, complementary)
**Frozen target:** `src/lyra/monitoring/` and `deploy/lyra-monitor.{service,timer}` — deprecation banners land in the cleanup PR concurrent with this issue. **Do not extend the host-timer monitor.**

## Problem

`src/lyra/monitoring/` is a host-side Python timer that runs every 5 min and pokes `systemctl --user`, `podman logs <ctr>`, `127.0.0.1:8443/health/detail`, `127.0.0.1:8222/varz`. Three structural problems:

1. **Host-only deps.** Now that Lyra runs in Quadlet containers, the monitor needs host tooling (systemd-user, podman, loopback) — it cannot be containerized cleanly without `--network=host` + mounting the podman socket + dbus, which erodes the security boundary for no real gain.
2. **Pull, not push.** Each tick re-queries the world from outside. Stale by up to 5 min, no real-time signal, redundant work even when nothing has changed.
3. **No human surface.** Layer 2 escalates anomalies to a Telegram message — that is the entire UI. No dashboard, no history view, no acknowledgement, no in-app alert muting, no per-source drill-down.

The cleanup PR landing alongside this issue stops/disables the timer on prod and adds deprecation banners. The code is preserved for the spec author of this epic to mine.

## Vision — Monitoring v2

A push-based, event-driven observability layer with a desktop dashboard:

| Layer | Role | Tech |
|---|---|---|
| **Source** | Lyra processes publish health/metric events to NATS (\`lyra.metric.*\`, \`lyra.event.*\`) | embedded in hub/adapters/clipool |
| **Stream** | NATS JetStream with retention (e.g. 24 h hot, 7 d warm) | already deployed |
| **Aggregator** *(optional)* | Lightweight reader → DuckDB ring buffer for history queries | Rust or Python sidecar |
[...]

Issue #280, opened 2026-03-16
feat: per-agent TTS configuration


MickaelV0:
## Problem

`TTSService` is instantiated **once globally** at startup, using only the first agent alphabetically that has a `[tts]` section. All other agents' TTS configs are silently ignored (warning logged, nothing more).

This means every agent shares the same voice, engine, and parameters — making per-agent voice identity impossible despite the infrastructure being mostly ready.

## What already exists

- `tts_json` column in the `agents` SQLite table ✅
- `AgentTTSConfig` dataclass with all voiceCLI fields ✅
- TOML `[tts]` section parsing + DB serialization ✅
- `apply_agent_tts_overlay()` helper ✅

## What needs to be built

1. **`TTSService.generate()` accepts a per-call config override** — pass `AgentTTSConfig | None` at call time, merge with global defaults
2. **Pass the current agent's `tts` config** at voice generation time (in the response pipeline)
3. **Remove the "first agent wins" logic** and the multi-agent `[tts]` startup warning
4. **Expose missing voiceCLI fields** in `AgentTTSConfig`: `exaggeration` (float), `cfg_weight` (float)
5. **`lyra agent edit`** — interactive editing of TTS fields (currently not supported)

## Example target config

```toml
[tts]
engine      = "qwen"
voice       = "Vivian"
language    = "French"
accent      = "Accent parisien"
personality = "Chaleureuse et directe"
speed       = "Rythme soutenu"
emotion     = "Légèrement enthousiaste"
exaggeration = 0.6
cfg_weight   = 0.5
```

Each agent defines its own voice identity; TTS switches per-agent on each response.

PR #2256, opened 2026-07-05
chore(monitoring): thin V1-pull detection slice — re-arm NATS ACL/log alerting

### LINKED ISSUE #2245: chore(monitoring): thin V1-pull detection slice — re-arm NATS ACL/log alerting without reviving the frozen host-timer
**Origin:** #2203 (incident-response runbook) — split out per architect/axial-drift/security/devops panel review, 2026-07-04.

**Problem**: The 2026-04-27 NATS ACL incident took 2h44m to detect — zero alerts fired, found by manually reading logs. The alerting code that *would* have caught it (`check_nats_log_errors` in `src/factory/monitoring/checks_log.py`, scans container logs for "permissions violation", escalates via `src/factory/monitoring/escalation.py` → Telegram) already exists and was marked done in the original postmortem. But its scheduler — the host-side `deploy/lyra-monitor.{service,timer}` systemd timer — was removed in `deploy/provision.sh` ("superseded by Monitoring v2, tracked in #1035"). #1035 (Monitoring v2 — NATS event stream + Tauri desktop dashboard) closed without shipping that replacement (no `lyra.event.*`/`lyra.metric.*` producers exist in `src/`, no dashboard). ADR-091 (four observability planes, "design-only ratification — no runtime ships") reframes monitoring a third time but also has zero shipped runtime today. Net: 3 generations of monitoring plans, 0 shipped detection in production.

**Do NOT re-arm the retired `deploy/lyra-monitor.{service,timer}` unit as-is**: #1035 explicitly says "Do not extend the host-timer monitor" — reviving it verbatim contradicts that closed epic's own directive, and it has the host-systemd-dependency problems (`systemctl --user`, podman socket) #1035 correctly flagged.

**Direction**: ADR-091 plane ③ (internal infra state) itself prescribes "V1: pull (`/health/detail`, monitoring checks)" as the current-correct mechanism — V2 (subscribe-based, `factory.metric.*`) is future work under #1035 that was never built. So the right thin slice is a **new, minimal, container-native periodic check** — NOT the old host-systemd-timer architecture — that:
1. Reuses the existing pure check functions (`check_nats_log_errors`, `check_hub_dict_stream_gen_timeout` in `checks_log.py`) — these only need `podman logs <container>`, not the host `systemctl --user` calls that made the old timer host-bound.
2. Reuses the existing, still-functional Telegram escalation path (`escalation.py`).
3. Runs from inside the container stack (e.g. a small Quadlet timer unit or an in-process scheduler on an already-running service with a scoped podman-socket mount) — not a host-side `--user` systemd timer.
4. Is explicitly framed as the ADR-091 plane③ V1-pull mechanism, not a resurrection of the frozen `src/factory/monitoring/` host-timer nor a preview of full Monitoring v2/Sentinelle (both far larger scopes).

This is the actual closer of the April 2h44m detection gap — #2203's incident-response runbook only defines process *after* a human notices; without this, detection stays manual.

**Acceptance**:
- [ ] A periodic check runs `check_nats_log_errors` (and ideally `check_hub_dict_stream_gen_timeout`) against the NATS/hub containers on a short interval (≤5 min) from inside the container stack, no host-timer/systemd --user dependency
- [ ] On anomaly, a Telegram alert fires via the existing `escalation.py` path
- [ ] `deploy/lyra-monitor.{service,timer}` remains retired/untouched — this does not resurrect it
- [ ] Design note in the PR explains why this is plane③ V1-pull (ADR-091) and not a Monitoring-v2/Sentinelle preview

Ref: validated by architect/axial-drift/security/devops panel review, 2026-07-04.

MickaelV0:
## Summary
- Adds `factory-log-monitor`, a standalone Quadlet container that periodically re-runs the two existing log-scanning checks (`check_nats_log_errors`, `check_hub_dict_stream_gen_timeout`) via a Loki LogQL pull, alerting to Telegram on failure — restoring detection for the class of incident that took 2h44m to notice on 2026-04-27.
- Refactors `checks_log.py` behind an injectable `LogFetcher` seam (`SubprocessLogFetcher` default, unchanged behavior; new `LokiLogFetcher` for the deployed container, which has no Podman socket/CLI).
- Narrows `factory.monitoring`'s package-level `DeprecationWarning` off the two modules this change makes live again (`checks_log.py`, `escalation.py`), leaving the aggregate Layer-1 `checks.py` runner still deprecated.

## Design note (ADR-091 plane③, SC8)

This is the **V1-pull** mechanism ADR-091 already prescribes for plane③ (infra health as *state*, consumed via pull — `/health/detail`, monitoring checks) — **not** a revived host-timer and **not** Monitoring v2 / full Sentinelle (`factory.metric.*` JetStream subscribe, Tauri dashboard — that's #1035, unshipped, explicitly out of scope here). `deploy/lyra-monitor.{service,timer}` is untouched (see SC6 below); no `systemctl --user` introspection, no `localhost:8443` host-network assumption — the two problems #1035 flagged when retiring the old timer.

**Standalone container vs. ADR-091's "Sentinelle is a hub-module, not a standalone container" default:** ADR-091 describes Sentinelle (the eventual V2 in-process capability) as living inside `factory-hub`. This slice deliberately ships a **separate, standalone** `factory-log-monitor` container instead. That is not a contradiction: `factory-log-monitor` is not Sentinelle — it's an external, independent safety-net layer with no dependency on hub process health, so a hub crash/restart/stall (exactly the failure mode this alerting exists to catch) can't silently take the monitor down with it. When V2/Sentinelle ships as a hub-module, this slice's pull-based container can be retired or downgraded to a dead-man's-switch without re-architecting the detection story.

**Accepted residual risk:** this mechanism depends on Loki being up and receiving container logs (ADR-093). If Loki itself is down, `LokiLogFetcher` raises `LogFetchError`, which is reported as a failing `CheckResult` (visible failure, not silent) but does not independently alert on Loki's own health — monitoring the monitor is out of scope for this slice.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #2245: thin V1-pull detection slice | Approved |
| Frame | [2245-thin-v1-pull-detection-slice-frame.mdx](artifacts/frames/2245-thin-v1-pull-detection-slice-frame.mdx) | Present |
| Spec | [2245-thin-v1-pull-detection-slice-spec.mdx](artifacts/specs/2245-thin-v1-pull-detection-slice-spec.mdx) | Present |
| Plan | [2245-thin-v1-pull-detection-slice-plan.mdx](artifacts/plans/2245-thin-v1-pull-detection-slice-plan.mdx) | Present |
| Implementation | 17 commits on `feat/2245-thin-v1-pull-detection-slice` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (39 new: 25 checks_log + 7 log_watch + 7 standalone-paths, + 3 deprecation) | Passed |

## Test Plan
- [x] `PYTHONPATH=src pytest tests/test_monitoring_checks_log.py tests/test_monitoring_log_watch.py tests/bootstrap/ tests/test_monitoring_deprecation.py` — 366 passed
- [x] Full suite: `PYTHONPATH=src pytest -q` — 6537 passed, 26 skipped, 1 xfailed
- [x] `podman quadlet --dryrun --user` against `factory-log-monitor.container` (Podman 6.0.0 local)
- [x] `tools/check_quadlet_template_purity.sh` — exit 0
- [ ] Post-merge: mint `~/.roxabi/factory/telegram-monitor-{token,chat-id}.tok` on M1 (currently `policy=optional` — container will crash-loop with a clear "optional secret missing" log until provisioned, by design, does not block `make converge`)

## SC → Test Matrix

| SC | Test(s) | Status |
|----|---------|--------|
| SC1: `factory-log-monitor` runs both checks on ≤5min interval, no `systemctl --user`, no host-timer unit | `tests/test_monitoring_log_watch.py :: TestRunOnce` (all 3), `tests/test_monitoring_checks_log.py :: TestSubprocessInvocationShape` (both — proves `podman logs`, never `systemctl`) | ✓ proven |
| SC2: Loki LogQL over `roxabi.network`, no Podman socket/CLI in image, no image/Dockerfile/publish.yml changes | `tests/test_monitoring_checks_log.py :: TestLokiLogFetcher` (all 4) + `git diff --stat origin/staging..HEAD -- '*.hcl' '*Dockerfile*' '.github/workflows/publish.yml'` (empty) | ✓ proven |
| SC3: default (subprocess) fetcher behavior unchanged — pre-existing tests pass unmodified | `tests/test_monitoring_checks_log.py :: TestCheckNatsLogErrors`, `TestCheckHubDictStreamGenTimeout`, `TestLogCheckExceptionHandling` (T0 baseline, 18 tests, green before AND after the T1-T3 refactor) | ✓ proven |
| SC3b: both fetchers raise shared `LogFetchError`, caught uniformly | `tests/test_monitoring_checks_log.py :: TestLogCheckExceptionHandling` (subprocess side) + `TestLokiLogFetcher::test_fetch_raises_log_fetch_error_on_malformed_json`, `test_fetch_raises_log_fetch_error_on_http_error` | ✓ proven |
| SC4: failed check → `send_telegram_raw_alert` called with correct `HealthReport` | `tests/test_monitoring_log_watch.py :: TestRunForever::test_run_forever_one_fail_alerts_once_with_both_results`, `test_run_forever_both_fail_alerts_once_no_duplicates` | ✓ proven |
| SC5: all-pass → zero Telegram calls | `tests/test_monitoring_log_watch.py :: TestRunForever::test_run_forever_all_pass_does_not_alert` | ✓ proven |
| SC6: `deploy/lyra-monitor.{service,timer}` zero diff | `git diff --stat origin/staging..HEAD -- 'deploy/lyra-monitor.*'` (empty) | ✓ proven |
| SC7: package-level `DeprecationWarning` no longer fires on `checks_log`/`escalation` import | `tests/test_monitoring_deprecation.py` (all 3) | ✓ proven |
| SC8: PR description design note (ADR-091 plane③, non-host-timer, non-Sentinelle framing, Loki residual risk) | — (this PR description, "Design note" section above) | ✓ proven |
| SC9: `quadlet-lint` passes for the new unit | `podman quadlet --dryrun --user` (local, Podman 6.0.0) + `tools/check_quadlet_template_purity.sh` — CI's `quadlet-lint.yml` is authoritative, confirmed via `/ci-watch` | ⏳ not run (CI) |
| SC10: no `factory.event.*`/`factory.metric.*` bus producer/consumer introduced | `tests/bootstrap/test_log_monitor_standalone_paths.py :: TestNoEventMetricSubjects` (all 3) | ✓ proven |

## Falsification Evidence (Gate #280)

```
broke checks_log.py (LokiLogFetcher LogQL filter removed) → test_fetch_query_param_includes_case_insensitive_regex_filter FAILED
  assert '|~ "(?i)permissions\ violation"' in '{systemd_unit="factory-nats.service"}'
→ reverted → 25/25 GREEN

broke checks_log.py (count = 0 unconditionally, both check functions) → test_fails_with_correct_count_on_occurrences
  AND test_fails_when_count_meets_threshold FAILED (assert True is False)
→ reverted → 25/25 GREEN

broke log_watch.py (alert guard bypassed, always fires) → test_run_forever_all_pass_does_not_alert FAILED
  AssertionError: Expected 'mock' to not have been called. Called 1 times.
→ reverted → 7/7 GREEN

broke worker_standalone.py (once-mode sys.exit removed) → test_once_mode_all_pass_exits_zero,
  test_once_mode_failure_exits_one, test_once_mode_never_calls_run_forever all FAILED
  ("DID NOT RAISE SystemExit")
→ reverted → 9/9 GREEN

broke log_watch.py (inserted "# factory.metric.foo" literal) → test_log_watch_module_has_no_event_or_metric_subjects
  FAILED (AssertionError: 'factory.metric.' in source)
→ reverted → 9/9 GREEN
```

Fixes #2245

---

### DIFF
--- a/artifacts/frames/2245-thin-v1-pull-detection-slice-frame.mdx
+++ b/artifacts/frames/2245-thin-v1-pull-detection-slice-frame.mdx
@@ -1,0 +1,106 @@
+---
+title: "chore(monitoring): thin V1-pull detection slice — re-arm NATS ACL/log alerting without reviving the frozen host-timer"
+issue: 2245
+status: approved
+tier: F-lite
+date: 2026-07-05
+---
+
+## Problem
+
+The 2026-04-27 NATS ACL incident took 2h44m to detect — zero alerts fired, found only
+by a human manually reading container logs. The alerting logic that would have caught
+it (`check_nats_log_errors` in `src/factory/monitoring/checks_log.py`, scans container
+logs for "permissions violation", escalates via `src/factory/monitoring/escalation.py`
+→ Telegram) already exists and was marked done in the original postmortem. Its
+scheduler — the host-side `deploy/lyra-monitor.{service,timer}` systemd timer — was
+removed in `deploy/provision.sh` ("superseded by Monitoring v2, tracked in #1035").
+#1035 (Monitoring v2 — NATS event stream + Tauri dashboard) closed without shipping
+that replacement (no `lyra.event.*`/`lyra.metric.*` producers in `src/`, no dashboard).
+ADR-091 (four observability planes) reframes monitoring a third time but also has zero
+shipped plane③ runtime today. Net: three generations of monitoring plans, zero shipped
+detection in production — the April 2h44m gap is still open.
+
+**Why now:** split out of #2203 (incident-response runbook) per architect/axial-drift/
+security/devops panel review (2026-07-04) — the runbook only defines process *after* a
+human notices; without this slice, detection stays manual. **Observable impact:** any
+repeat NATS ACL misconfiguration or hub `_dict_stream_gen` stall goes undetected until
+someone happens to read logs.
+
+## Who
+
+- **Primary:** Mickael (sole on-call operator) — needs a Telegram page instead of a
+  cold log read when a permissions-violation burst or hub stream-gen timeout occurs.
+- **Secondary:** future Sentinelle/Monitoring-v2 work (#1035) — this slice must not
+  foreclose the plane③ V2 subscribe-based design; it is explicitly the V1-pull
+  placeholder ADR-091 already prescribes.
+
+## Constraints
+
+- ADR-091 plane③ boundary rule: infra health is **state**, consumed via **V1: pull**
+  (`/health/detail`, monitoring checks) until #1035's V2 (`factory.metric.*` subscribe)
+  ships — this slice is the V1-pull mechanism, nothing upstream of it.
+- Must **not** re-arm `deploy/lyra-monitor.{service,timer}` verbatim — #1035 explicitly
+  said "do not extend the host-timer monitor," and flagged its host-systemd-dependency
+  problems (`systemctl --user` service introspection via `check_process`, direct
+  `localhost:8443` HTTP assumption via `check_http_health`).
+- Must reuse the existing pure check functions (`check_nats_log_errors`,
+  `check_hub_dict_stream_gen_timeout` in `checks_log.py`) — they only shell out to
+  `podman logs <container>`, no host systemd coupling.
+- Must reuse the existing Telegram escalation path (`escalation.py`) — still
+  functional, no changes needed to its contract.
+- Interval ≤5 minutes (matches the acceptance criterion and the existing
+  `check_interval_minutes` default in `MonitoringConfig`).
+- `deploy/lyra-monitor.{service,timer}` stays retired/untouched — no file changes there.
+- Depends on #2203 (incident-response runbook) only for a documentation cross-reference
+  (this issue's PR should point at the runbook path once it exists) — not a build
+  dependency; #2203 may land before or after this PR.
+- Size `F-lite`, `P1-high` — single domain (`src/factory/monitoring/` + one new
+  deploy surface), no new architecture, panel-reviewed scope.
+
+## Out of Scope
+
+- Reviving `deploy/lyra-monitor.{service,timer}` as-is, or any host-side `--user`
+  systemd timer that re-implements `check_process`'s `systemctl --user is-active`
+  introspection or `check_http_health`'s `localhost:8443` assumption.
+- Monitoring v2 / full Sentinelle (`factory.metric.*` JetStream producers/subscribers,
+  Tauri dashboard, event-bus plane③) — that is #1035, explicitly future work.
+- Any other check beyond the two named log checks (disk, blobstore, audio-lag,
+  process-liveness, queue-depth, etc. from the full `run_checks()` pipeline) — those
+  stay on their current path (or unmonitored) and are not this issue's concern.
+- Writing #2203's incident-response runbook content — this issue only references it.
+- New alert channels beyond Telegram (Discord, PagerDuty, etc.).
+
+## Premise Validity
+
+**Success in 6 months:** a periodic, container-native check (≤5 min interval) runs
+`check_nats_log_errors` (and `check_hub_dict_stream_gen_timeout`) against the
+NATS/hub containers with no host-timer/`systemctl --user` dependency; on a real
+anomaly a Telegram alert fires via `escalation.py` within one check interval — a
+repeat of an April-2026-class incident is caught by an automated alert, not by a
+human reading logs.
+
+**Failure in 6 months (falsifiable):** at ≥6 months (2027-01), a NATS
+permissions-violation burst or hub `_dict_stream_gen` stall occurs in production logs
+but zero Telegram alert is found in the corresponding window — i.e. the periodic
+check silently stopped running, was never wired into the deploy surface, or its
+escalation path was broken — verified by diffing incident log timestamps against
+Telegram alert history.
+
+**Simplest alternative:** re-enable `deploy/lyra-monitor.{service,timer}` unchanged.
+**Why not simplest:** it re-introduces exactly the host-systemd-dependency problems
+(`systemctl --user` introspection of other services, `localhost:8443` host-network
+assumption) that #1035 flagged when retiring it, and reviving it verbatim directly
+contradicts that closed epic's own directive ("do not extend the host-timer
+monitor") — it does not fit the now fully-containerized (Quadlet) deployment model
+this repo has since standardized on.
+
+## Complexity
+
+**Tier: F-lite** — label `size:F-lite`; single domain (`src/factory/monitoring/` +
+one new small deploy/scheduling surface), clear panel-reviewed scope, no new
+architecture pattern beyond picking a container-native scheduling mechanism among the
+two options ADR-091/the issue already names (Quadlet timer vs. in-process scheduler
+with scoped podman-socket mount) — that choice is deferred to `/spec` + `/plan`.
+
+Signals: clear scope ∨ single domain ∨ issue label F-lite — all present.
--- a/artifacts/plans/2245-thin-v1-pull-detection-slice-plan.mdx
+++ b/artifacts/plans/2245-thin-v1-pull-detection-slice-plan.mdx
@@ -131,8 +131,13 @@ flowchart LR
 | Agent | Instances | Task count | Files |
 |-------|-----------|------------|-------|
 | backend-dev | A, B | 6 (A: 3, B: 3) | `src/factory/monitoring/checks_log.py`, `src/factory/monitoring/log_watch.py` (new), `src/factory/monitoring/__init__.py`, `src/factory/bootstrap/standalone/worker_standalone.py`, `src/factory/cli/main.py` |
-| tester | A | 4 | `tests/test_monitoring_checks_log.py` (new), `tests/test_monitoring_log_watch.py` (new), `tests/bootstrap/test_log_monitor_standalone_paths.py` (new), `tests/test_monitoring_deprecation.py` (new) |
-| devops | A | 3 | `deploy/quadlet/factory-log-monitor.container` (new), `deploy/quadlet.toml` |
+| tester | A, B | 5 (A: 3, B: 2) | `tests/test_monitoring_checks_log.py` (new), `tests/test_monitoring_log_watch.py` (new), `tests/bootstrap/test_log_monitor_standalone_paths.py` (new), `tests/test_monitoring_deprecation.py` (new) |
+| devops | A | 3 | `deploy/quadlet/factory-log-monitor.container` (new), `deploy/quadlet.toml`, `deploy/secrets-policy.toml`, `docs/CONFIGURATION.md` |
+
+**Post-`/plan` correction (advisor review, pre-`/implement`):** the tester agent force-splits into
+A/B because Wave 0's new baseline task (T0, see below) pushes tester-A's task count to 5 (>4 cap).
+T0 + T4 + T6 → tester-A (checks_log/log_watch subject-chain); T13 + T8 → tester-B (deprecation +
+CLI-smoke subject-chain).
 
 No frontend-dev (no UI surface), no architect/security-auditor/doc-writer instance needed for
 implementation — architecture and security tradeoffs (Podman-socket rejection, Loki pivot) were
@@ -148,7 +153,7 @@ Every Success Criterion (SC1–SC10, SC3b) traces to ≥1 micro-task; every micr
 |----|-----------|
 | SC1 (≤5min interval, no systemctl/host-timer) | T5, T9 |
 | SC2 (Loki fetch, no Podman socket/CLI, no new image) | T2, T9 |
-| SC3 (default-fetcher behavior unchanged) | T1, T3, T4 |
+| SC3 (default-fetcher behavior unchanged) | T0, T1, T3 |
 | SC3b (unified `LogFetchError`) | T1, T2, T3, T4 |
 | SC4 (Telegram alert on failure) | T5, T6 |
 | SC5 (all-pass → zero Telegram calls) | T5, T6 |
@@ -158,11 +163,11 @@ Every Success Criterion (SC1–SC10, SC3b) traces to ≥1 micro-task; every micr
 | SC9 (`quadlet-lint` passes) | T9, T10, T11 |
 | SC10 (no event/metric bus producer) | T5, T7, T8 |
 
-Slices: SL1→T1–T4, SL2→T5–T6, SL3→T7–T8, SL4→T9–T11, SL5→T12–T13, SL6→exempted (PR-description
+Slices: SL1→T0–T4, SL2→T5–T6, SL3→T7–T8, SL4→T9–T11, SL5→T12–T13, SL6→exempted (PR-description
 deliverable, written during `/pr`, not a code micro-task — consistent with `/plan`'s "Tier F"
 process, which generates micro-tasks for code slices only).
 
-Covered: 11/11 SC · 6/6 slices (5 code slices → 13 tasks, 1 doc slice → PR-step). Uncovered: none.
+Covered: 11/11 SC · 6/6 slices (5 code slices → 14 tasks, 1 doc slice → PR-step). Uncovered: none.
 
 ## Reference Patterns (Step 3)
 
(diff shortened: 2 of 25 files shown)


MickaelV0:
## Code Review

Multi-domain review (6 fresh agents: security-auditor, architect, product-lead, tester, backend-dev, devops) against the full diff (single-chunk, ~35k tokens). Spec-compliance cross-checked: all 10 SC + SC3b spot-verified against actual test file content, not just the PR body's claimed matrix.

**Verdict: Request changes**

### Blockers

**issue (blocking): SC7's guarding test can't detect its own regression under realistic multi-file collection order (test-tautology)**
`tests/test_monitoring_deprecation.py:52-73`
-- tester
Root cause: `_fresh_import()` evicts only the named submodule (+ `also_evict` targets) from `sys.modules`, never the parent `factory.monitoring` package. Under this repo's real pytest-xdist collection (`-n auto --dist=loadgroup`), other test files import `factory.monitoring.*` before this file runs, so `sys.modules["factory.monitoring"]` is already cached — reimporting a child module doesn't re-execute the parent's `__init__.py`. Reproduced empirically: re-appending the exact `warnings.warn(...)` block this PR deletes from `factory/monitoring/__init__.py`, then running `pytest tests/ -k "monitoring or log_monitor"` (the realistic condition) → 121 passed, 0 failed, even though the warning fired 24 times elsewhere in the run.
Independently verified (code-review lead): read `src/factory/monitoring/__init__.py` directly — it currently contains zero `warnings.warn` code, confirming SC7's actual production behavior is correct today. This is a test-quality gap, not a live defect: the test would fail to catch a *future* regression, not indicate a *current* one.
Class: [test-tautology, candidate/test-isolation-cache-leak]
Solutions:
  1. Add `also_evict=("factory.monitoring",)` to both `_fresh_import` calls in `test_live_v1_pull_modules_no_deprecation_warning` (one-line fix).
Confidence: 88%

**issue (blocking): unguarded exception paths let a deterministic Loki response-shape bug permanently kill the monitor**
`src/factory/monitoring/checks_log.py:86`, `src/factory/monitoring/log_watch.py:69-70,77`
-- backend-dev, architect (independent convergence — flagged from two different angles)
Root cause: `LokiLogFetcher.fetch()`'s except tuple (`httpx.HTTPError, KeyError, ValueError`) doesn't cover `TypeError` (e.g. `data["data"]` is `None`) or `IndexError` (e.g. a `values` entry has fewer than 2 elements) from a malformed-but-200-OK Loki response. Any such exception isn't `LogFetchError`, so the check functions' `except LogFetchError:` doesn't catch it either. `LogMonitorLoop.run_forever()` only wraps the Telegram-alert dispatch in try/except — not the `report = await self.run_once()` call — so the coroutine dies. Because a response-shape bug is deterministic (reproduces every restart, not transient), `Restart=on-failure` exhausts `StartLimitBurst=5`/`StartLimitIntervalSec=60` and systemd permanently stops the unit — silently disabling the exact safety net this slice exists to re-arm.
Raw callsites: [{file: "src/factory/monitoring/checks_log.py", line: 86}, {file: "src/factory/monitoring/log_watch.py", line: 70}, {file: "src/factory/monitoring/log_watch.py", line: 77}]
Solutions:
  1. Reuse the shared `_MONITORING_HTTP_ERRORS` tuple from `_errors.py` (already includes `TypeError`) plus `KeyError`, instead of `LokiLogFetcher`'s one-off tuple (also closes Warning W2 below).
  2. Defense-in-depth: wrap `run_once()` itself in `run_forever()`'s loop — log and continue to the next tick on any unexpected exception, so no single bad tick can kill the safety net.
Confidence: 78%

### Warnings (non-blocking, but flagged for explicit human sequencing decision)

**issue: unprovisioned optional secrets can hard-fail the next fleet-wide `make converge`, not just leave this unit degraded**
`deploy/quadlet/factory-log-monitor.container:30`, `deploy/converge.sh:164,167`
-- devops
Root cause: this PR's new Quadlet unit files are host-carried (not image-carried/inert), so converge classifies them as structural drift and runs a full client-restart loop on next converge. The two new secrets (`factory-telegram-monitor-token`/`-chat-id`) are net-new and unprovisioned — `podman run --secret <missing>` fails synchronously, `systemctl --user restart factory-log-monitor` returns non-zero, and `converge.sh` exits 1 for the whole batch before the convergence stamp is written. This is more severe than the PR body's own framing ("crash-loop with a clear log, does not block make converge") — it's a pre-existing `converge.sh` gap (not introduced by this PR), but this PR is what exposes it, since its secrets are genuinely absent (unlike the one other `policy=optional` secret in the repo, which is already provisioned in prod).
Corroborated independently by product-lead's review (thought, 60%): the PR's own Test Plan checklist item "mint secrets on M1 post-merge" is unchecked.
Solutions:
  1. Mint both secrets on M1 (`podman secret create factory-telegram-monitor-token -` / `...-chat-id -`) BEFORE this PR's merge triggers a converge — pure sequencing, no code change.
  2. Separately (out of scope for #2245): harden `converge.sh` to treat a restart failure as non-fatal when the failing unit's `required_secrets` are all `policy=optional` and absent.
Confidence: 78%

**suggestion: `LokiLogFetcher` forked its own narrower exception tuple instead of reusing the shared `_MONITORING_HTTP_ERRORS`**
`src/factory/monitoring/checks_log.py:11,86`, `src/factory/monitoring/_errors.py:10`
-- backend-dev
Root cause: `checks.py`, `checks_varz.py`, `checks_audio.py` all import and catch the shared `_MONITORING_HTTP_ERRORS` for the same "GET + parse JSON" shape; this new fetcher wrote its own tuple instead. Direct root cause of the Blocker above.
Class: [parallel-path-drift]
Confidence: 70%

**suggestion: `container_name` interpolated unescaped into the LogQL label matcher (asymmetric with `pattern`, which is `re.escape()`'d)**
`src/factory/monitoring/checks_log.py:64`, `src/factory/monitoring/config.py:43-44`
-- security-auditor, backend-dev
Root cause: `nats_container_name`/`hub_container_name` have no `field_validator`, unlike `service_names` (guarded by `_SERVICE_NAME_RE`). No live exploit path — both fields are operator-controlled `config.toml` values, not attacker-facing.
Solutions:
  1. Reuse `_SERVICE_NAME_RE` as a validator on both fields, for parity with `service_names`.
Confidence: 60%

**issue: `docs/CONFIGURATION.md:667`'s "retained for Monitoring v2 (#1035) spec mining" claim is now stale**
`docs/CONFIGURATION.md:667`
-- product-lead
Root cause: this PR correctly updates the near-identical `#1035` reference in the Bot-credentials table (~line 438) but leaves this second, related section untouched, even though the spec itself (`artifacts/specs/2245-...-spec.mdx:100`) quotes nearly the same phrase to justify a different change (SC7) — meaning the author was aware of this exact language. `checks_log.py`/`escalation.py` are no longer "retained for spec mining"; they're live, scheduled, alerting code as of this PR.
Solutions:
  1. Update the line-667 section in this same PR to distinguish the now-live modules from the still-dormant `checks.py`/`__main__.py`.
Confidence: 88%

**thought: the V1-pull-vs-hub-module design rationale should land in the durable SSoT, not only the PR description**
`docs/architecture/observability.md`
-- architect
Root cause: ADR-091's deployment table doesn't list a standalone external-monitor container; `observability.md` (the doctrine-designated SSoT) never records that this layer is now shipped, or that it should be retired/subsumed when V2/#1035 ships as a hub module. Without that, a future implementer risks duplicating the scheduler.
Class: [candidate/architecture-ssot-drift]
Confidence: 80%

### Suggestions (non-blocking, lower priority)

- Mutable default fetcher instances (`LokiLogFetcher()`, `SubprocessLogFetcher()`) — confirmed stateless today, latent footgun if either gains state later. (architect, 70%)
- `factory/monitoring/__init__.py` docstring still frames the whole package as dormant/spec-reference despite `checks_log`/`escalation` now being live. (architect, 65%)
- No test directly asserts `LokiLogFetcher.fetch()`'s "zero matching streams" return value (`== ""`) — currently covered only incidentally by an unrelated test's fixture. (tester, 68%)
- No test at the tight `threshold - 1` boundary for `check_hub_dict_stream_gen_timeout` at the realistic default `threshold=3`. (tester, 55%)
- `LogFetchError` collapses "Loki unreachable" and "genuine threshold exceeded" into an identical `CheckResult`/Telegram shape — acceptable as a documented V1 limitation. (backend-dev, 55%)
- `docs/CONFIGURATION.md`'s updated rows say unqualified "Yes" without noting the variable stays optional/best-effort for other (legacy) consumers. (devops, 55%)
- Loki query's `limit`/`start`/`end` params are never asserted despite a source comment calling `limit` load-bearing. (tester, 40%)
- Quadlet `After=` includes `factory-hub.service factory-nats.service`, which this unit doesn't actually depend on (only Loki). (devops, 40%)
- `_bootstrap_log_monitor_standalone` omits the stop/shutdown log line its sibling bootstrap functions use. (backend-dev, 35%)
- SC1's "hard cap ≤5 minutes" is enforced only by the config default, not a validator ceiling. (product-lead, 55%)
- `_log_monitor` CLI wrapper has zero executed-path coverage — consistent with the existing convention for all sibling standalone commands, not a regression specific to this PR. (tester, 30%)

### Praise

- Clean scope discipline + genuinely falsification-gated tests: the T0-baseline-preserved claim and the Falsification Evidence block both check out against real test source, not just the PR body's assertion. Zero touches on `deploy/lyra-monitor.*`, zero event/metric-bus creep (statically guarded). (product-lead, 95%)
- Container hardening (`NoNewPrivileges`, `ReadOnly`, `DropCapability=all`, `UserNS=keep-id`) and secret delivery match sibling units exactly; the log-to-Telegram path was confirmed to never leak raw log content — only counts/prose reach the alert body, and delivery uses no `parse_mode`. (security-auditor, devops)

---

MickaelV0:
## Code Review

**Provenance disclosure:** this review's Phase 3 was performed as a single implementer self-review, not literal fresh multi-agent dispatch — the execution environment for this pipeline run has no `Task` tool available to spawn independent `security-auditor`/`architect`/`product-lead`/`tester`/`backend-dev`/`devops` agents as the skill specifies. I applied each domain's lens directly against the full diff (`git diff staging...HEAD`, 24 files) and against the three new/changed test files read in full. Weight this accordingly: it lacks the independence the fresh-agent design exists to provide, since I also authored the fixes under review.

Iteration 2 of the fix→review loop (cap = 2). Latest commit `c880127b2`. CI green (`run PASSED in 3m 9s`, all jobs).

**Verdict: Approve with comments — code-complete; do NOT add `reviewed` until the two M1 Telegram secrets are minted (see Warning below) — the next `make converge` will hard-fail the fleet-restart batch otherwise**

### Blockers

None. Both blockers from the prior review round are verified fixed:

- **SC7 test-tautology** (`tests/test_monitoring_deprecation.py`): `_fresh_import`'s `also_evict` now correctly evicts the parent `factory.monitoring` package via `importlib.reload()` (preserves sibling submodule attributes set by other xdist-collected test files) while evicting *leaf* modules via `sys.modules.pop()` (forces the importer's own import statement to be responsible for re-execution, proving transitive propagation rather than re-triggering the target directly). Verified: deletion check (removing `__main__.py`'s `from .checks import run_checks` correctly fails `test_dormant_modules_still_warn`), regression-injection check (re-adding `warnings.warn()` to `__init__.py` correctly fails `test_live_v1_pull_modules_no_deprecation_warning`), single-worker collision check with `test_monitoring_escalation.py` (12 passed, 0 errors).
- **Unguarded exception paths** (`checks_log.py`, `log_watch.py`): `_LOKI_FETCH_ERRORS = (*_MONITORING_HTTP_ERRORS, KeyError, IndexError)` now covers the malformed-response shapes; `LogMonitorLoop.run_forever()` wraps `run_once()` itself in `except Exception` (logs + continues to next tick) — verified this deliberately does not swallow `asyncio.CancelledError` (a `BaseException` subclass since 3.8, so graceful shutdown/task cancellation still propagates). Directly tested in `tests/test_monitoring_log_watch.py::TestRunForever::test_run_forever_survives_unexpected_run_once_exception`.

### Warnings (non-blocking, flagged for explicit human sequencing decision — unchanged from prior round)

**issue: unprovisioned optional secrets can hard-fail the next fleet-wide `make converge`, not just leave this unit degraded**
`deploy/quadlet/factory-log-monitor.container:30`
Still open — out of scope for a worktree-local code fix. The two new secrets (`factory-telegram-monitor-token`/`-chat-id`) remain unprovisioned on M1. This unit's Quadlet file is host-carried, so the next `make converge` will attempt a full client-restart batch; `podman run --secret <missing>` fails synchronously and (per the pre-existing `converge.sh` gap this PR exposes, not introduces) can exit the whole batch non-zero before the convergence stamp is written.
Solutions: mint both secrets on M1 **before** merge triggers a converge (pure sequencing, no code change) — this is a real pre-merge operational action, intentionally not attempted as part of this fix cycle (out of scope: no SSH, no converge/deploy tooling, no minting real credentials from this worktree).
Confidence: 78%

### Suggestions (non-blocking, lower priority — carried over, still unaddressed; each individually low-confidence and accepted as V1 scope)

- Mutable default fetcher instances (`LokiLogFetcher()`, `SubprocessLogFetcher()`) — confirmed stateless today, latent footgun if either gains state later. 70%
- No test at the tight `threshold - 1` boundary for `check_hub_dict_stream_gen_timeout` at the realistic default `threshold=3`. 55%
- `LogFetchError` collapses "Loki unreachable" and "genuine threshold exceeded" into an identical `CheckResult`/Telegram shape — documented V1 limitation. 55%
- `docs/CONFIGURATION.md`'s updated Telegram rows say unqualified "Yes" without noting the variable stays optional/best-effort for other (legacy) consumers. 55%
- Quadlet `After=` includes `factory-hub.service factory-nats.service`, which this unit doesn't actually depend on (only Loki). 40%
- `_bootstrap_log_monitor_standalone` omits the stop/shutdown log line its sibling bootstrap functions use. 35%
- SC1's "hard cap ≤5 minutes" is enforced only by the config default, not a validator ceiling. 55%
- `_log_monitor` CLI wrapper has zero executed-path coverage — consistent with the existing convention for all sibling standalone commands, not a regression specific to this PR. 30%

Checked and found no new issue: poll-window coverage (`check_interval_minutes=5` vs. `nats_log_check_minutes=30`/`stream_gen_timeout_minutes=30`) — fetch windows are 6x wider than the tick interval, so consecutive ticks overlap and there is no unscanned gap between them.

### Praise

- Both prior blockers fixed with verification evidence, not just code changes: deletion checks, regression-injection checks, and a real single-worker xdist-collision reproduction, all re-run and confirmed after the fix.
- Went beyond the letter of the original SC7 finding: found and fixed a second, more subtle bug in the same `_fresh_import` helper — the fix for the "evict the parent package" issue initially risked breaking the *leaf*-module eviction case (`test_dormant_modules_still_warn`'s `also_evict=("factory.monitoring.checks",)`), because reloading a leaf module in place would fire its warning immediately inside the same `catch_warnings` block regardless of whether `__main__` actually imports it — silently testing nothing about transitive propagation. Correctly resolved by branching on `hasattr(cached, "__path__")` (packages reload in place, leaves get popped). Confirmed via an explicit deletion check.
- Container-name `field_validator` reuses the existing `_SERVICE_NAME_RE` pattern for parity with `service_names`, and its test (`test_monitoring_config.py`) includes a genuinely adversarial LogQL-injection-shaped value (`"factory-nats\"} |~ \".*"`), not just a generic invalid string.
- SSoT documentation debt from the architect's original finding is closed: `docs/architecture/observability.md` now has a "Standalone log monitor (#2245) — plane③, V1-pull" subsection, and `docs/CONFIGURATION.md`'s stale `#1035` framing is corrected.
- Full test suite green (6546 passed, 26 skipped, 1 xfailed, 0 errors) and full pre-push quality gate green (typecheck 0 errors/warnings, doc-drift, secrets-drift, ACL-drift all OK) — re-verified after working around an unrelated shared-venv cross-worktree contamination (see report to task owner for details; not a defect in this PR).

---
Generated with [Claude Code](https://claude.com/claude-code) via `/code-review` (implementer self-review — no `Task` tool available in this execution environment; see provenance disclosure above)

